PR #1, opened 2020-04-09
onboard ibm-common-service-operator

chenzhiwei:


go.mod
```diff
@@ -0,0 +1,42 @@
+module github.com/IBM/ibm-common-service-operator
+
+go 1.13
+
+require (
+	github.com/ghodss/yaml v1.0.1-0.20190212211648-25d852aebe32
+	github.com/operator-framework/operator-sdk v0.16.0
```
DanielXLee, line 7: do we need to upgrade sdk version to v0.16.0?

chenzhiwei: these code are generated automatically, maybe because I use the v0.16.0 to generate the code. Let's use the v0.16.0 in this repo.

PR #9, opened 2020-04-15
Update operandconfig instance

horis233:
This PR is to update `operandconfig` instance, remove `navconfiguration: {}` from ibm-commonui.


morvencao:
Please add `OWNERS` file in the root of the repo in separate PR, so that `/approve` in follow up PRs can take effect.

PR #40, opened 2020-05-05
Moved Secretshare operator to quay.io/opencloudio

mikekaczmarski:


build/resources/extra/secretshare_deploy.yaml
```diff
@@ -26,7 +25,6 @@ apiVersion: v1
 kind: ServiceAccount
 metadata:
   name: secretshare
-  namespace: ibm-common-services
```
horis233, line 29: @mikekaczmarski Why do we need to remove the namespace here

chenzhiwei: maybe a rebase conflict?

mikekaczmarski: Yes, looks like an oversight.  I will re-do

PR #41, opened 2020-05-06
Using dev channel for common-service and common-service-dev

horis233:
When I test today, I found there would be an issue if we keep both registry `common-service` and `common-service-dev` 

The OperandRequest in cs operators for supporting operator dependency, point to the OperandRegistry common-service, so when users install with the dev registry, it may cause some of the operators are deployed by dev channel and others use the stable channel.
For example. installing dev version MongoDB (from common-service-dev registry) but depends on stable version cert-manager (from common-service registry).

This PR updates the `common-service` registry to use `dev` channel as well to avoid the above conflict. 

I am wondering if we can leave one of the OperandRegiestry and OperandConfig. If users want to install the stable or dev channel, they need to manually update them and keep the name as common-service
For this release, We only keep one pair of the dev registry and config and named them as common-service. For the users who want to install the stable version, they should use the stable registry and config to replace the dev registry and config. I hope to use dev as default for this release is because this is the first release of the ibm common service operator, some of the features like (operandbindinfo and the dependency management) can’t be tested with the stable channel.

When we go GA, we convert the dev registry and config to the stable and in the following release, we use stable version as default and when users want to test, they need to add dev registry and config.

### DIFF
--- a/build/resources/cs-operandconfig.yaml
+++ b/build/resources/cs-operandconfig.yaml
@@ -53,7 +53,6 @@ spec:
   - name: ibm-ingress-nginx-operator
     spec:
       nginxIngress: {}
-      operandRequest: {}
   - name: ibm-auditlogging-operator
     spec:
       auditLogging: {}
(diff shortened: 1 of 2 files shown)


horis233:
I will build multiple channels for the ibm common service operator. close this PR

PR #52, opened 2020-05-12
Update the install doc

DanielXLee:


docs/install.md
```diff
@@ -94,7 +94,7 @@ So far, the IBM Common Service Operator and ODLM operator installation is comple
 
 ## 4.Install Individual Common Services
 
-Install individual common services by creating an OperandRequest.
+Install individual common services by creating an OperandRequest. All these common services will be deployed into namespace `ibm-common-services`, it cannot change.
```
horis233, line 97: All these common services will be deployed into the namespace `ibm-common-services`, which cannot be changed.

PR #54, opened 2020-05-12
Add document for installing with cli

horis233:
This is a requirement from CP4APP to install ibm common service operator from the command line


horis233:
@chenzhiwei @DanielXLee 
CP4App and CP4Auto request an installation document executed by CLI.
Please review.

PR #55, opened 2020-05-12
Change ocp version from 4.2 to 4.3

nyu058:
CS 3.4 does not support OCP4.2, should be change to 4.3


horis233:
@nyu058 This PR is duplicate with #52

PR #62, opened 2020-05-13
remove existing 3.3 ODLM subscription and csv

horis233:


pkg/bootstrap/init.go
```diff
@@ -271,3 +277,41 @@ func createOrUpdateResourcesFromDir(dir string, client client.Client, reader cli
 
 	return nil
 }
+
+func deleteExistingODLM(client client.Client) error {
+	objSub := &unstructured.Unstructured{}
+	objSub.SetGroupVersionKind(schema.GroupVersionKind{Group: "operators.coreos.com", Kind: "Subscription", Version: "v1alpha1"})
+	err := client.Get(context.TODO(), types.NamespacedName{
```
chenzhiwei, line 284: does `client.Get` work for this? This client is cache client, not sure if it can be used here or not.

horis233: It works effectively when I test. It can get and delete the subscription and CSV.

I will investigate if there is further impact. 
https://github.com/kubernetes-sigs/controller-runtime/blob/master/pkg/manager/manager.go#L86

horis233: @chenzhiwei Using `client.Delete` to delete resources directly.

PR #63, opened 2020-05-14
remove the unnecessary copies

chenzhiwei:
Got the message from @liqlin2015 that the `icp-management-ingress-tls-secret` does not need to be copied to `kube-system`.

Sync the secretshare PR: https://github.ibm.com/IBMPrivateCloud/secretshare/pull/2

/cc @mikekaczmarski @liqlin2015


liqlin2015:
Secret icp-management-ingress-tls-secret and route-tls-secret  is not needed anymore.

PR #75, opened 2020-05-15
Add rbac to kube-public

horis233:


chenzhiwei:
please test this manually, and leave a comment if it can work, then I will approve and merge this PR.

horis233:
@chenzhiwei 

It works for me when I install, I will collaborate with CP4I to check if the workaround works for them.

horis233:
@zhiwei The workaround works for CP4I team

renshichuang:
@liqlin2015  @yuwei  I think later we need to move this logic to mgmt ingress operator.  cc @chenzhiwei @horis233

PR #77, opened 2020-05-16
generate odlm in the ibm-common-services namespace

horis233:
Generate ODLM in the ibm-common-services namespace for 3.3.

I have verified it locally.


chenzhiwei:
/hold

Let's hold on this, it does not affect 3.4 release.

PR #83, opened 2020-05-18
Add required RBAC for ibmcloud-cluster-ca-cert

liqlin2015:
Secret ibmcloud-cluster-ca-cert will be copied to kube-public
namespace. But the required RBAC is missing. Add the RBAC so any
k8s user can access the info.


DanielXLee:
Added in PR https://github.com/IBM/ibm-common-service-operator/pull/91

PR #92, opened 2020-05-20
remove odlm csv v1.2.0

chenzhiwei:
/cc @horis233 @DanielXLee


DanielXLee:
to cover the re-install case?

chenzhiwei:
@DanielXLee To cover the normal case, the ODLM will be automatically upgraded to v1.2.0 during upgrade. So there is not v1.1.0 in the cluster.

PR #99, opened 2020-05-25
Updated to add deployment for setting IBM status in a configmap

mikekaczmarski:


DanielXLee:
fixed in https://github.com/IBM/ibm-common-service-operator/pull/100

PR #101, opened 2020-05-28
increase ansible operator cpu/memorty

chenzhiwei:
Seems on Power node, it requires more CPU than x86.


DanielXLee:
you also need to update it into deploy.yaml

PR #104, opened 2020-05-31
remove ODLM csv in ibm-common-services namespace

chenzhiwei:


pkg/bootstrap/init.go
```diff
@@ -315,5 +315,29 @@ func deleteExistingODLM(client client.Client) error {
 		klog.Error("Failed to delete ODLM Cluster Service Version v1.2.0 in the ibm-common-services namespace")
 		return err
 	}
+
```
DanielXLee, line 318: I think we can get the `installedCSV` from the subscription status, if do this, we will not hard code the csv version int the code.

chenzhiwei: I did not know this, could you create a PR about this?

chenzhiwei: ensure that there are not too many changes on this, we are near the release, I don't want to increase the risks.

PR #125, opened 2020-07-02
Enhance logic for delete old ODLM

DanielXLee:
Remove original hard code logic
Current logic:
1. Get ODLM subscription, if don't exist, exit delete method, if existing, continue step 2
2. Get the ODLM CSV name from the subscription of step 1 and delete the CSV
3. Delete ODLM subscription

### DIFF
--- a/pkg/bootstrap/init.go
+++ b/pkg/bootstrap/init.go
@@ -28,11 +28,12 @@ import (
 	"time"
 
 	utilyaml "github.com/ghodss/yaml"
+	olmv1alpha1 "github.com/operator-framework/operator-lifecycle-manager/pkg/api/apis/operators/v1alpha1"
 	"github.com/operator-framework/operator-sdk/pkg/k8sutil"
 	appsv1 "k8s.io/api/apps/v1"
 	"k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
-	"k8s.io/apimachinery/pkg/runtime/schema"
 	"k8s.io/apimachinery/pkg/runtime/serializer/json"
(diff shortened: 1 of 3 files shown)


pkg/bootstrap/init.go
```diff
@@ -285,58 +286,40 @@ func yamlToObject(yamlContent []byte) (*unstructured.Unstructured, error) {
...
+	csvNs := subNs
+	if csvName != "" {
+		csv := &olmv1alpha1.ClusterServiceVersion{
+			ObjectMeta: metav1.ObjectMeta{
+				Name:      csvName,
+				Namespace: csvNs,
+			},
+		}
+		if err := c.Delete(context.TODO(), csv); err != nil && !errors.IsNotFound(err) {
+			klog.Error("Failed to delete ODLM Cluster Service Versio in the ibm-common-services namespace")
```
horis233, line 315: miss an  `n` in the `Cluster Service Version`

pkg/bootstrap/init.go
```diff
@@ -285,58 +286,40 @@ func yamlToObject(yamlContent []byte) (*unstructured.Unstructured, error) {
...
 	}
 
-	// delete csv v1.2.3
-	objCSV.SetName("operand-deployment-lifecycle-manager.v1.2.3")
-	err = client.Delete(context.TODO(), objCSV)
-	if err != nil && !errors.IsNotFound(err) {
-		klog.Error("Failed to delete ODLM Cluster Service Version v1.2.3 in the ibm-common-services namespace")
+	// Delete existing ODLM subscription
+	if err := c.Delete(context.TODO(), sub); err != nil && !errors.IsNotFound(err) {
+		klog.Error("Failed to delete ODLM Cluster Service Versio in the ibm-common-services namespace")
```
horis233, line 322: Same typo as above

pkg/bootstrap/init.go
DanielXLee: nice catch

PR #133, opened 2020-07-27
Create bundle image for the cs operator

DanielXLee:


Makefile
```diff
@@ -1,5 +1,13 @@
-QUAY_REPO ?= quay.io/opencloudio
+IMAGE_REPO ?= quay.io/danielxlee
```
horis233, line 1: Just a reminder that we need to change the repo name back before the merge.

PR #136, opened 2020-07-30
adjust sizing by common service CR

horis233:
I set the priority of deep merge.

common service CR > sizing configurations > OperandConfig

### DIFF
--- a/controllers/commonservice_controller.go
+++ b/controllers/commonservice_controller.go
@@ -18,10 +18,17 @@ package controllers
 
 import (
 	"context"
+	"encoding/json"
+	"reflect"
 
+	utilyaml "github.com/ghodss/yaml"
 	"github.com/go-logr/logr"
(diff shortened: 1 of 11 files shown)


controllers/commonservice_controller.go
```diff
@@ -57,40 +64,143 @@ func (r *CommonServiceReconciler) Reconcile(req ctrl.Request) (ctrl.Result, erro
...
+		return err
+	}
 
-func deepMerge(src []apiv3.ServiceConfig, dest string) []apiv3.ServiceConfig {
 	return nil
 }
 
-func patchOperandConfig(newConfigs []apiv3.ServiceConfig) error {
-	return nil
+func deepMerge(src []interface{}, dest string) []interface{} {
```
Daniel-Fan, line 138: /lgtm cancel
According to the bot test, Should the type of parameter `dest` be one of services instead of string?

PR #146, opened 2020-08-06
integrate the services size

chenzhiwei:


controllers/size/small_amd64.go
```diff
@@ -47,41 +60,462 @@ const Small = `
...
+              cpu: 540m
+              memory: 160Mi
+          routerResource:
+            limits:
+              cpu: 25m
+              memory: 20Mi
+            requests:
+              cpu: 10m
+              memory: 20Mi
+ - name: ibm-monitoring-grafana-operator
```
horis233, line 434: There is a space before  `- name:`

PR #149, opened 2020-08-12
Create cs operator subscription when cs operator doesn't exist in nam…

DanielXLee:
…espace ibm-common-service

### DIFF
--- a/bundle/manifests/ibm-common-service-operator.clusterserviceversion.yaml
+++ b/bundle/manifests/ibm-common-service-operator.clusterserviceversion.yaml
@@ -259,6 +259,20 @@ metadata:
           scope: public
           sourceName: opencloud-operators
           sourceNamespace: openshift-marketplace
+    csOperatorSubscription: |-
+      apiVersion: operators.coreos.com/v1alpha1
+      kind: Subscription
+      metadata:
+        name: ibm-common-service-operator
+        namespace: ibm-common-services
+        annotations:
+          version: "5"
+      spec:
+        channel: dev
+        installPlanApproval: Automatic
+        name: ibm-common-service-operator
+        source: opencloud-operators
+        sourceNamespace: openshift-marketplace
     description: The IBM Common Service Operator is used to deploy IBM Common Services
     extraResources: webhookCRD,webhookOperator,secretShareOperator,rbac
     odlmSubscription: |-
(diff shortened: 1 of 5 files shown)


bundle/manifests/ibm-common-service-operator.clusterserviceversion.yaml
```diff
@@ -489,36 +503,56 @@ metadata:
...
-                value: "secretshare"
-              resources:
-                limits:
-                  cpu: 500m
-                  memory: 512Mi
-                requests:
-                  cpu: 200m
-                  memory: 200Mi
-            terminationGracePeriodSeconds: 10
+              - name: ansible
```
horis233, line 507: @DanielXLee Do we need to change the secretshare and cs webhook in the CSV?

config/manifests/bases/ibm-common-service-operator.clusterserviceversion.yaml
```diff
@@ -467,7 +481,7 @@ metadata:
               productID: 068a62892a1e4db39641342e592daa25
               productMetric: FREE
               productName: IBM Cloud Platform Common Services
-              productVersion: 3.4.0
+              productVersion: 3.5.0
```
horis233, line 484: @DanielXLee ProductVersion would be removed in 3Q.

bundle/manifests/ibm-common-service-operator.clusterserviceversion.yaml
DanielXLee: I don't change any code for secretshare and webhook, This bundle manifests generated by template https://github.com/IBM/ibm-common-service-operator/blob/master/config/manifests/bases/ibm-common-service-operator.clusterserviceversion.yaml, maybe I need to update this template.

config/manifests/bases/ibm-common-service-operator.clusterserviceversion.yaml
DanielXLee: okay

controllers/bootstrap/init.go
```diff
@@ -77,6 +78,30 @@ func InitResources(mgr manager.Manager) error {
 		return err
 	}
 
+	// Checking existing CommonService subscription
```
chenzhiwei, line 81: if ns is not in openshift-operators or ibm-common-services {
  create cs subscription
  return
}

PR #151, opened 2020-08-12
use ubi as the base image

horis233:
- Using go 1.14.6 to solve vulnerabilities CVE-2020-15586 and CVE-2020-14039
- Using ubi as the base image

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,5 +1,5 @@
 # Build the manager binary
-FROM golang:1.13 as builder
+FROM golang:1.14.6 as builder
 
 WORKDIR /workspace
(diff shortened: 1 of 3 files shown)


DanielXLee:
@horis233 if you want to update bundle/manifests/ibm-common-service-operator.clusterserviceversion.yaml, you can update the template https://github.com/IBM/ibm-common-service-operator/blob/master/config/manifests/bases/ibm-common-service-operator.clusterserviceversion.yaml, and the run `make bundle`, it will auto update bundle/manifests/ibm-common-service-operator.clusterserviceversion.yaml

PR #157, opened 2020-08-14
Refact makefile

DanielXLee:


deploy/olm-catalog/ibm-common-service-operator/3.5.0/ibm-common-service-operator.clusterserviceversion.yaml
```diff
@@ -466,14 +466,6 @@ metadata:
         - patch
         - update
         - watch
-      - apiGroups:
```
horis233, line 469: @DanielXLee Please remain this rbac. It is for secretshare.

horis233: I will add it back in the following PR

PR #187, opened 2020-08-20
Refactor common service operator conntroller

DanielXLee:


controllers/commonservice_controller.go
```diff
@@ -209,6 +233,50 @@ func checkKeyBeforeMerging(key string, defaultMap interface{}, changedMap interf
...
+	opreg := common.NewUnstructured("operator.ibm.com", "OperandRegistry", "v1alpha1")
+	opreg.SetName("common-service")
+	opreg.SetNamespace("ibm-common-services")
+	if err := r.Client.Delete(ctx, opreg); err != nil && !errors.IsNotFound(err) {
+		klog.Error("Delete OperandRegistry failed: ", err)
+		return err
+	}
+
+	// Remove ODLM sub and csv
+	if err := common.DeleteExistingODLM(r.Reader, r.Client); err != nil && !errors.IsNotFound(err) {
```
horis233, line 263: `DeleteExistingODLM` seems only delete the odlm in the ibm-common-service namespace. Should we delete it from the `openshift-operators` namespace as well?

DanielXLee: @horis233 For the which resources need to be deleted, I think we need some discuss with @chenzhiwei.
This PR only provide a cleanup framework and UT, after this PR merged, we can add more detail cleanup func into func `finalizeCommonService()`

PR #201, opened 2020-08-25
adding switcheritem to commonui

aniruddhj:


aniruddhj:
@horis233 We have pak switcher item that needs to be installed with the commonui-operator https://github.com/IBM/ibm-commonui-operator/blob/master/deploy/crds/operators.ibm.com_switcheritems_crd.yaml and I am guessing it would be needed to be added to the CSV?

horis233:
@aniruddhj Please update the `OperandConfig` in https://github.com/IBM/ibm-common-service-operator/blob/master/bundle/manifests/ibm-common-service-operator.clusterserviceversion.yaml and https://github.com/IBM/ibm-common-service-operator/blob/master/config/manifests/bases/ibm-common-service-operator.clusterserviceversion.yaml as well.

DanielXLee (approved):
/lgtm

PR #209, opened 2020-08-28
Use medium as the default sizing profile

horis233:
This PR changes the default profile from Small to Medium.


chenzhiwei (commented):
/lgtm

PR #215, opened 2020-09-03
update memory limit from latest x86 measurement

Daniel-Fan:
@horis233 Could you please take a look? Thank you!


horis233 (approved):
/lgtm

PR #216, opened 2020-09-04
init and update size for Z

Daniel-Fan:
init size for Z platform from x86 and update them based on actually usage.


horis233 (approved):
/lgtm

PR #218, opened 2020-09-08
update operand config for Licensing Operator

adamdyszy:
remove metering dependency by deleting operandRequest from Licensing Operator templates and force custom resource change for Licensing to not use metering from deleted dependency


chenzhiwei:
/lgtm

fyi @horis233 @DanielXLee

PR #230, opened 2020-09-13
fix request is over limit for nginxIngress

Daniel-Fan:
fix request is over limit for nginxIngress in small profile on P


horis233 (approved):
/lgtm

PR #251, opened 2020-10-01
remove configmap-watcher

horis233:
Remove configmap-watcher from the IAM deployment list, confirm with IAM team it is a component from the CERT team instead of IAM.


Daniel-Fan (approved):
/lgtm

PR #259, opened 2020-10-14
disable leader election to avoid pod restarting

Daniel-Fan:
remove flag of `enable-leader-election` from command args in order to disable this function.

@chenzhiwei @DanielXLee @horis233 Please correct me if it is wrong.


config/default/manager_auth_proxy_patch.yaml
```diff
@@ -21,5 +21,4 @@ spec:
           name: https
       - name: manager
         args:
-        - "--metrics-addr=127.0.0.1:8080"
-        - "--enable-leader-election"
+        - "--metrics-addr=127.0.0.1:8080"
```
horis233, line 24: @Daniel-Fan Please append a new blank line

bundle/manifests/ibm-common-service-operator.clusterserviceversion.yaml
```diff
@@ -534,8 +534,6 @@ metadata:
             containers:
             - command:
               - /manager
-              args:
```
horis233, line 537: We need to bump the `annotations.version` in the secretshare deployment

horis233 (approved):
/lgtm

PR #261, opened 2020-10-15
Fix lint error

horis233:
Update     `capabilities: Basic Install`	to  `capabilities: Seamless Upgrades`
Update display name to avoid ending up with `operator`.


Daniel-Fan (approved):
/lgtm

PR #263, opened 2020-10-19
push 3.5.5 to stable channel

horis233:
1. push 3.5.5 to the stable-v1 channel
2. revert the ODLM package name, will add it back in the next version


Daniel-Fan (approved):
/lgtm

PR #267, opened 2020-10-29
Creating ODLM depending on install mode

horis233:
1. Replace cluster-admin permission with specific cluster roles.
2. Create different ODLM depends on the ibm-common-service operator install mode.


Daniel-Fan (approved):
/lgtm

PR #278, opened 2020-11-01
create operatorgroup

horis233:
create operatorgroup in `ibm-common-services` when common service operator is deployed in `openshift-operators` namespace

### DIFF
--- a/controllers/bootstrap/init.go
+++ b/controllers/bootstrap/init.go
@@ -127,6 +127,10 @@ func (b *Bootstrap) InitResources() error {
 	// Install ODLM Operator
 	klog.Info("Installing ODLM Operator")
 	if operatorNs == constant.ClusterOperatorNamespace {
+		// Create OperatorGroup when common service operator is in the openshift-operators namespace
+		if err := b.createOperatorGroup(); err != nil {
+			return err
+		}
 		if err := b.createOrUpdateResource(annotations, OdlmClusterSubResource); err != nil {
 			return err
 		}


controllers/bootstrap/init.go
```diff
@@ -127,6 +127,10 @@ func (b *Bootstrap) InitResources() error {
 	// Install ODLM Operator
 	klog.Info("Installing ODLM Operator")
 	if operatorNs == constant.ClusterOperatorNamespace {
+		// Create OperatorGroup when common service operator is in the openshift-operators namespace
+		if err := b.createOperatorGroup(); err != nil {
+			return err
+		}
```
Daniel-Fan, line 133: May I ask that do we need to consider the case that cs-operator was directly installed in `ibm-common-service` ns.
I noticed that we only create `OperatorGroup` in `CreateCsSubscription` when cs-operator is not installed in `ibm-common-service` or `openshift-operators` ns.
https://github.com/IBM/ibm-common-service-operator/blob/012654658b0a4b31bbc1a15f470c101acf117b39/controllers/bootstrap/init.go#L171

horis233: @Daniel-Fan When cs operator is installed in ibm-common-service ns, an operator group will be created when deploying the cs operator.

Daniel-Fan: I got it, when we directly deploy cs-operator in `ibm-common-service` NS, we already create the `operatorgroup`.

PR #283, opened 2020-11-04
Adding navconfiguration to operandrequess

ericabr:
We moved the navconfiguration CRD as being owned by the commonui operator. We need to make sure it is installed with the operand request for commonui.

### DIFF
--- a/bundle/manifests/ibm-common-service-operator.clusterserviceversion.yaml
+++ b/bundle/manifests/ibm-common-service-operator.clusterserviceversion.yaml
@@ -131,6 +131,7 @@ metadata:
             commonWebUI: {}
             switcheritem: {}
             operandRequest: {}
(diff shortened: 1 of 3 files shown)


horis233:
@ericabr please add this change into these two files as well.
https://github.com/IBM/ibm-common-service-operator/blob/master/bundle/manifests/ibm-common-service-operator.clusterserviceversion.yaml
https://github.com/IBM/ibm-common-service-operator/blob/master/deploy/olm-catalog/ibm-common-service-operator/3.6.0/ibm-common-service-operator.v3.6.0.clusterserviceversion.yaml

ericabr:
@horis233 It's updated!

horis233:
@ericabr Thanks, I will merge the PR

horis233 (approved):
/lgtm

PR #295, opened 2020-11-09
42346: remove deprecated helm and catalog-ui operators

cheewaio:
Remove deprecated helm and catalog-ui operators from operandconfig and operandregistry


chenzhiwei:
@cheewaio Could you remove them from `config/manifests/bases/ibm-common-service-operator.clusterserviceversion.yaml`, and run `make generate-all` to generate the final CSVs?

And please also remove them from the document: https://github.com/IBM/ibm-common-service-operator/blob/master/docs/install.md#5install-individual-common-services

I suggest you run a `grep -ri operator-name sourcecode` to find all of them and remove accordingly.

chenzhiwei:
By the way, some of CS operators may add these operators as dependency in their operandrequest, please check it and remove them accordingly.

cheewaio:
@chenzhiwei I did check, there is no other operator in common services that depends on helm and catalog-ui operators.
For other references and test cases in cs-operator, I think I should leave it for your squad to adjust the changes.
Thanks

horis233 (approved):
/lgtm

PR #299, opened 2020-11-12
update digest of secretshare and webhook

Daniel-Fan:


horis233:
@Daniel-Fan Please update the digest in the /testdata as well

horis233 (approved):
/lgtm

PR #308, opened 2020-11-20
Fix upgrade backward compatible issue

DanielXLee:
Upgrade from version 3.4.x, if we don't create namespace-scope configmap, some operators(depend on this configmap) will not upgrade.


horis233 (approved):
/lgtm

PR #311, opened 2020-12-01
update the memory limit for receiverContainer

horis233:
IBMLicenseServiceReporter is a CR that wouldn't be created by default. The memory increase wouldn't impact the release test.


Daniel-Fan (approved):
/lgtm

PR #316, opened 2020-12-04
onboard postgreservice operator

jinguiy:
onboard postgres operator to bedrockservice.

### DIFF
--- a/config/manifests/bases/ibm-common-service-operator.clusterserviceversion.yaml
+++ b/config/manifests/bases/ibm-common-service-operator.clusterserviceversion.yaml
@@ -96,17 +96,7 @@ metadata:
             operandRequest: {}
         - name: ibm-postgreservice-operator
           spec:
-            Postgreservice: 
-              license:
-                accept: false
-              storageClassName: ""
-              cert:
-                caIssuerEnabled: false
-                caIssuer: ""
-                certSecret: ""
-            PostgresDB:
-              license:
-                accept: false
+            Postgreservice: {}
             operandBindInfo: {}
             operandRequest: {}
     csOperandRegistry: |-
--- a/config/manifests/bases/ibm-common-service-operator.clusterserviceversion.yaml
+++ b/config/manifests/bases/ibm-common-service-operator.clusterserviceversion.yaml
@@ -96,7 +96,6 @@ metadata:
             operandRequest: {}
         - name: ibm-postgreservice-operator
           spec:
-            Postgreservice: {}
             operandBindInfo: {}
             operandRequest: {}
     csOperandRegistry: |-
--- a/config/manifests/bases/ibm-common-service-operator.clusterserviceversion.yaml
+++ b/config/manifests/bases/ibm-common-service-operator.clusterserviceversion.yaml
@@ -100,10 +100,10 @@ metadata:
               license:
                 accept: false
               storageClassName: ""
-            cert:
-              caIssuerEnabled: false
-              caIssuer: ""
-              certSecret: ""
+              cert:
+                caIssuerEnabled: false
+                caIssuer: ""
+                certSecret: ""
             PostgresDB:
               license:
                 accept: false
--- a/config/manifests/bases/ibm-common-service-operator.clusterserviceversion.yaml
+++ b/config/manifests/bases/ibm-common-service-operator.clusterserviceversion.yaml
@@ -94,6 +94,21 @@ metadata:
             elasticStack: {}
             operandBindInfo: {}
             operandRequest: {}
+        - name: ibm-postgreservice-operator
+          spec:
+            Postgreservice: 
+              license:
+                accept: false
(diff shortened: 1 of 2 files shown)


DanielXLee:
This operator is not included Q4 release, hold on merge

config/manifests/bases/ibm-common-service-operator.clusterserviceversion.yaml
```diff
@@ -94,6 +94,21 @@ metadata:
             elasticStack: {}
             operandBindInfo: {}
             operandRequest: {}
+        - name: ibm-postgreservice-operator
+          spec:
+            Postgreservice:
```
horis233, line 99: Any reasons we need to add these parameters for CR `Postgreservice` and `PostgresDB`?
The parameters added in the OperandConfig is for overwriting the CRs in the alm-example.
I find the parameters set here are the same as the alm-example.

Could we just add them as the following example?
```
        - name: ibm-postgreservice-operator
          spec:
             Postgreservice: {}
             PostgresDB: {}
```
Moreover, I  don't find `operandBindInfo` and `operandRequest` in the alm-example. They are used to manage secret/configmap sharing and dependency management. You could take iam operator as an example:

operandBindInfo: https://github.com/IBM/ibm-iam-operator/blob/master/deploy/olm-catalog/ibm-iam-operator/3.8.1/ibm-iam-operator.v3.8.1.clusterserviceversion.yaml#L535 
operandrequest: https://github.com/IBM/ibm-iam-operator/blob/master/deploy/olm-catalog/ibm-iam-operator/3.8.1/ibm-iam-operator.v3.8.1.clusterserviceversion.yaml#L507

By the way
```
            Postgreservice: 
              license:
                accept: false
              storageClassName: ""
            cert:
              caIssuerEnabled: false
              caIssuer: ""
              certSecret: ""
```
I believe there should be an indentation before `cert:`

jinguiy: @horis233 
thanks for helping review this pr. 
1. these parameter require user must input a value. For example, license.accept must be defined as true and user must either input one caIssuer or cert secret. In this case, how to define these kind of parameters? 
2. you are checking the csv in ibm-postgreservice-operator repo? It is defined in odlm1 branch, I didn't merge it to master branch yet. 
3. updated the format already.

horis233: @jinguiy 
Just want to confirm. Since ODLM will deploy the CR automatically, is it expected to ask adopters to update the OperandConfig before requesting the postgreservice or need adopters to update the generated CR after CR created?

for the caIssuer, the cs MongoDB creates caIssuer by itself: https://github.com/IBM/ibm-mongodb-operator/blob/master/controllers/certificate.go#L58 or postgreservice can use the caIssuer created by cs cert-manager as default value.

@chenzhiwei @DanielXLee any suggestions?

jinguiy: @horis233 
postgreservice can use the caIssuer created by cs cert-manager as default value, in this case, What is the default caissuer name I can use? Do I need to define cert-manager as the operand request in postgreservice csv? Any sample I can reference about how to define it? 
For the license.accept, bedrock service doesn't require user to accept license before install the operator?  Besides this,  one more question to confirm with you about the bedrock service installation, can user choose to install postgreservice instance and deploy the db instances or not deploy postgreservice instance and only deploy db instances?

horis233: I see the cert-manager has already been added to the operandrequest in the `odlm1` branch. https://github.ibm.com/IBMPrivateCloud/ibm-postgreservice-operator/blob/bee2b288bc776fa2b9a356dbca699b77dd941988/deploy/olm-catalog/ibm-postgreservice-operator/manifests/ibm-postgreservice-operator.clusterserviceversion.yaml#L136

jinguiy: @horis233 I updated pr per the following solution:
1. bedrockservice by default create one postgreservice cr instance for apicatalog which is component in bedrock service and apicatalog create postgresdb instances by operand request way. 
2. For the cloudpacks, they will create both postgreservice cr instance and postgresdb cr instance for their own usage and decided when to create them.

horis233 (commented):
/lgtm

config/manifests/bases/ibm-common-service-operator.clusterserviceversion.yaml
horis233: @jinguiy Thanks for your clarification. It makes sense to me. I will approve and merge it if there is no further change.

jinguiy:
This onboard activity is pending, I will close this pr for now.  If there is requirement to onboard in the future, I will open a new pr.

PR #324, opened 2020-12-13
Support reconciling the common service CR from other namespaces

horis233:
1. Initialize resources only happens when reconciling master CR.
2. Apply different rules for the parameters within the Rules and without Rules

**TODO:** Implement the sizing comparison functions.


Daniel-Fan (approved):
/lgtm

PR #327, opened 2020-12-15
Add status phase

horis233:


controllers/commonservice_controller.go
```diff
@@ -70,22 +79,45 @@ func (r *CommonServiceReconciler) ReconcileMasterCR(req ctrl.Request) (ctrl.Resu
...
+		}
 		return ctrl.Result{}, err
 	}
 
 	newConfigs, err := r.getNewConfigs(req)
 	if err != nil {
+		if err := r.updatePhase(instance, CRFailed); err != nil {
+			klog.Error(err)
+		}
 		return ctrl.Result{}, client.IgnoreNotFound(err)
```
Daniel-Fan, line 110: Did the new err form `updatePhase` override the previous err from `getNewConfigs` in the return?
Would we expect this situation since the first err is not reported in log?

controllers/commonservice_controller.go
```diff
@@ -70,22 +79,45 @@ func (r *CommonServiceReconciler) ReconcileMasterCR(req ctrl.Request) (ctrl.Resu
...
 	}
 
-	if err = r.updateOperandConfig(newConfigs); err != nil {
+	if err := r.updateOperandConfig(newConfigs); err != nil {
+		if err := r.updatePhase(instance, CRFailed); err != nil {
+			klog.Error(err)
+		}
+		return ctrl.Result{}, err
+	}
```
Daniel-Fan, line 118: This is the same as above one.
Did the new err form `updatePhase` override the previous err from `UpdateOperandConfig` in the return?
Would we expect this situation since the first err is not reported in log?

controllers/commonservice_controller.go
```diff
@@ -96,22 +128,52 @@ func (r *CommonServiceReconciler) ReconcileMasterCR(req ctrl.Request) (ctrl.Resu
...
+		}
+		return ctrl.Result{}, err
 	}
 
 	if err = r.updateOperandConfig(newConfigs); err != nil {
+		if err := r.updatePhase(instance, CRFailed); err != nil {
+			klog.Error(err)
+		}
+		return ctrl.Result{}, err
+	}
```
Daniel-Fan, line 174: This is the same as above one.

Would we expect this situation since the first err is not reported in log?

Daniel-Fan (commented):
Please review the above comments.

controllers/commonservice_controller.go
horis233: @Daniel-Fan 
The `err` in the `if` condition wouldn't override the outside `err`. It is a local variable within the if logic.

The out layer `err` is recorded in the reconcile logs.

Daniel-Fan: Get it!
That's a new point I haven't noticed before, thank you!

Daniel-Fan (approved):
/lgtm

PR #339, opened 2020-12-24
get masterNs to replace hardcoded ns

Daniel-Fan:
Need to optimize the method usage and code structure

### DIFF
--- a/controllers/bootstrap/init.go
+++ b/controllers/bootstrap/init.go
@@ -66,15 +66,17 @@ type Bootstrap struct {
 	client.Reader
 	Config *rest.Config
 	*deploy.Manager
+	MasterNamespace string
 }
 
 // NewBootstrap is the way to create a NewBootstrap struct
 func NewBootstrap(mgr manager.Manager) *Bootstrap {
 	return &Bootstrap{
-		Client:  mgr.GetClient(),
-		Reader:  mgr.GetAPIReader(),
-		Config:  mgr.GetConfig(),
-		Manager: deploy.NewDeployManager(mgr),
+		Client:          mgr.GetClient(),
+		Reader:          mgr.GetAPIReader(),
+		Config:          mgr.GetConfig(),
+		Manager:         deploy.NewDeployManager(mgr),
+		MasterNamespace: util.GetMasterNs(mgr.GetAPIReader()),
 	}
 }
 
@@ -95,7 +97,7 @@ func (b *Bootstrap) InitResources(manualManagement bool) error {
 	// Grant cluster-admin to namespace scope operator
 	if operatorNs == constant.ClusterOperatorNamespace {
 		klog.Info("Creating cluster-admin permission RBAC")
-		if err := b.createOrUpdateFromYaml([]byte(util.Namespacelize(constant.ClusterAdminRBAC))); err != nil {
+		if err := b.createOrUpdateFromYaml([]byte(util.Namespacelize(constant.ClusterAdminRBAC, b.MasterNamespace))); err != nil {
 			return err
 		}
 	}
@@ -119,7 +121,7 @@ func (b *Bootstrap) InitResources(manualManagement bool) error {
 	}
 
 	// Create NamespaceScope CR
-	if err := b.createOrUpdateFromYaml([]byte(util.Namespacelize(constant.NamespaceScopeCR))); err != nil {
+	if err := b.createOrUpdateFromYaml([]byte(util.Namespacelize(constant.NamespaceScopeCR, b.MasterNamespace))); err != nil {
 		return err
 	}
 
@@ -131,7 +133,7 @@ func (b *Bootstrap) InitResources(manualManagement bool) error {
 			return err
 		}
 		// Create Operator RBAC
-		if err := b.createOrUpdateFromYaml([]byte(util.Namespacelize(operator.RBAC))); err != nil {
+		if err := b.createOrUpdateFromYaml([]byte(util.Namespacelize(operator.RBAC, b.MasterNamespace))); err != nil {
 			return err
 		}
 		// Create Operator Deployment
@@ -143,7 +145,7 @@ func (b *Bootstrap) InitResources(manualManagement bool) error {
 			return err
 		}
 		// Create Operator CR
-		if err := b.createOrUpdateFromYaml([]byte(util.Namespacelize(operator.CR))); err != nil {
+		if err := b.createOrUpdateFromYaml([]byte(util.Namespacelize(operator.CR, b.MasterNamespace))); err != nil {
 			return err
 		}
 	}
@@ -194,7 +196,7 @@ func (b *Bootstrap) CreateNamespace() error {
 			APIVersion: "v1",
 		},
 		ObjectMeta: metav1.ObjectMeta{
-			Name: constant.MasterNamespace,
+			Name: b.MasterNamespace,
 		},
 	}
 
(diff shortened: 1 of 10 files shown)


controllers/bootstrap/init.go
```diff
@@ -432,3 +480,35 @@ func (b *Bootstrap) deleteSubscription(name, namespace string) error {
...
+func (b *Bootstrap) GetMasterNs() (string, bool) {
+	operatorNs, err := util.GetOperatorNamespace()
+	if err != nil {
+		klog.Errorf("Getting operator namespace failed: %v", err)
+		return "", false
+	}
+
+	cm, _ := b.getCmOfMapCs()
+	if masterNs, found := cm.Data[operatorNs]; found {
+		return masterNs, true
```
horis233, line 511: If the master namespace isn't found in the configmap, we should set the MasterNamespace to `ibm-common-services`

controllers/bootstrap/init.go
```diff
@@ -432,3 +480,35 @@ func (b *Bootstrap) deleteSubscription(name, namespace string) error {
...
+// GetMasterNs gets MasterNamespaces of deploying Common Services
+func (b *Bootstrap) GetMasterNs() (string, bool) {
+	operatorNs, err := util.GetOperatorNamespace()
+	if err != nil {
+		klog.Errorf("Getting operator namespace failed: %v", err)
+		return "", false
+	}
+
+	cm, _ := b.getCmOfMapCs()
+	if masterNs, found := cm.Data[operatorNs]; found {
```
horis233, line 510:
```
data:
common-service-maps:
  - requested-from-namespace: cp4iforFord
    map-common-service-namespace: common-services-ford
  - requested-from-namespace: cp4iforToyota
    map-common-service-namespace: common-services-toyota
  map-default: ibm-common-services
 ```

Please follow the above example to add descriptive keys into the configmap to make it more readable.

Comment (horis233) [reply controllers/bootstrap/init.go:510]: How about
```
data:
  common-service-maps.yaml: |
    namespaceMapping:
    - requested-from-namespace: cp4iforFord
      map-common-service-namespace: common-services-ford
    - requested-from-namespace: cp4iforToyota
      map-common-service-namespace: common-services-toyota
    defaultNs: ibm-common-services
```

Comment (Daniel-Fan) [reply controllers/bootstrap/init.go:510]: @horis233 
It looks good to me. I will follow this format to revise the code

**Comment by Daniel-Fan** [comment]
@horis233 
Have added descriptive keys into the ConfigMap.
It is ready to merge, please review it.

**Comment by horis233** [comment]
@Daniel-Fan Please replace the hardcoded `ibm-common-services` in the CSV files and constant files with the placeholder. Then use method `util.Namespacelize` to replace them to the master namespace

File: controllers/common/util.go [hunk 3]
```diff
@@ -159,6 +176,57 @@ func Reverse(original []string) []string {
...
+	csConfigmap := getCmOfMapCs(r)
+	commonServiceMaps := csConfigmap.Data["common-service-maps.yaml"]
+	var cmData csMaps
+	if err := utilyaml.Unmarshal([]byte(commonServiceMaps), &cmData); err != nil {
+		klog.Errorf("Failed to fetch data of configmap common-service-maps: %v", err)
+		return ""
+	}
+
+	var masterNs string
+	if cmData.DefaultCsNs == "" {
```
horis233, line 218: Do we need to check the `DefaultCsNs`?

I believe if the operator Namespace can't be found from the configmap, the master namespace should be `ibm-common-services`.

`DefaultCsNs` is only used when the webhook replacing the registry namespace in the operandrequest.

Daniel-Fan: Please correct me if I am wrong for the following scenario:

Use the example above,
```
data:
  common-service-maps.yaml: |
    namespaceMapping:
      - requested-from-namespace: cp4iforFord
        map-common-service-namespace: common-services-ford
      - requested-from-namespace: cp4iforToyota
        map-common-service-namespace: common-services-toyota
    defaultCsNs: ibm-common-services
```

The customer would like to specify Common Services installed in `common-services-ford` and `common-services-toyota` for `cp4iforFord` and `cp4iforToyota` respectively.

And there also exists other `cp4iforHonda` and `cp4iforBMW` which customer would like to use the Common Services installed in `DefaultCsNs` called `common-services-cars` for.

In the above case, I think we need to install Common Services in `common-services-cars` and once the operator Namespace can't be found from the configmap, webhook will still replacing registry namespace in the operandrequest by `DefaultCsNs`

horis233: In your case, I believe the data in the configmap should be 
```
data:
  common-service-maps.yaml: |
    namespaceMapping:
      - requested-from-namespace: cp4iforFord
        map-common-service-namespace: common-services-ford
      - requested-from-namespace: cp4iforToyota
        map-common-service-namespace: common-services-toyota
      - requested-from-namespace: cp4iforHonda
        map-common-service-namespace: common-services-cars
      - requested-from-namespace: cp4iforBMW
        map-common-service-namespace: common-services-cars
    defaultCsNs: ibm-common-services
```

From my understanding,  `defaultCsNs` is the registry namespace name that should be replaced by the webhook.

Considering two cases:
1. How about users want to deploy a group of common services in common-services-ford for cp4iforFord and deploy another group of common services in the `ibm-common-services`?
2. How to handle the operand requests generated from CP4MCM, if the webhook replaces all the registry namespaces?

Daniel-Fan: Just Pin me on slack anytime to discuss this.

Original--------------------------------------------------------
Sorry, I made some mistakes on configmap in the example.

And I understand that we hope to minimize the perception for our adopters which means the registry namespace in operand request will usually be `ibm-common-services`. Please correct me if I am wrong here.

I agree that it makes sense that the configmap would point out the requestedNs as `common-services-cars` for both Honda and BMW.
and if there is a new service installed for example CP4MCM, and user did not update the configmap, the group of common services will be deployed into `ibm-common-services`. But what if they want all other services use Common services in `common-services-cars`?

/*
Therefore, there is a tradeoff that if user want to set defaultNs to `common-services-cars`, it will conflict with minimizing perception for adopters which means operand requests created by them will have registry namespace `common-services-cars`.
*/

From my perspective, if we are able to let adopters make changes on registry namespace. I wonder that whether there is a way to simplify the configmap like this
```
data:
  common-service-maps.yaml: |
    namespaceMapping:
      - requested-from-namespace: cp4iforFord
        map-common-service-namespace: common-services-ford
      - requested-from-namespace: cp4iforToyota
        map-common-service-namespace: common-services-toyota
    defaultCsNs: common-services-cars
```

For the two cases you mentioned,

ConfigMap for first case: `common-services-ford` for `cp4iforFord`, and others will remain in `ibm-common-services`(it could be changed to `common-services-cars`)
```
data:
  common-service-maps.yaml: |
    namespaceMapping:
      - requested-from-namespace: cp4iforFord
        map-common-service-namespace: common-services-ford
    defaultCsNs: ibm-common-services #common-services-cars
```

ConfigMap for second case: All the operand request without `opreq.Namespace` `cp4iforFord` and `cp4mcmforToyota` will remain `registryNs` at `ibm-common-services` or change `registryNs` to `common-services-cars`.  
```
data:
  common-service-maps.yaml: |
    namespaceMapping:
      - requested-from-namespace: cp4iforFord
        map-common-service-namespace: common-services-ford
      - requested-from-namespace: cp4mcmforToyota
        map-common-service-namespace: common-services-Toyota
    defaultCsNs: ibm-common-services #common-services-cars
```

horis233 (approved):
/lgtm

Issue #345, opened 2021-01-02
Permission error when webhook tries to fetch configmap in kube-public


Daniel-Fan:
```
admission webhook "ibm-cloudpak-operandrequest.operator.ibm.com" denied the request: failed to fetch configmap kube-public/common-service-maps: configmaps "common-service-maps" is forbidden: User "system:serviceaccount:common-service-installed:ibm-common-service-webhook" cannot get resource "configmaps" in API group "" in the namespace "kube-public"
```

I will submit a PR to add the corresponding items.

/cc @horis233

Daniel-Fan:
implemented, close the issue.

PR #346, opened 2021-01-02
add permission for webhook to fetch configmap

Daniel-Fan:
fix issue #345 

verb `get` is enough for webhook to fetch configmap in `kube-public` ns


horis233 (approved):
/lgtm

PR #347, opened 2021-01-04
remove finalizer

### LINKED ISSUE #343: Unable to delete commonservice CR due to the existence of finalizer
When I try to uninstall the Common Service, the CR commonservice in cs-operator cannot be deleted because there exists finalizer.

Need to use patching to remove finalizer

/cc @horis233

horis233:
fix #343

Remove the finalizer logic to make CRs being deleted easily.


Daniel-Fan (approved):
/lgtm

PR #351, opened 2021-01-05
cover the case configmap doesn't exist

horis233:
/assign @Daniel-Fan 

Solve the defect when the configmap doesn't exist.


Daniel-Fan (approved):
/lgtm

PR #358, opened 2021-01-08
Convert digitals to string, then compare their size

### LINKED ISSUE #357: Several replicas don't shrink when update the size

horis233:
Fix #357


Daniel-Fan (approved):
/lgtm

PR #359, opened 2021-01-09
add storageclass in spec

horis233:
Add a common entrypoint of the storage class should be set in the common-service CR, like
```
kind: CommonService
metadata:
  name: common-service
  namespace: ibm-common-services
spec:
  storageClass: cephfs
```

bump common service version


Daniel-Fan (approved):
/lgtm

PR #365, opened 2021-01-12
update cs mapping configmap structure

horis233:
- Update the cs mapping configmap structure 
```
common-service-maps.yaml:
- requested-from-namespaces:
  - cp4iforFord
  map-to-common-service-namespace: common-service-ford
map-default: ibm-common-servies
```
- Update image digest for cs webhook


Daniel-Fan (approved):
/lgtm

PR #368, opened 2021-01-14
expose images as environmental variables

horis233:
**Let's hold this PR until the beta release**

This PR is to leverage the CICD image replace function, then we don't need to replace webhook and secretshare images manually each time.


horis233:
@Daniel-Fan @bitscuit Please review this PR. It is ready to be merged

Daniel-Fan (approved):
/lgtm

PR #370, opened 2021-01-18
switch 3.7 to v3 channel

Daniel-Fan:
1. Remove the `replace` section as it is the origin of `v3` channel.

2. Keep `olm.skipRange: '>=3.3.0 <3.7.0'` as it supports upgrading from the version smaller than `3.7.0`.

3. ~~Update the channel to `v3` for the subscription and the operator in the `opencloud-operators` CatalogSource.~~
Currently, we did not update the channel of subscription and operator in `OperandRegistry` in `beta` stage, leaving the task to CICD pipeline.

4. Did not update the version in annotation since there is no release published yet.

### DIFF
--- a/bundle/manifests/ibm-common-service-operator.clusterserviceversion.yaml
+++ b/bundle/manifests/ibm-common-service-operator.clusterserviceversion.yaml
@@ -748,5 +748,4 @@ spec:
   maturity: alpha
   provider:
     name: IBM
-  replaces: ibm-common-service-operator.v3.6.2
   version: 3.7.0
--- a/config/manifests/bases/ibm-common-service-operator.clusterserviceversion.yaml
+++ b/config/manifests/bases/ibm-common-service-operator.clusterserviceversion.yaml
@@ -550,5 +550,4 @@ spec:
   maturity: alpha
   provider:
     name: IBM
-  replaces: ibm-common-service-operator.v3.6.2
   version: 0.0.0


bundle/manifests/ibm-common-service-operator.clusterserviceversion.yaml
```diff
@@ -112,95 +112,95 @@ metadata:
         operators:
         - name: ibm-licensing-operator
           namespace: placeholder
-          channel: beta
+          channel: v3
```
horis233, line 115: @Daniel-Fan We don't need to update channels in this PR, v3 will be a release channel. I believe we still need to use the `beta` channel for the dev purpose.

horis233: I can update channels and catalogsource in another PR.

Daniel-Fan: Could I understand that we only need to update the channel to `v3` in subscription of operator owned by installer?

horis233: My concern is if we update the channel to v3, it will break the CICD daily build.
My understanding is we just remove the `replace` in the CSV and leave the channel updating works to CICD when they build catalog.

Daniel-Fan: @horis233 Please review again, we only remove the `replace` section and keep `olm.skipRange: '>=3.3.0 <3.7.0'`

horis233 (approved):
/lgtm

PR #377, opened 2021-01-23
add event to remind no update although CR changed

Daniel-Fan:
Working in testing

Is there any advice on the logic, the current method looks complicated and unclear

### DIFF
--- a/controllers/commonservice_controller.go
+++ b/controllers/commonservice_controller.go
@@ -18,10 +18,13 @@ package controllers
 
 import (
 	"context"
+	"fmt"
 
+	corev1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/types"
+	"k8s.io/client-go/tools/record"
 	"k8s.io/klog"
 	ctrl "sigs.k8s.io/controller-runtime"
 	"sigs.k8s.io/controller-runtime/pkg/builder"
@@ -40,7 +43,8 @@ type CommonServiceReconciler struct {
 	client.Reader
 	*deploy.Manager
 	*bootstrap.Bootstrap
-	Scheme *runtime.Scheme
+	Scheme   *runtime.Scheme
+	Recorder record.EventRecorder
 }
 
 const (
@@ -64,6 +68,7 @@ func (r *CommonServiceReconciler) Reconcile(req ctrl.Request) (ctrl.Result, erro
 			if err := r.handleDelete(); err != nil {
 				return ctrl.Result{}, err
 			}
+			klog.Info("Deleted reconciling CommonService CR successfully")
 		}
 		return ctrl.Result{}, client.IgnoreNotFound(err)
 	}
@@ -117,12 +122,18 @@ func (r *CommonServiceReconciler) ReconcileMasterCR(instance *apiv3.CommonServic
 		return ctrl.Result{}, client.IgnoreNotFound(err)
 	}
 
-	if err := r.updateOperandConfig(newConfigs); err != nil {
+	isEqual, err := r.updateOperandConfig(newConfigs)
+	if err != nil {
 		if err := r.updatePhase(instance, CRFailed); err != nil {
 			klog.Error(err)
 		}
 		klog.Errorf("Fail to reconcile %s/%s: %v", instance.Namespace, instance.Name, err)
 		return ctrl.Result{}, err
+	}
+
+	// Create Event if there is no update in OperandConfig after applying current CR
+	if isEqual {
+		r.Recorder.Event(instance, corev1.EventTypeNormal, "Noeffect", fmt.Sprintf("No update, resource sizings in the OperandConfig %s/%s are larger than the profile from CommonService CR %s/%s", r.Bootstrap.MasterNamespace, "common-service", instance.Namespace, instance.Name))
 	}
 
 	if err := r.updatePhase(instance, CRSucceeded); err != nil {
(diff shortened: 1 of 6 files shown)


Daniel-Fan:
Get a blocker at fetching the `ibm-common-service-operator` pod

usually we will use `Client.Get()` to get the resource, for example
```
csPodName := ibm-common-service-operator
r.Bootstrap.MasterNamespace := ibm-common-services

csPod := &corev1.Pod{}
	if err := r.Client.Get(ctx, types.NamespacedName{Name: csPodName, Namespace: r.Bootstrap.MasterNamespace}, csPod); err != nil {
		klog.Errorf("Fail to get pod %s/%s: %v", r.Bootstrap.MasterNamespace, csPodName, err)
		return ctrl.Result{}, err
	}
```
We can't get the pod since the name of pod is `ibm-common-service-operator-b69776c9b-sdzdq` with arbitrary suffix.

Any help will be greatful : )

horis233:
@Daniel-Fan Why do we need to get the ibm common service operator pod?

Daniel-Fan:
@horis233 
I thought we should create the event to the IBM-common-service-operator pod once there is no update. But it will be much easier if we choose to have the event in the instance of CommonService CR.

controllers/commonservice_controller.go
```diff
@@ -117,14 +124,19 @@ func (r *CommonServiceReconciler) ReconcileMasterCR(instance *apiv3.CommonServic
...
+	if err != nil {
 		if err := r.updatePhase(instance, CRFailed); err != nil {
 			klog.Error(err)
 		}
 		klog.Errorf("Fail to reconcile %s/%s: %v", instance.Namespace, instance.Name, err)
 		return ctrl.Result{}, err
 	}
 
+	if !isChanged {
+		r.Recorder.Event(instance, "Normal", "Noeffect", fmt.Sprintf("No update, existing CommonService CR has larger profile %s/%s", instance.Namespace, instance.Name))
```
horis233, line 137: This message is not very clear to me.
I assume it should be "No update, resource sizings in the OperandConfig %s/%s are larger than the profile from CommonService CR %s/%s"

horis233:
@Daniel-Fan 
> I thought we should create the event to the IBM-common-service-operator pod once there is no update. But it will be much easier if we choose to have the event in the instance of CommonService CR.

No, we just need to update the event for the CommonService CR instead of the operator pod. The operator can only add events to the resource it manages. The pod is managed by k8s development controller. We shouldn't append events for it.

controllers/operandconfig.go
```diff
@@ -241,29 +250,30 @@ func (r *CommonServiceReconciler) updateOperandConfig(newConfigs []interface{})
...
+		return false, err
 	}
 
 	opcon.Object["spec"].(map[string]interface{})["services"] = opconServices
 
 	if err := r.Update(ctx, opcon); err != nil {
 		klog.Errorf("failed to update OperandConfig %s: %v", opconKey.String(), err)
-		return err
+		return false, err
 	}
```
horis233, line 275: I believe we can add the event here instead of in the reconcile functions.

controllers/operandconfig.go
```diff
@@ -241,29 +250,30 @@ func (r *CommonServiceReconciler) updateOperandConfig(newConfigs []interface{})
...
 
 	// Checking all the common service CRs to get the minimal size
-	opconServices, err = r.getMinimalSizes(opconServices, ruleSlice)
+	opconServices, _, err = r.getMinimalSizes(opconServices, ruleSlice)
 	if err != nil {
-		return err
+		return false, err
 	}
 
 	opcon.Object["spec"].(map[string]interface{})["services"] = opconServices
```
horis233, line 269: @Daniel-Fan Could we just make a comparison here? if the services don't change then report an event. Then we don't need to pass the ischange button up.

controllers/operandconfig.go
```diff
@@ -241,29 +250,30 @@ func (r *CommonServiceReconciler) updateOperandConfig(newConfigs []interface{})
 			}
 			if rules != nil && rules.(map[string]interface{})["spec"] != nil && rules.(map[string]interface{})["spec"].(map[string]interface{})[cr] != nil {
 				ruleForCR := rules.(map[string]interface{})["spec"].(map[string]interface{})[cr].(map[string]interface{})
-				opService.(map[string]interface{})["spec"].(map[string]interface{})[cr] = mergeCRsIntoOperandConfig(spec.(map[string]interface{}), newConfigForCR, ruleForCR, overwrite)
+				opService.(map[string]interface{})["spec"].(map[string]interface{})[cr], isChangedWithRules = mergeCRsIntoOperandConfig(spec.(map[string]interface{}), newConfigForCR, ruleForCR, overwrite)
 			} else {
 				if overwrite {
```
bitscuit, line 255: does this overwrite variable represent whether OperandConfig values should be overwritten? If so, can we not just return this variable to indicate whether there have been updates?

bitscuit:
Will need to generalize this a bit more for when storageClass changes are merged in

Daniel-Fan:
Yeah, I will rewrite the code to adopt the addition of storage class

controllers/operandconfig.go
Daniel-Fan: The `r.Recorder.Event` will require instance of  `apiv3.CommonService`.

If we add event in `operandconfig.go`, should we pass the instance as the parameter into the `updateOperandConfig` function

controllers/operandconfig.go
Daniel-Fan: The simple DeepCopy and DeepEqual did not work in this case.

Will try a loop to iterate the `opconServices` for the comparison.

controllers/operandconfig.go
```diff
@@ -260,17 +263,30 @@ func (r *CommonServiceReconciler) updateOperandConfig(newConfigs []interface{})
...
+	isEqual := true
+
+	for _, opService := range opconServices {
+		existingOpService := getItemByName(existingOpconServices.([]interface{}), opService.(map[string]interface{})["name"].(string))
+		for cr, spec := range opService.(map[string]interface{})["spec"].(map[string]interface{}) {
+			existingCr := existingOpService.(map[string]interface{})["spec"].(map[string]interface{})[cr].(map[string]interface{})
+			if isEqual = reflect.DeepEqual(spec, existingCr); !isEqual {
+				break
+			}
+		}
```
Daniel-Fan, line 279: @horis233 
After do a `DeepEqual` for each individual CR's spec, it could correctly handle the size switch between large and medium. Also it is working fine when customizing the sizes of individual CRs.

But once the small profile is involved, the `isEqual` will always be `false`.

horis233: Interesting...

Daniel-Fan:
@horis233 @bitscuit 
This PR is ready for reviewing.

horis233 (approved):
/lgtm

PR #379, opened 2021-01-25
added storageClass to CRD

bitscuit:
Needed to generalize `mergeChangedMap` function to handle values which were not resource values, and handle cases where user explicitly sets value to null.

### DIFF
--- a/controllers/operandconfig.go
+++ b/controllers/operandconfig.go
@@ -141,8 +141,18 @@ func mergeChangedMap(key string, defaultMap interface{}, changedMap interface{},
 			//Check if the value was set, otherwise set it
 			if changedMap == nil {
 				finalMap[key] = defaultMap
+			} else if defaultMap == nil {
+				// need to handle case where user explicitly set value to null
+				finalMap[key] = changedMap
 			} else {
-				finalMap[key], _ = rules.ResourceComparison(defaultMap, changedMap)
+				comparableKeys := []string{"replicas", "cpu", "memory"}
+				for _, comparable := range comparableKeys {
+					if key == comparable {
+						finalMap[key], _ = rules.ResourceComparison(defaultMap, changedMap)
+					} else {
+						finalMap[key] = changedMap
+					}
+				}
 			}
 		}
 	}
--- a/controllers/operandconfig.go
+++ b/controllers/operandconfig.go
@@ -141,9 +141,6 @@ func mergeChangedMap(key string, defaultMap interface{}, changedMap interface{},
 			//Check if the value was set, otherwise set it
 			if changedMap == nil {
 				finalMap[key] = defaultMap
-			} else if defaultMap == nil {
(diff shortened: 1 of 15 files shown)


controllers/operandconfig.go
```diff
@@ -141,8 +141,18 @@ func mergeChangedMap(key string, defaultMap interface{}, changedMap interface{},
 			//Check if the value was set, otherwise set it
 			if changedMap == nil {
 				finalMap[key] = defaultMap
+			} else if defaultMap == nil {
+				// need to handle case where user explicitly set value to null
+				finalMap[key] = changedMap
```
horis233, line 146: @bitscuit Could you explain why this is required?

From my understanding, finalMap[key] is always equal to changedMap.

bitscuit: good catch. Will remove it. It was there initially to solve an issue, but the comparable key check solves it as well.

bitscuit:
Added missing elastic-stack-operator configuration.

controllers/operandconfig.go
```diff
@@ -142,7 +142,14 @@ func mergeChangedMap(key string, defaultMap interface{}, changedMap interface{},
 			if changedMap == nil {
 				finalMap[key] = defaultMap
 			} else {
-				finalMap[key], _ = rules.ResourceComparison(defaultMap, changedMap)
+				comparableKeys := []string{"replicas", "cpu", "memory"}
+				for _, comparable := range comparableKeys {
```
bitscuit, line 146: not sure if we want to do the comparable key check like this. If we add more configurations in the future, and squads determine they can be changed after deployment, then comparing this way will not scale well.

horis233:
> Added missing elastic-stack-operator configuration.

elastic-stack-operator has been deprecated, we don't need to add it.

controllers/operandconfig.go
horis233: I think we can add the comparable key check since the ResourceComparison is for resource size comparison.

bitscuit: right, but I mean in the future, if we add any more configurations, e.g. maybe operators include a new `healthcheck-interval` configuration that is the number of seconds between pinging a health check endpoint, and we want to add this to CS CR as well, then this new variable could be comparable (maybe we want to take the lowest value).

The way this code is checking for comparable keys won't scale very well if we add more configurations like this in the future. We would also need to keep updating the list every time. Should we consider this? Or is looping over a list of keys good enough?

horis233: I think we keep them for now. The price to maintain the list is not high for us. 
Considering the case you list, we need to update the ResourceConparison function anyway without the comparable keys.

Daniel-Fan (approved):
@horis233 
I have OK with the PR. Could you please review it again?

horis233 (approved):
/lgtm

PR #384, opened 2021-01-27
add script for moving to ibm-catalogsource

Daniel-Fan:
To use the script in beta stage, we need to replace `ibm-operator-catalog` by dev catalogsource `opencloud-operators`

```
oc patch sub ${cssub} -n ${ns} --type="json" -p '[{"op": "replace", "path":"/spec/source", "value":"ibm-operator-catalog"}]' 2> /dev/null
```

### DIFF
--- a/common/scripts/migrate-to-ibm-catalog.sh
+++ b/common/scripts/migrate-to-ibm-catalog.sh
@@ -1,0 +1,92 @@
+#!/bin/bash
+#
+# Copyright 2021 IBM Corporation
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+# http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+
+# counter to keep track of installation steps
+STEP=0
+
+# script base directory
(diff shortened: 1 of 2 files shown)


common/scripts/utils.sh
```diff
@@ -0,0 +1,51 @@
...
+function error() {
+    msg "\33[31m[✘] ${1}\33[0m"
+    exit 1
+}
+
+function title() {
+    msg "\33[34m# ${1}\33[0m"
+}
+
+function check_command() {
```
bitscuit, line 43: this function isn't being used anywhere

common/scripts/migrate-to-ibm-catalog.sh
```diff
@@ -0,0 +1,98 @@
...
+
+    title "[${STEP}] Switch to IBM Operator Catalog Source ..."
+    msg "-----------------------------------------------------------------------"
+
+    declare -A cssubs
+    while read -r ns cssub; do
+        cssubs["$ns"]="$cssub";
+    done < <(oc get sub --all-namespaces | grep ibm-common-service-operator | awk '{print $1" "$2}')
+
+    for ns in "${!cssubs[@]}"; do
```
bitscuit, line 74: can't this for loop be removed, and then patching steps be done in the while loop?

common/scripts/utils.sh
Daniel-Fan: Nice catch, I will remove it.

common/scripts/migrate-to-ibm-catalog.sh
Daniel-Fan: Yeah, it makes sense! And after removing the for loop, we don't even need to declare the array.

Daniel-Fan:
Update the script, removing the redundant loop and function.

bitscuit (approved):
/lgtm

PR #389, opened 2021-02-02
create secretshare CR

horis233:


controllers/bootstrap/init.go
```diff
@@ -420,6 +421,33 @@ func (b *Bootstrap) CreateNsScopeConfigmap() error {
 	return nil
 }
 
+// CreateNsScopeConfigmap creates nss configmap for operators
```
bitscuit, line 424: comment does not match function name nor purpose

horis233: Nice catch!

horis233: It is updated

bitscuit (approved):
/lgtm

PR #390, opened 2021-02-02
feat(tmp): Add openshift-redhat-marketplace in nss CR.

horis233:
We can remove it after we can input the namespace from OperandRegistry into the nss CR.

### DIFF
--- a/controllers/constant/namespace.go
+++ b/controllers/constant/namespace.go
@@ -26,6 +26,7 @@ metadata:
 spec:
   namespaceMembers:
   - placeholder
+  - openshift-redhat-marketplace
 ---
 apiVersion: operator.ibm.com/v1
 kind: NamespaceScope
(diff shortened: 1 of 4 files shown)


Daniel-Fan (approved):
/lgtm

pgodowski:
thanks. do we want also to add `kube-system` same way?

controllers/constant/namespace.go
```diff
@@ -26,6 +26,7 @@ metadata:
 spec:
   namespaceMembers:
   - placeholder
+  - openshift-redhat-marketplace
```
pgodowski, line 29: do we want to add `kube-system` same way?

horis233:
@pgodowski 
We can add `kube-system` in 1Q and remove them after we can get namespaces from the OperandRegistry in the fixpack.

PR #397, opened 2021-02-10
Check the status of operator before upgrading

horis233:


controllers/bootstrap/init.go
```diff
@@ -457,3 +462,39 @@ func (b *Bootstrap) deleteSubscription(name, namespace string) error {
...
+		csvName := sub.Status.InstalledCSV
+		if csvName != "" {
+			csv := &olmv1alpha1.ClusterServiceVersion{}
+			if err := b.Client.Get(context.TODO(), types.NamespacedName{Namespace: namespace, Name: csvName}, csv); err != nil && !errors.IsNotFound(err) {
+				klog.Errorf("Failed to get Cluster Service Version: %v", err)
+				return false, err
+			} else if errors.IsNotFound(err) {
+				klog.Errorf("Notfound Cluster Service Version: %v", err)
+				return false, nil
+			}
```
Daniel-Fan, line 490: Could we use the same if and else structure as the above line 471-476 for a more clear view?

controllers/bootstrap/init.go
```diff
@@ -457,3 +462,39 @@ func (b *Bootstrap) deleteSubscription(name, namespace string) error {
 
 	return nil
 }
+
+func (b *Bootstrap) waitOperatorReady(name, namespace string) error {
+	if err := utilwait.PollImmediate(time.Second*10, time.Minute*10, func() (done bool, err error) {
```
Daniel-Fan, line 467: Do we need to use named return value for this anonymous function?

controllers/bootstrap/init.go
horis233: I will update them to 
```
			if err := b.Client.Get(context.TODO(), types.NamespacedName{Namespace: namespace, Name: csvName}, csv); errors.IsNotFound(err) {
				klog.Errorf("Notfound Cluster Service Version: %v", err)
				return false, nil
			} else if err != nil {
				klog.Errorf("Failed to get Cluster Service Version: %v", err)
				return false, err
			}
```

I believe it is clearer.

controllers/bootstrap/init.go
horis233: I want me to be clear about the meaning of this Boolean value.

horis233: Otherwise, I believe readers may be confused about if the boolean means `continue` or `finish`.

controllers/bootstrap/init.go
horis233: done

controllers/bootstrap/init.go
Daniel-Fan: Get it!

Daniel-Fan (approved):
/lgtm

PR #402, opened 2021-02-12
support multiple ns for upgrading to 3.7

Daniel-Fan:
Remove step of updating catalogsource

to upgrade the operator in specific namespace. please add namespaces as the command line arguments.
```
./upgrade_to_continuous_delivery.sh common-service-2 common-service-3
```


bitscuit (approved):
/lgtm

PR #403, opened 2021-02-15
updated channel switch messages

pgodowski:
minor updates to the user facing messages


davesoroka:
Those strings look like hard coded English strings.  Are these exempt from translation?

pgodowski:
@horis233 do we have any g11n method for this script?

horis233:
@pgodowski @davesoroka We don't have g11n method for this script.
If we need to support multiple languages, we can ask ID to provide translations. When using the script, users can specify the language with a parameter like `--LANG=zh_CN`.

davesoroka:
it's not something for us to fix for this release.  But we should talk through translation requirements for the next release.

horis233 (approved):
/lgtm

PR #405, opened 2021-02-17
Waiting for cert-manager upgrade then update the operandconfig

horis233:


controllers/bootstrap/init.go
```diff
@@ -212,7 +220,15 @@ func (b *Bootstrap) InitResources(manualManagement bool) error {
 		return err
 	}
 
-	if err := b.createOrUpdateResources(annotations, OdlmCrResources); err != nil {
+	if err := b.createOrUpdateResources(annotations, RegistryCrResources); err != nil {
```
Daniel-Fan, line 223: Do we need to remain the second parameter as the string slice in this case?
or we could use func `createOrUpdateResource` if it is only the string.

controllers/bootstrap/init.go
```diff
@@ -497,13 +531,53 @@ func (b *Bootstrap) waitOperatorReady(name, namespace string) error {
...
 			}
-			if csv.Status.Phase == olmv1alpha1.CSVPhaseSucceeded {
-				return true, nil
+			if csv.Status.Phase != olmv1alpha1.CSVPhaseSucceeded {
+				return false, nil
+			}
+			if csv.Status.Reason != olmv1alpha1.CSVReasonInstallSuccessful {
+				return false, nil
 			}
+			klog.Errorf("Cluster Service Version %s/%s is ready", csv.Namespace, csv.Name)
```
Daniel-Fan, line 540: Should we use `klog.infof` instead of `errorf`?

controllers/bootstrap/init.go
horis233: Nice catch, updated

controllers/bootstrap/init.go
horis233: Nice catch, updated

Daniel-Fan (approved):
/lgtm

PR #412, opened 2021-02-22
hold to switch channel until upgrading to eus

Daniel-Fan:
Currently, check all csv of operators to be in 3.6 version and `succeeded` status, then switch the channel.

Did not add the step for checking the version of OperandConfig, will discuss whether it is necessary in sync up.


horis233 (approved):
/lgtm

PR #413, opened 2021-02-22
Update admin role

horis233:


controllers/bootstrap/init.go
```diff
@@ -498,7 +498,14 @@ func (b *Bootstrap) waitOperatorReady(name, namespace string) error {
 			if sub.Spec.Channel != "v3" {
 				return false, nil
 			}
-			csvVersion := strings.Split(sub.Status.InstalledCSV, ".v")[1]
+			if sub.Status.InstalledCSV == ""{
```
Daniel-Fan, line 501: @horis233 I would guess we miss a space before the bracket that causes the test failed?

Daniel-Fan (approved):
/lgtm

PR #415, opened 2021-02-22
add upgrade script for airgap

Daniel-Fan:
For the airgap, we would switch channel first and then update catalogsource


horis233 (approved):
/lgtm

PR #416, opened 2021-02-22
update upgrade script

horis233:


common/scripts/upgrade_to_continuous_delivery_airgap.sh
```diff
@@ -77,6 +80,25 @@ function switch_to_continous_delivery() {
     success "Updated all ibm-common-service-operator subscriptions successfully."
     msg ""
 
+    odlmsub=$(oc get sub operand-deployment-lifecycle-manager-app -n openshift-operators --ignore-not-found)
```
Daniel-Fan, line 83: I think we should consider the upgrade from 3.6 to 3.7 for namespace-scope ODLM.

could use
```
# update subscription of ODLM and cs-operator in all namespaces
    while read -r ns cssub; do

        msg "Updating subscription ${cssub} in namespace ${ns}..."
        msg "-----------------------------------------------------------------------"
        
        in_step=1
        msg "[${in_step}] Removing the startingCSV ..."
        oc patch sub ${cssub} -n ${ns} --type="json" -p '[{"op": "remove", "path":"/spec/startingCSV"}]' 2> /dev/null

        in_step=$((in_step + 1))
        msg "[${in_step}] Switching channel from stable-v1 to v3 ..."
        oc patch sub ${cssub} -n ${ns} --type="json" -p '[{"op": "replace", "path":"/spec/channel", "value":"v3"}]' 2> /dev/null

        msg ""
    done < <(oc get sub --all-namespaces | grep operand-deployment-lifecycle-manager-app | awk '{print $1" "$2}')
```

Daniel-Fan (approved):
/lgtm

PR #423, opened 2021-02-24
update script to be compatible with bash 3

Daniel-Fan:
The previous associative array are not supported in bash version 3 which is mac default version


horis233 (approved):

horis233 (commented):
/lgtm

PR #424, opened 2021-02-25
Enhance iam status check

horis233:
This PR is for resetting the status to `NotReady` after deleting IAM operator


Daniel-Fan (approved):
/lgtm

PR #426, opened 2021-03-02
update IAM config when there is saas configmap

horis233:


controllers/render_template.go
```diff
@@ -96,3 +115,24 @@ func applySizeTemplate(cs *unstructured.Unstructured, sizeTemplate string) ([]in
...
+		return false, nil
+	} else if err != nil {
+		return false, err
+	}
+	v, ok := saasConfigmap.Data["ibm_cloud_saas"]
+	if !ok {
+		return false, nil
+	}
+	if v != "true" {
+		klog.Info(v)
```
Daniel-Fan, line 134: is the `klog.Info(v)` too unclear as a information message, i assume that `v` is just `false`?

horis233: Sorry, it is for testing. I forget to remove it.

Daniel-Fan (approved):
/lgtm

PR #430, opened 2021-03-05
scale up power and z limits

Daniel-Fan:


horis233 (approved):
/lgtm

Daniel-Fan:
Confirmed with performance squad, those changes are based on 3.7 version. Do not need to pick up by EUS version.

PR #431, opened 2021-03-05
added more automation to make testing and version updates easier

bitscuit:
added make targets for deploying operator in bundle format to avoid needing to set up CatalogSource for quicker testing

removed OPERATOR_VERSION from Makefile since it should the the same as
RELEASE_VERSION. When changing this operator's version, only need to
update version.go file and PREVIOUS_VERSION variable, and then run `make
bundle-manifests`

built CSV using operator-sdk v1.3.0


bitscuit:
wait until 1Q is released before merging to avoid any issues

bitscuit:
still missing a way to fully automate version updates because the README and `.spec.description` in CSV reference the operator version. Not sure if there is an easy way to edit those.

bundle/manifests/ibm-common-service-operator.clusterserviceversion.yaml
```diff
@@ -318,20 +318,19 @@ spec:
...
-      - description: The size of deployment profile for IBM Cloud Pak foundational services
-        displayName: Size
-        type: string
-        path: size
-        x-descriptors:
-          - 'urn:alm:descriptor:com.tectonic.ui:select:small'
-          - 'urn:alm:descriptor:com.tectonic.ui:select:medium'
-          - 'urn:alm:descriptor:com.tectonic.ui:select:large'
+      - description: CommonService is the Schema for the commonservices API
+        displayName: Common Service
```
horis233, line 322: I believe we need to use `displayName: CommonService`

bundle/manifests/ibm-common-service-operator.clusterserviceversion.yaml
```diff
@@ -302,8 +302,8 @@ metadata:
           env:
           - name: INSTALL_SCOPE
             value: namespaced
-    olm.skipRange: '>=3.3.0 <3.7.1'
-    operators.operatorframework.io/builder: operator-sdk-v1.2.0
+    olm.skipRange: ">=3.5.0 <3.7.1"
```
horis233, line 305: Any reason we need to change it to >=3.5.0?

horis233: Is it because https://github.com/IBM/ibm-common-service-operator/blob/7e263de5c0bfbd5eeb233373ca8a523afcf6c6c6/Makefile#L208

bundle/manifests/ibm-common-service-operator.clusterserviceversion.yaml
bitscuit: weird, I thought I rebased from master already

bundle/manifests/ibm-common-service-operator.clusterserviceversion.yaml
bitscuit: no, that is a typo. Good catch

testing.md
```diff
@@ -0,0 +1,91 @@
...
+1. add features/fixes to repo
+2. change `image` value in config/manager/manager.yaml to `quay.io/<your_namespace>/common-service-operator:dev`
+   - and any of the operand image values if necessary
+3. build operator with changes
+```
+make build-dev
+```
+4. build bundle containing changes and bundle image
+```
+make bundle-manifests VERSION=99.99.99
```
pgodowski, line 45: please avoid version `99.99.99` as we'd like to use this version as a magic version for feature branches

pgodowski (dismissed):

testing.md
bitscuit: the 99.99.99 is just a way to signify that this bundle is newer than what was just deployed. The `VERSION` could be anything as long as it follows semantic versioning and is higher than the previous version. During development, it is also expected that developer will use personal namespace for testing, so nothing should be pushed to `opencloudio`

bitscuit:
@horis233 @Daniel-Fan this is ready for review again. I added a github workflow to automate testing fresh installation. This is meant to layout the testing infrastructure, so more tests can be added later.

The `bundle/manifest/csv.yaml` file got some formatting changes after I ran `make bundle-manifests`, but this should be fine. I tested installing with it already.

Makefile
```diff
@@ -59,20 +64,23 @@ else
 endif
 
 # Default image repo
-QUAY_REGISTRY ?= quay.io/opencloudio
+QUAY_REGISTRY ?= quay.io/bitscuit
```
horis233, line 67: Should we put images to your registry?

Makefile
```diff
@@ -95,15 +103,20 @@ include common/Makefile.common.mk
 
 ##@ Development
 
+OS    = $(shell uname -s | tr '[:upper:]' '[:lower:]')
+ARCH  = $(shell uname -m | sed 's/x86_64/amd64/')
```
horis233, line 107: We already have `LOCAL_OS` and `LOCAL_ARCH`, should we reuse them?

Makefile
```diff
@@ -137,29 +179,45 @@ uninstall: manifests ## Uninstall CRDs from a cluster
...
-	@docker push $(QUAY_REGISTRY)/$(OPERATOR_IMAGE_NAME):dev
+	@docker push $(REGISTRY)/$(OPERATOR_IMAGE_NAME)-$(LOCAL_ARCH):dev
 
 build-bundle-image:
 	@cp -f bundle/manifests/ibm-common-service-operator.clusterserviceversion.yaml /tmp/ibm-common-service-operator.clusterserviceversion.yaml
 	yq eval -i 'del(.spec.replaces)' bundle/manifests/ibm-common-service-operator.clusterserviceversion.yaml
-	docker build -f bundle.Dockerfile -t $(QUAY_REGISTRY)/$(BUNDLE_IMAGE_NAME):$(VERSION) .
-	docker push $(QUAY_REGISTRY)/$(BUNDLE_IMAGE_NAME):$(VERSION)
+	docker build -f bundle.Dockerfile -t $(QUAY_REGISTRY)/$(BUNDLE_IMAGE_NAME):$(RELEASE_VERSION) .
+	docker push $(QUAY_REGISTRY)/$(BUNDLE_IMAGE_NAME):$(RELEASE_VERSION)
```
horis233, line 196: Any reason why we use `release_version` here instead of the version of git status?

Makefile
horis233: Could we create another registry in opencloudio for testing?

.github/workflows/ocp.yaml
```diff
@@ -0,0 +1,91 @@
...
+          make build-dev-image
+
+      - name: Generate bundle
+        run: |
+          mkdir ./bin
+          export PATH=$PATH:$(pwd)/bin
+          echo "$(pwd)/bin" >> $GITHUB_PATH
+          make clis
+
+          yq -i eval ".spec.template.spec.containers[0].image = \"$QUAY_REGISTRY/common-service-operator:dev\"" config/manager/manager.yaml
```
horis233, line 67: The image built from `make build-dev-image` has `local-arch` as suffix

Makefile
horis233: I see you input the `RELEASE_VERSION` from workflow

testing.md
```diff
@@ -0,0 +1,91 @@
...
+6. run tests
+7. clean up
+```
+make cleanup-bundle
+```
+
+## Test upgrade with bundle
+Similar to fresh installation test except you will first deploy the operator/bundle of the most recent release without any changes, i.e. the operator/bundle code from the most recent commit in master branch
+
+1. change `image` value in config/manager/manager.yaml to `quay.io/<your_namespace>/common-service-operator:3.7.1`
```
horis233, line 61: It can't test the CSV change during the upgrade, right?

Makefile
bitscuit: Yes, I'll change it. This was leftover pieces of code I didn't clean up

Makefile
bitscuit: So the reason why we even need to use quay.io in the first place is because operator-sdk run bundle does not support private registries yet, otherwise I would have just used Artifactory. Yea, we can change this to use another registry in opencloudio. I was just using my own to test.

.github/workflows/ocp.yaml
bitscuit: good catch, will fix it

testing.md
bitscuit: `operator-sdk run bundle-upgrade` command will test CSV changes. The `run bundle` and `run bundle-upgrade` commands create a local CatalogSource and actually create a Subscription for the operator.

Makefile
bitscuit: yes, I decided to use `RELEASE_VERSION` since the default for that makes more sense when testing upgrade path

testing.md
horis233: Yes, but from my understanding, these steps only replace the images used in csv. Both 3.7.1 and dev image are using the same CSV template, which is the latest version. So we miss the change between the 3.7.1 version CSV and the dev version CSV template.

Since we are using the CSV to keep some configurations, this upgrade testing ignores these configuration updates.

bitscuit: Steps 4 - 7 would be where the developer can make changes to the CSV. And the bundle is re-created with the new CSV https://github.com/IBM/ibm-common-service-operator/pull/431/files#diff-c7f6993de081c6827242e2bb49530eb5cb078a7900d9d9391fdeef2ac9060970R71-R82

horis233: From my understanding, steps 1-3 should happen on branch release-3.6/release3.7 and steps 4-7 should happen on the master branch.

bitscuit: Oh, I see what you mean now. I think step 1 and 2 aren't actually needed. Because the bundle image for an existing release is already on quay.io, which should be referencing the correct operator image. So developer should just need to push the official bundle image to their registry and run 
```
make run-bundle
``` 
or optionally 
```
make run-bundle QUAY_REGISTRY=dev-registry BUNDLE_IMAGE_NAME=ibm-common-service-operator-bundle
```
Does that look right?

horis233: Thanks, It sounds good

Makefile
bitscuit: new commit resolves this

testing.md
bitscuit: new commits should resolve this

bitscuit:
@horis233 @Daniel-Fan any concerns or can this be merged?

Daniel-Fan (approved):
It looks good to me.

@horis233 If it is good to you as well. Please help to merge it. Thanks!

PR #433, opened 2021-03-10
fix: Keep both storageclass and size info

### LINKED ISSUE #432: keep both storageclass and size info
When the stoageclass is added into a commonservice CR before creating the operandconfig, operandconfig will only have storageclass info but no size info

horis233:
iterate services in the newconfig instead of services in the common service CR

fix #432


bitscuit (approved):
/lgtm

PR #436, opened 2021-03-12
uncomment RLEASE_VERSION to avoid overwriting

Daniel-Fan:


Makefile
```diff
@@ -223,7 +223,7 @@ build-push-image: $(CONFIG_DOCKER_TARGET) $(CONFIG_DOCKER_TARGET_QUAY) build-ope
 	@docker push $(QUAY_REGISTRY)/$(OPERATOR_IMAGE_NAME)-$(LOCAL_ARCH):$(VERSION)
 
 multiarch-image: $(CONFIG_DOCKER_TARGET) $(CONFIG_DOCKER_TARGET_QUAY) ## Generate multiarch images for operator image.
-	@MAX_PULLING_RETRY=20 RETRY_INTERVAL=30 common/scripts/multiarch_image.sh $(ARTIFACTORYA_REGISTRY) $(OPERATOR_IMAGE_NAME) $(VERSION) $(RELEASE_VERSION)
+	# @MAX_PULLING_RETRY=20 RETRY_INTERVAL=30 common/scripts/multiarch_image.sh $(ARTIFACTORYA_REGISTRY) $(OPERATOR_IMAGE_NAME) $(VERSION) $(RELEASE_VERSION)
 	@MAX_PULLING_RETRY=20 RETRY_INTERVAL=30 common/scripts/multiarch_image.sh $(QUAY_REGISTRY) $(OPERATOR_IMAGE_NAME) $(VERSION) $(LATEST_VERSION)
```
horis233, line 227: Should be this line, right?

horis233: @Daniel-Fan ^^^

Daniel-Fan: @horis233 
Will `make multiarch-image` be used by CICD to build the image?
If not, we should remove `RELEASE_VERSION` tag in case it will overwrite the build by CICD like tag `3.7.2`.

And I don't think the `latest` tag matters since we have both `EUS` and continuous delivery at the same time. `latest` tag is not reliable at all.

horis233: Oh, the command you commented out is for pushing images to the artifactory, which is required.

Daniel-Fan: Oh i see, maybe I could just comment the command for multi-arch image on `quay.io` for `latest` tag

Makefile
```diff
@@ -223,8 +223,8 @@ build-push-image: $(CONFIG_DOCKER_TARGET) $(CONFIG_DOCKER_TARGET_QUAY) build-ope
 	@docker push $(QUAY_REGISTRY)/$(OPERATOR_IMAGE_NAME)-$(LOCAL_ARCH):$(VERSION)
```
bitscuit: Should comment out this line and [line 221](https://github.com/IBM/ibm-common-service-operator/pull/436/files/c8cdb8bdb8ddc67e3538559120332aa4a237f226..6f2ad30fc3a9d1abeeca96df021ca9c51314d8d5#diff-76ed074a9305c04054cdebb9e9aad2d818052b07091de1f20cad0bbac34ffb52L221) as well. No need for PROW to push to quay at all

Daniel-Fan: Actually I don't think this is the problem. 
For example, it pushes to repo `ibm-common-service-operator-amd64`, not the repo `ibm-common-service-operator` with `multiarch`.

and anyhow we should avoid to push image to `quay.io`

Daniel-Fan:
The PR is ready to merge

horis233 (approved):
/lgtm

PR #446, opened 2021-03-25
SaaS deployment at resource initialization stage

Daniel-Fan:
1. check `saas-config` and `common-service-maps` to check the SaaS deployment and obtain the corresponding configurations.
2. generate control namespace if it is SaaS deployment.
3. Deploy `secretshare` and `webhook` into control namespace

### DIFF
--- a/controllers/bootstrap/init.go
+++ b/controllers/bootstrap/init.go
@@ -112,13 +112,14 @@ func (b *Bootstrap) InitResources(manualManagement bool) error {
 	}
 
 	// Check Saas Deployment
+	controlNs := b.MasterNamespace
 	saasEnable, err := checkSaas(b.Reader)
 	if err != nil {
 		return err
 	}
 	if saasEnable {
 		klog.Info("Saas Deployment Enabled for Common Services")
-		controlNs := util.GetControlNs(b.Reader)
+		controlNs = util.GetControlNs(b.Reader)
 		if err := bs.CreateNamespace(controlNs); err != nil {
 			klog.Errorf("Failed to create control namespace: %v", err)
 			return err
@@ -180,11 +181,11 @@ func (b *Bootstrap) InitResources(manualManagement bool) error {
 			return err
 		}
 		// Create Operator RBAC
-		if err := b.createOrUpdateFromYaml([]byte(util.Namespacelize(operator.RBAC, b.MasterNamespace))); err != nil {
+		if err := b.createOrUpdateFromYaml([]byte(util.Namespacelize(operator.RBAC, controlNs))); err != nil {
 			return err
 		}
 		// Create Operator Deployment
-		if err := b.createOrUpdateFromYaml([]byte(util.ReplaceImages(util.Namespacelize(operator.Deployment, b.MasterNamespace)))); err != nil {
+		if err := b.createOrUpdateFromYaml([]byte(util.ReplaceImages(util.Namespacelize(operator.Deployment, controlNs)))); err != nil {
 			return err
 		}
 		// Wait for CRD ready
@@ -192,7 +193,7 @@ func (b *Bootstrap) InitResources(manualManagement bool) error {
 			return err
 		}
 		// Create Operator CR
-		if err := b.createOrUpdateFromYaml([]byte(util.Namespacelize(operator.CR, b.MasterNamespace))); err != nil {
+		if err := b.createOrUpdateFromYaml([]byte(util.Namespacelize(operator.CR, controlNs))); err != nil {
 			return err
 		}
 	}
@@ -222,7 +223,7 @@ func (b *Bootstrap) InitResources(manualManagement bool) error {
 		}
 	}
 
-	// create or ODLM  OperandRegistry and OperandConfig CR resources
+	// create and wait ODLM OperandRegistry and OperandConfig CR resources
 	if err := b.waitResourceReady("operator.ibm.com/v1alpha1", "OperandRegistry"); err != nil {
 		return err
 	}
--- a/controllers/bootstrap/init.go
+++ b/controllers/bootstrap/init.go
@@ -111,19 +111,15 @@ func (b *Bootstrap) InitResources(manualManagement bool) error {
 		return err
 	}
 
-	// Check Saas Deployment
-	controlNs := b.MasterNamespace
-	saasEnable, err := checkSaas(b.Reader)
-	if err != nil {
-		return err
-	}
-	if saasEnable {
-		klog.Info("Saas Deployment Enabled for Common Services")
-		controlNs = util.GetControlNs(b.Reader)
-		if err := bs.CreateNamespace(controlNs); err != nil {
+	// Check Saas or Multi intances Deployment
+	controlNs := util.GetControlNs(b.Reader)
+	if len(controlNs) > 0 {
+		if err := b.CreateNamespace(controlNs); err != nil {
 			klog.Errorf("Failed to create control namespace: %v", err)
 			return err
 		}
+	} else {
+		controlNs = b.MasterNamespace
 	}
 
 	operatorNs, err := util.GetOperatorNamespace()
--- a/controllers/bootstrap/init.go
+++ b/controllers/bootstrap/init.go
@@ -111,6 +111,20 @@ func (b *Bootstrap) InitResources(manualManagement bool) error {
 		return err
 	}
 
+	// Check Saas Deployment
+	saasEnable, err := checkSaas(b.Reader)
+	if err != nil {
+		return err
+	}
+	if saasEnable {
+		klog.Info("Saas Deployment Enabled for Common Services")
+		controlNs := util.GetControlNs(b.Reader)
+		if err := bs.CreateNamespace(controlNs); err != nil {
+			klog.Errorf("Failed to create control namespace: %v", err)
+			return err
+		}
+	}
+
 	operatorNs, err := util.GetOperatorNamespace()
 	if err != nil {
 		klog.Errorf("Getting operator namespace failed: %v", err)
@@ -237,14 +251,14 @@ func (b *Bootstrap) InitResources(manualManagement bool) error {
 	return nil
 }
 
-func (b *Bootstrap) CreateNamespace() error {
+func (b *Bootstrap) CreateNamespace(name string) error {
 	nsObj := &corev1.Namespace{
 		TypeMeta: metav1.TypeMeta{
 			Kind:       "Namespace",
 			APIVersion: "v1",
 		},
 		ObjectMeta: metav1.ObjectMeta{
-			Name: b.MasterNamespace,
+			Name: name,
 		},
 	}
 
--- a/controllers/common/util.go
+++ b/controllers/common/util.go
@@ -43,12 +43,13 @@ import (
 )
 
 type CsMaps struct {
+	ControlNs     string      `json:"controlNamespace"`
 	NsMappingList []nsMapping `json:"namespaceMapping"`
-	DefaultCsNs   string      `json:"defaultCsNs"`
+	// DefaultCsNs   string      `json:"defaultCsNs"`
 }
 
 type nsMapping struct {
-	RequestNS []string `json:"requested-from-namespace"`
+	RequestNs []string `json:"requested-from-namespace"`
 	CsNs      string   `json:"map-to-common-service-namespace"`
 }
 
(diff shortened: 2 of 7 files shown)


controllers/bootstrap/init.go
```diff
@@ -111,6 +111,21 @@ func (b *Bootstrap) InitResources(manualManagement bool) error {
...
 
+	// Check Saas Deployment
+	controlNs := b.MasterNamespace
+	saasEnable, err := checkSaas(b.Reader)
+	if err != nil {
+		return err
+	}
+	if saasEnable {
+		klog.Info("Saas Deployment Enabled for Common Services")
+		controlNs = util.GetControlNs(b.Reader)
```
horis233, line 122: We need a check to make sure if controlNs is empty

Daniel-Fan: Did you mean that even if the SaaS deployment is enabled, the `control namespace` in `common-service-maps` ConfigMap may be empty, then we should have a default value like what we did for `MasterNamespace` which is `ibm-common-services`?

horis233: Yes, exactly.

controllers/common/util.go
```diff
@@ -308,3 +309,54 @@ func findNamespace(nsList []string, nsName string) (exist bool) {
...
+	if v != "true" {
+		return false, nil
+	}
+	return true, nil
+}
+
+// GetControlNs gets control namespace of deploying cluster scope services
+func GetControlNs(r client.Reader) (controlNs string) {
+	// default control namespace
+	controlNs = constant.ControlNamespace
```
Daniel-Fan, line 337: I set the default value for `controlNs` to `ibm-cs-control` temporarily.
We could have a discussion about choosing a proper name on the sync up meeting.

Also I firstly set `controlNs` to `MasterNamespce` here https://github.com/IBM/ibm-common-service-operator/blob/36a20289b90b930091ac98b30a719be61a41b112/controllers/bootstrap/init.go#L115
then if it is not a SaaS deployment, all the services will be deployed into `MasterNamespace`.

controllers/bootstrap/init.go
Daniel-Fan: ah i see, Updated. Please see the comment here
https://github.com/IBM/ibm-common-service-operator/pull/446#discussion_r602811791

bitscuit (changes requested):

controllers/common/util.go
```diff
@@ -308,3 +309,54 @@ func findNamespace(nsList []string, nsName string) (exist bool) {
...
+
+	csConfigmap, err := GetCmOfMapCs(r)
+	if err != nil {
+		klog.V(2).Infof("Don't find configmap kube-public/common-service-maps: %v", err)
```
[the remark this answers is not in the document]
Daniel-Fan: Great, Updated!

Daniel-Fan:
@horis233 @bitscuit This PR is ready for reviewing.

For deploying webhook and secretshare in control namespace, we add below to cs-operator's cluster role.
1. several permission for `services`
2. `get`, `create` permission for `serviceaccounts` and `deployments`

controllers/bootstrap/init.go
bitscuit: Can users not install normally (shared CS) on SaaS? So SaaS installation automatically means multiple CS instances? If they can still install normally (shared CS), then control namespace should be the same as master namespace.

controllers/bootstrap/init.go
```diff
@@ -111,6 +111,21 @@ func (b *Bootstrap) InitResources(manualManagement bool) error {
 		return err
 	}
 
+	// Check Saas Deployment
+	controlNs := b.MasterNamespace
+	saasEnable, err := util.CheckSaas(b.Reader)
+	if err != nil {
+		return err
+	}
+	if saasEnable {
```
bitscuit, line 120: we should have additional check here for `common-service-maps` to support installing multiple CS instances in non-SaaS environment

```suggestion
	}
	
	// check multiple CS instances installation
	isMultiCs, err := util.IsMultiCs(b.Reader)
	if err != nil {
	    return err
	}
	
	if saasEnable || isMultiCs{
```

bitscuit (changes requested):

controllers/bootstrap/init.go
Daniel-Fan: Based on slack message, the difference of SaaS and MultiCS is `OperandRegistry` and `OperandConfig`, does it indicate that we will deploy cluster scope resources in cs control namespace for both cases?

bitscuit: If we do not deploy in control namespace, how would multiple CS work?

Daniel-Fan: But there is a problem, I think that it will introduce a shared control namespace even for single instance installation.
How do we know it is a multiple CS deployment in non-SaaS?

From my understanding, so far in the non-SaaS deployment, the `common-services-maps` is created by user manually.

Firstly, they create the ConfigMap with only one instance of CS like this
```
apiVersion: v1
kind: ConfigMap
metadata:
  name: common-service-maps
  namespace: kube-public
data: 
  common-service-maps.yaml: |
    namespaceMapping:
    - requested-from-namespace:
      - cp4i
      - cp4mcm
      map-to-common-service-namespace: cp-common-service
```

After the installation of CS, they add another mapping into the ConfigMap and start another installation.
apiVersion: v1
kind: ConfigMap
metadata:
  name: common-service-maps
  namespace: kube-public
data: 
  common-service-maps.yaml: |
    namespaceMapping:
    - requested-from-namespace:
      - cp4i
      - cp4mcm
      map-to-common-service-namespace: cp-common-service
     - requested-from-namespace:
      - iaf
      - watson
      map-to-common-service-namespace: watson-common-service
```

So at the first stage, the common service knows that there is only one instance, and deploy everything(current case) or into one namespace called `cp-common-service`. Or deploy cluster-scope resources into shared namespace(new design).

After the user add another mapping into ConfigMap and start the installation of second instance of CS, the CS will recognize that it is a multiple instance case.
But at this time some cluster scope resources have been deployed in the first namespace.

If we have a shared namespace, everything goes fine.

If we don't have a shared namespace for cluster scope resources in Non-SaaS multi-instance deployment.
We need to use the first CS-deployed namespace as the master namespace for the cluster scope resources.

/cc @horis233

**Comment by Daniel-Fan** [comment]
After the discussion with @horis233 

Firstly, user need to decide whether the multi instances of CS will be enabled for this cluster before installing the first CS, indicating that we *did not* support the switch between singleton CS and multiple CS.

If it is the singleton CS, the installation flow will be the same as now. we would check `len(requested-from-namespace)` is 1 to ensure the compatibility between arbitrary namespace installation and multi instances.

If it is the multiple CS installation, user need to set `multiInstanceEnabled: true` in `common-service-maps`
```
apiVersion: v1
kind: ConfigMap
metadata:
  name: common-service-maps
  namespace: kube-public
data: 
  common-service-maps.yaml: |
    multiInstanceEnabled: true
    controlNamespace: ibm-cs-control
    namespaceMapping:
    - requested-from-namespace:
      - cp4i
      - cp4mcm
      map-to-common-service-namespace: cp-common-service
     - requested-from-namespace:
      - iaf
      - watson
      map-to-common-service-namespace: watson-common-service
```

The control namespace `ibm-cs-control` is for share cluster scope resource.

Additional check:

1. We are not allowed to for one `CloudPakANS` in `requested-from-namespace`  to map twice.

2. We only allow to add more mapping into into ConfigMap, did not support replace or change the existing mapping.

3. If user wanna to remove one of namespace mapping. For example, removing `cp4mcm` mapping, firstly user needs to remove the services `cp4mcm` and `cs-operator` in `cp4mcm` namespace.

@bitscuit Please check the above note, welcome for the objection ^_^

File: controllers/bootstrap/init.go (hunk 1 above)
Comment (horis233) [reply controllers/bootstrap/init.go:122]: @bitscuit 
As we discussed today, end-user or SaaS automatically can use `multiInstanceEnabled` to decide install common service into shared or multi-instance mode.

Also, we can't set the control namespace as one of the master namespaces. Since each tenant scope needs to have permission to control namespace. If the control namespace is within one scope, that will cause there is a permission overlap between tenant scope.  We can set a check for it.

Comment (Daniel-Fan) [reply controllers/bootstrap/init.go:122]: for the SaaS deployment, we should have a separate control namespace which distinguishes from any master namespace for the cluster scope resource.

It is because the one item in `namespaceMapping` list has its own scope like `cp4i`, `cp4mcm` and `cp-common-service`, we should not include control namespace in any of them.

```
controlNamespace: ibm-cs-control
    namespaceMapping:
     - requested-from-namespace:
       - cp4i
       - cp4mcm
       map-to-common-service-namespace: cp-common-service
     - requested-from-namespace:
       - iaf
       - watson
       map-to-common-service-namespace: watson-common-service
```

File: controllers/bootstrap/init.go (hunk 4 above)
Comment (Daniel-Fan) [reply controllers/bootstrap/init.go]: update to check the SaaS and multi instances deployment by `control namespace`

**Comment by Daniel-Fan** [comment]
Temporarily disable the `lint-go` check for avoiding cyclomatic complexity check, will optimize it later.

Also will have a separate issue to track the enhancement for multi CS instances on non-SaaS deployment

@horis233 @bitscuit This PR is ready for reviewing.

File: common/Makefile.common.mk [hunk 5]
```diff
@@ -44,10 +44,11 @@ lint-copyright-banner:
 	@${FINDFILES} \( -name '*.go' -o -name '*.cc' -o -name '*.h' -o -name '*.proto' -o -name '*.py' -o -name '*.sh' \) \( ! \( -name '*.gen.go' -o -name '*.pb.go' -o -name '*_pb2.py' \) \) -print0 |\
 		${XARGS} common/scripts/lint_copyright_banner.sh
 
-lint-go:
-	@${FINDFILES} -name '*.go' \( ! \( -name '*.gen.go' -o -name '*.pb.go' \) \) -print0 | ${XARGS} common/scripts/lint_go.sh
+# lint-go:
```
horis233, line 47: Let's disable individual check from here: https://github.com/IBM/ibm-common-service-operator/blob/master/common/config/.golangci.yml#L32 instead of disabling all the check

Daniel-Fan: Sure, updated

controllers/bootstrap/init.go
```diff
@@ -166,19 +177,19 @@ func (b *Bootstrap) InitResources(manualManagement bool) error {
...
+		if err := b.createOrUpdateFromYaml([]byte(util.ReplaceImages(util.Namespacelize(operator.Deployment, controlNs)))); err != nil {
 			return err
 		}
 		// Wait for CRD ready
 		if err := b.waitResourceReady(operator.APIVersion, operator.Kind); err != nil {
 			return err
 		}
 		// Create Operator CR
-		if err := b.createOrUpdateFromYaml([]byte(util.Namespacelize(operator.CR, b.MasterNamespace))); err != nil {
+		if err := b.createOrUpdateFromYaml([]byte(util.Namespacelize(operator.CR, controlNs))); err != nil {
```
horis233, line 192: Just a reminder, we don't need to deploy secretshare CR in the SaaS situation.
Thus I believe we can create separate functions for SaaS and non-SaaS instead of sharing the same flow. It keeps flexible when we need to remove the secretshare operator.
 
We can update it in the future PR.

controllers/bootstrap/init.go
```diff
@@ -111,6 +111,17 @@ func (b *Bootstrap) InitResources(manualManagement bool) error {
...
+	controlNs := util.GetControlNs(b.Reader)
+	if len(controlNs) > 0 {
+		if err := b.CreateNamespace(controlNs); err != nil {
+			klog.Errorf("Failed to create control namespace: %v", err)
+			return err
+		}
+	} else {
+		controlNs = b.MasterNamespace
+	}
+
```
horis233, line 124: Just a thought, could we add `ControlNamespace` and `SaaSenabled` into the `Bootstrap` structure, like what we did for MasterNamespace. Then we don't need to trigger the function everytime

horis233 (approved):
I will approve this PR, please consider my above comments

Daniel-Fan:
No problem, I will integrated those two in `bootstrap` structure in this PR https://github.com/IBM/ibm-common-service-operator/pull/450

Also will separate the install flow for SaaS and non-SaaS and remove secretshare CR after the separation in future PR

PR #448, opened 2021-03-26
Workflow

bitscuit:
Current workflow only runs properly if opening a pull request from a branch in the upstream repo. The workflow will not get Github secrets from forked PRs using the `pull_request` event, so need to use `pull_request_target`. 

There are some security concerns with this though, as described [here](https://securitylab.github.com/research/github-actions-preventing-pwn-requests/). The main concern is that the workflow will run code from any forked PR, which could contain malicious code. To prevent this, a check is added so that the code will only be checked out if repo maintainers label the PR as `ok-to-test`.

### DIFF
--- a/.github/workflows/ocp.yaml
+++ b/.github/workflows/ocp.yaml
@@ -18,7 +18,7 @@ name: ocp.yaml
 
 on:
   pull_request_target:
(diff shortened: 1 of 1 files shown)


Daniel-Fan (approved):
/lgtm

PR #455, opened 2021-03-31
update crd version

horis233:


Daniel-Fan:
@bitscuit 

Any idea that rules for make can't be found?
```
make: *** No rule to make target 'build-dev-image'.  Stop.
Error: Process completed with exit code 2.
```
https://github.com/IBM/ibm-common-service-operator/pull/455/checks?check_run_id=2234287678

bitscuit:
> 
> 
> @bitscuit
> 
> Any idea that rules for make can't be found?
> 
> ```
> make: *** No rule to make target 'build-dev-image'.  Stop.
> Error: Process completed with exit code 2.
> ```
> 
> https://github.com/IBM/ibm-common-service-operator/pull/455/checks?check_run_id=2234287678

Because the job will not checkout any code unless we tag a PR with `ok-to-test`. Since the repo is not checked out, the job, the Makefile does not exist

Daniel-Fan:
Thanks @bitscuit for the clarification!

It is working after I added the `ok-to-test` label.

PR #463, opened 2021-04-02
watching cs-maps change, validate cs-mapping and filter its cs-scope

Daniel-Fan:
1. When the `cs-mapping` ConfigMap is changed, it will reconcile the master CR, going through the steps of initializing the resource to update the scope list in NSS CR for both `common-service` and `odlm-scope` to update the scope for individual cs operator and ODLM.

2. Also `cs-operator` will validate the `cs-mapping` ConfigMap, if it is an unsupported case, the `cs-operator` will raise the error. In the future, we will handle it by validating web hook, to prohibit the invalid change to ConfigMap from the user.

3. Also `cs-operator` will filter its `cs-scope` from `cs-mapping` ConfigMap. If the triggered CR is not in the namespace of this scope. It will skip the reconciliation.

### DIFF
--- a/controllers/common/util.go
+++ b/controllers/common/util.go
@@ -380,26 +380,27 @@ func ValidateCsMaps(cm *corev1.ConfigMap) error {
 		return fmt.Errorf("failed to fetch data of configmap common-service-maps: %v", err)
 	}
 
-	nsSet := make(map[string]interface{})
+	CsNsSet := make(map[string]interface{})
+	RequestNsSet := make(map[string]interface{})
 
 	for _, nsMapping := range cmData.NsMappingList {
 		// validate masterNamespace and controlNamespace
 		if cmData.ControlNs == nsMapping.CsNs {
 			return fmt.Errorf("invalid controlNamespace: %v", cmData.ControlNs)
 		}
-		if _, ok := nsSet[nsMapping.CsNs]; ok {
+		if _, ok := CsNsSet[nsMapping.CsNs]; ok {
 			return fmt.Errorf("invalid map-to-common-service-namespace: %v", nsMapping.CsNs)
 		}
-		nsSet[nsMapping.CsNs] = struct{}{}
+		CsNsSet[nsMapping.CsNs] = struct{}{}
 		// validate CloudPak Namespace and controlNamespace
 		for _, ns := range nsMapping.RequestNs {
 			if cmData.ControlNs == ns {
 				return fmt.Errorf("invalid controlNamespace: %v", cmData.ControlNs)
 			}
-			if _, ok := nsSet[ns]; ok {
+			if _, ok := RequestNsSet[ns]; ok {
 				return fmt.Errorf("invalid requested-from-namespace: %v", ns)
 			}
-			nsSet[ns] = struct{}{}
+			RequestNsSet[ns] = struct{}{}
 		}
 	}
 	return nil
--- a/controllers/common/util.go
+++ b/controllers/common/util.go
@@ -258,9 +258,9 @@ func GetMasterNs(r client.Reader) (masterNs string) {
 }
 
 // UpdateNSList updates adopter namespaces of Common Services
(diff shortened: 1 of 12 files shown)


controllers/common/util.go
```diff
@@ -367,3 +367,71 @@ func GetControlNs(r client.Reader) (controlNs string) {
...
+		nsSet[nsMapping.CsNs] = struct{}{}
+		// validate CloudPak Namespace and controlNamespace
+		for _, ns := range nsMapping.RequestNs {
+			if cmData.ControlNs == ns {
+				return fmt.Errorf("invalid controlNamespace: %v", cmData.ControlNs)
+			}
+			if _, ok := nsSet[ns]; ok {
+				return fmt.Errorf("invalid requested-from-namespace: %v", ns)
+			}
+			nsSet[ns] = struct{}{}
```
horis233, line 402: @Daniel-Fan 

My concern is how about users what to install common services in the same namespace as the cloudpaks. This function looks can't fulfill that.

Could we use a set for `CsNs` and another set for `RequestNs`？

Daniel-Fan: Great point, I did not consider this case at the first thought. Updated.

Daniel-Fan:
@horis233 @bitscuit 
This PR is ready, please review it when you are available.

controllers/common/util.go
```diff
@@ -367,3 +367,96 @@ func GetControlNs(r client.Reader) (controlNs string) {
...
+		return fmt.Errorf("failed to fetch data of configmap common-service-maps: %v", err)
+	}
+
+	CsNsSet := make(map[string]interface{})
+	RequestNsSet := make(map[string]interface{})
+
+	for _, nsMapping := range cmData.NsMappingList {
+		// validate masterNamespace and controlNamespace
+		if cmData.ControlNs == nsMapping.CsNs {
+			return fmt.Errorf("invalid controlNamespace: %v", cmData.ControlNs)
```
bitscuit, line 389:
```suggestion
			return fmt.Errorf("Invalid controlNamespace: %v", cmData.ControlNs. Cannot be the same as map-to-common-service-namespace.)
```
Clearer error log

controllers/common/util.go
```diff
@@ -367,3 +367,96 @@ func GetControlNs(r client.Reader) (controlNs string) {
...
+		}
+		if _, ok := CsNsSet[nsMapping.CsNs]; ok {
+			return fmt.Errorf("invalid map-to-common-service-namespace: %v", nsMapping.CsNs)
+		}
+		CsNsSet[nsMapping.CsNs] = struct{}{}
+		// validate CloudPak Namespace and controlNamespace
+		for _, ns := range nsMapping.RequestNs {
+			if cmData.ControlNs == ns {
+				return fmt.Errorf("invalid controlNamespace: %v", cmData.ControlNs)
```
bitscuit, line 398:
```suggestion
				return fmt.Errorf("Invalid controlNamespace: %v", cmData.ControlNs. Cannot be the same as requested-from-namespace.)
```

bitscuit (changes requested):

main.go
```diff
@@ -67,18 +72,46 @@ func main() {
...
+				constant.CsManagedLabel: "true",
+			})
+			if err := mgr.GetClient().Update(context.TODO(), cm); err != nil {
+				klog.Errorf("Failed to update labels for common-service-maps: %v", err)
+				os.Exit(1)
+			}
+		}
+	} else if !errors.IsNotFound(err) {
+		klog.Errorf("Failed to get common-service-maps: %v", err)
+		os.Exit(1)
```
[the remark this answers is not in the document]
Daniel-Fan: I think it is fine, this if statement is `true` only when we have an unexpected error to get the ConfigMap, not including the case that ConfigMap not existed.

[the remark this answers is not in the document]
horis233: Oh, my bad. I didn't notice that

horis233 (approved):
/lgtm

PR #469, opened 2021-04-08
Disable licensing route on SaaS

pgodowski:
When on SaaS, do not create the external OCP route for License Service.


horis233 (approved):
/lgtm

PR #476, opened 2021-04-09
add IAM SaaS config 

horis233:


bitscuit:
So IAM will check if SaaS environment on its own now?

horis233:
We can add IAM SaaS configuration into the OperandConfig template, so we can remove the logic for updating OperandConfig. I will update this PR after #463 is merged.

Daniel-Fan (approved):
/lgtm

PR #477, opened 2021-04-13
Retry for failed cert resource deploying

Daniel-Fan:


Daniel-Fan:
Hi @bitscuit 
From the error in github workflow, should we run `make cleanup-bundle` to remove catalog source as well?
https://github.com/IBM/ibm-common-service-operator/pull/477/checks?check_run_id=2335097281

horis233:
@Daniel-Fan Could we change retry to 10s? It can decrease the time of the installation and wouldn't cause much API pressure.

Daniel-Fan:
@horis233 
Sure, BTW is that ok if I changed wait time for CRD from 10s to 30s?

horis233:
@Daniel-Fan Could we change both of them to 10s?

Daniel-Fan:
@horis233 
No problem, updated!

horis233 (approved):
/lgtm

bitscuit:
> 
> 
> Hi @bitscuit
> From the error in github workflow, should we run `make cleanup-bundle` to remove catalog source as well?
> https://github.com/IBM/ibm-common-service-operator/pull/477/checks?check_run_id=2335097281

The workflow does have a step where cleanup is done https://github.com/IBM/ibm-common-service-operator/blob/master/.github/workflows/ocp.yaml#L98. But I think because the workflow runs multiple times in each PR due to being triggered whenever labels change, it's not cleaning up properly. I need to add additional checks for existing catalogsource and remove them for now

PR #480, opened 2021-04-16
Using template to render yaml files

horis233:
Since we are going to set approval mode and catalogsource dynamically, It's hard for us to use string replace for everything.
I believe using this kind of "helm like"  template can help us set some common parameters, like `namespace` and `version`.

Welcome to any other option that can help make the template becoming more flexible.
 
cc @Daniel-Fan


controllers/constant/namespace.go
```diff
@@ -16,35 +16,67 @@
...
+  name: ibm-namespace-scope-operator-restricted
+  namespace: {{ .MasterNs }}
+  annotations:
+    version: {{ .Version }}
+spec:
+  channel: {{ .Channel }}
+  installPlanApproval: Automatic
+  name: ibm-namespace-scope-operator-restricted
+  source: opencloud-operators
+  sourceNamespace: openshift-marketplace
```
Daniel-Fan, line 32: Could we change Nss's `source` and `sourceNamespace` to `{{ .CatalogSourceName }}` and `{{ .CatalogSourceNs }}` that is matching with CsSubscription?

controllers/constant/namespace.go
```diff
@@ -16,35 +16,67 @@
...
+  name: ibm-namespace-scope-operator
+  namespace: {{ .MasterNs }}
+  annotations:
+    version: {{ .Version }}
+spec:
+  channel: {{ .Channel }}
+  installPlanApproval: Automatic
+  name: ibm-namespace-scope-operator
+  source: opencloud-operators
+  sourceNamespace: openshift-marketplace
```
Daniel-Fan, line 48: Same as here

horis233: Nice catch! updated

ZhuoxiLi:
a question about the CI tests: in `test-ibm-common-service-operator-amd64`, there is `+ '[' x86_64 == ppc64le ']'`, what the purpose to compare with `ppc64le`?

horis233:
> a question about the CI tests: in `test-ibm-common-service-operator-amd64`, there is `+ '[' x86_64 == ppc64le ']'`, what the purpose to compare with `ppc64le`?

It comes from here: https://github.com/IBM/test-infra/blob/bd74eecf2edf4a7c7cad489aed97f6a58c2ca495/prow/docker/build-tool/docker-entrypoint.sh#L26

controllers/bootstrap/init.go
```diff
@@ -485,49 +490,62 @@ func (b *Bootstrap) installODLM(operatorNs string, annotations map[string]string
...
+			Channel:           b.CSData.Channel,
+			Version:           b.CSData.Version,
+			MasterNs:          b.CSData.MasterNs,
+			ODLMScopeEnable:   odlmScopeEnable,
+			CatalogSourceName: b.CSData.CatalogSourceName,
+			CatalogSourceNs:   b.CSData.CatalogSourceNs,
+		}
+
+		if err := b.renderTemplate(constant.ODLMNamespacedSubscription, odlmData); err != nil {
+			return err
```
Daniel-Fan, line 530: I am wondering that whether we could use CSDATA instead of a new `odlmData`.

if we could integrate `ODLMScopeEnable` into `CSDATA` with predefined value `false`, and update the value at line 500 here
https://github.com/IBM/ibm-common-service-operator/pull/480/files#diff-d871a1a5fb2d0fae940235f9a8aff9610fa4009086d909ddcc7392047894f32bR500

horis233: updated

Daniel-Fan: No, we don't need a default value.

In the ODLM, if env variable `ODLM_SCOPE` is not found, it will return `false` directly.

Daniel-Fan (approved):
/lgtm

PR #485, opened 2021-04-19
set manual approval for CS operators

Daniel-Fan:
Ref https://github.com/IBM/operand-deployment-lifecycle-manager/issues/627

### DIFF
--- a/controllers/commonservice_controller.go
+++ b/controllers/commonservice_controller.go
@@ -119,7 +119,7 @@ func (r *CommonServiceReconciler) ReconcileMasterCR(instance *apiv3.CommonServic
 	// Webhook Operator and Secretshare
 	// Delete ODLM from openshift-operators and deploy it in the masterNamespaces
 	// Deploy OperandConfig and OperandRegistry
-	if err := r.Bootstrap.InitResources(instance.Spec.ManualManagement); err != nil {
+	if err := r.Bootstrap.InitResources(instance.Spec.ManualManagement, instance.Spec.InstallPlanApproval); err != nil {
 		klog.Errorf("Failed to initialize resources: %v", err)
 		if err := r.updatePhase(instance, CRFailed); err != nil {
 			klog.Error(err)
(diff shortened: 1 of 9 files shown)


horis233:
@Daniel-Fan 

Could we set the manual approval mode when initializing the resources?
https://github.com/IBM/ibm-common-service-operator/blob/master/controllers/bootstrap/init.go#L140

We can a parameter `approvalMode` in the CSData. https://github.com/IBM/ibm-common-service-operator/blob/master/controllers/bootstrap/init.go#L80

Daniel-Fan:
@horis233 

I have some concerns about setting the manual approval mode in `InitResource`

We need to support the case that user will switch between `Manual` and `Automatic` after the installation of Common Service, indicating we need to update the subscription in each reconciliation.

However, In the `InitResource`, the logic of updating `OperandRegistry` and `Subscription` is based on predefined templates in `constant` package, and comparing the `annotation.version` between the existing one and template.

In this case,
1. We need to force the update for NSS `subscription` and `OperandRegistry` in each reconciliation, bypassing the version comparison in `CreateorUpdateFromYaml` function which is current update strategy.
2. It is risky that we overwrite the entire NSS `Subscription` and `OperandRegistry` by replacing it with the templates. Therefore, I choose to find the existing `OperandRegistry` and `Subscription`, and only change the field `installPlanApproval`, then apply the changed resource for updating.

horis233:
@Daniel-Fan 

I understand that we need to bypassing the version comparison when updating the operandregistry.

gyliu513:
@taylormgeorge91

controllers/commonservice_controller.go
```diff
@@ -119,7 +119,7 @@ func (r *CommonServiceReconciler) ReconcileMasterCR(instance *apiv3.CommonServic
 	// Webhook Operator and Secretshare
 	// Delete ODLM from openshift-operators and deploy it in the masterNamespaces
 	// Deploy OperandConfig and OperandRegistry
-	if err := r.Bootstrap.InitResources(instance.Spec.ManualManagement); err != nil {
+	if err := r.Bootstrap.InitResources(instance.Spec.ManualManagement, instance.Spec.InstallPlanApproval); err != nil {
```
horis233, line 122: @Daniel-Fan Could we just input the commonservice instance?

Daniel-Fan: Sure, Updated

horis233 (approved):
/lgtm

PR #490, opened 2021-04-21
Create a new NSS CR managed by ODLM

Daniel-Fan:
`odlm-scope-managedby-odlm` and `nss-managedby-odlm` will have the same `namespaceMember` value.
But the first one is used for extend permission on ODLM, the latter is for other individual operator's permission scope.

`odlm-scope-managedby-odlm` and `nss-managedby-odlm` should only be created when CS is deployed in Single instance case.

On SaaS or on-prem Multi Instances case, we will not have above two CRs and disable the NSS controller in ODLM to avoid the permission overlap in `control namespace`

### DIFF
--- a/controllers/bootstrap/init.go
+++ b/controllers/bootstrap/init.go
@@ -464,8 +464,15 @@ func (b *Bootstrap) installNssOperator(manualManagement bool) error {
 		return err
 	}
 
+	// Create General NSS CRs
 	if err := b.renderTemplate(constant.NamespaceScopeCR, b.CSData); err != nil {
 		return err
+	}
+	// Create NSS CRs managedby ODLM for Single CS instance case
+	if b.CSData.MasterNs == b.CSData.ControlNs {
+		if err := b.renderTemplate(constant.NamespaceScopeCRManagedbyODLM, b.CSData); err != nil {
+			return err
+		}
 	}
 
 	cm, err := util.GetCmOfMapCs(b.Reader)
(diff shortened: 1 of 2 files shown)


controllers/bootstrap/init.go
```diff
@@ -464,9 +464,16 @@ func (b *Bootstrap) installNssOperator(manualManagement bool) error {
 		return err
 	}
 
+	// Create General NSS CRs
 	if err := b.renderTemplate(constant.NamespaceScopeCR, b.CSData); err != nil {
 		return err
 	}
+	// Create NSS CRs managedby ODLM for Single CS instance case
+	if b.CSData.MasterNs == b.CSData.ControlNs {
```
horis233, line 472: Is my understanding correct? I believe ControlNs is empty ("") in the on-prem with single CS.

Daniel-Fan: When we first create the `bootstrap`, the `ControalNs` is empty.

But at the `initialResource`, we will assign `MasterNs` to `ControlNs` if `ControlNs` is empty, for the convenience to deploy `webhook` and `secretshare` in both Single and Multi cases.
https://github.com/IBM/ibm-common-service-operator/blob/master/controllers/bootstrap/init.go#L165
https://github.com/IBM/ibm-common-service-operator/blob/master/controllers/bootstrap/init.go#L198

horis233: Thanks for the explanation! I understand.

horis233 (approved):
/lgtm

PR #498, opened 2021-04-26
deploy cluster-scope in ControlNs for on-prem multi instance case

Daniel-Fan:
We have two templates for SaaS and on-prem deployment respectively.

In the on-prem deployment,
1. the Single instance will deploy all resources in `masterNs`. and `ControlNs == masterNs`.
2. the multi instances case will deploy cluster scope resource like `cert-manager-operator` and `licensing-operator` in `ControlNs`.

### DIFF
--- a/controllers/constant/odlm.go
+++ b/controllers/constant/odlm.go
@@ -106,7 +106,7 @@ metadata:
 spec:
   operators:
   - name: ibm-licensing-operator
-    namespace: {{ .MasterNs }}
+    namespace: {{ .ControlNs }}
     channel: {{ .Channel }}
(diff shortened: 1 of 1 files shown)


ZhuoxiLi:
If my understanding is correct, for single instance, we deploy all resources in `masterNs`. although we are not using `ControlNs` in single instance, we still set `ControlNs := masterNs`

horis233 (approved):
/lgtm

horis233:
> If my understanding is correct, for single instance, we deploy all resources in `masterNs`. although we are not using `ControlNs` in single instance, we still set `ControlNs := masterNs`

Yes, it's correct

PR #500, opened 2021-04-28
change env var ODLM_SCOPE to ISOLATED_MODE

Daniel-Fan:


testdata/deploy/deploy.yaml
```diff
@@ -14,409 +14,9 @@ spec:
       labels:
         name: ibm-common-service-operator
       annotations:
-        csV3OperandConfig: |-
```
ZhuoxiLi, line 17: May I ask the reason to delete those lines, since they are out-of-date for v3.8?

Daniel-Fan: Yes, in the latest 3.8, we will use template to render those yaml file.

We could see this PR for details https://github.com/IBM/ibm-common-service-operator/pull/480

horis233 (approved):
/lgtm

PR #501, opened 2021-04-29
add iamstatus with cloud pak prefix

Daniel-Fan:
Not 100 percent sure that I understand the design correctly.

Basically, I use `requested-from-namespace` (usually is cloud pak name) as the prefix of `-iamstatus`.
In this situation, one Common Service instance will have multiple namespaced cloud pak prefix `iamstatus` like `cp4i-iamstatus` and `cp4mcm-iamstatus` if the ConfigMap looks like this
```
apiVersion: v1
kind: ConfigMap
metadata:
  name: common-service-maps
  namespace: kube-public
data: 
  common-service-maps.yaml: |
    controlNamespace: cs-control
    namespaceMapping:
    - requested-from-namespace:
      - cp4i
      - cp4mcm
      map-to-common-service-namespace: common-service
```

Also there is a overall status called `iamstatus` as our previous release, it is only `Ready` when all other namespaced cloud pak prefix `-iamstatus` are `Ready` to keep the compatibility for the previous release.

### DIFF
--- a/controllers/check/check_iam.go
+++ b/controllers/check/check_iam.go
@@ -156,6 +156,9 @@ func createUpdateConfigmap(bs *bootstrap.Bootstrap, status string) error {
 				cm.Data[statusKey] = status
 				isUpdate = true
 			}
+		} else {
+			cm.Data[statusKey] = status
+			isUpdate = true
 		}
 	}
 
--- a/controllers/check/check_iam.go
+++ b/controllers/check/check_iam.go
@@ -18,6 +18,7 @@ package check
 
 import (
 	"context"
+	"regexp"
 	"time"
 
 	olmv1alpha1 "github.com/operator-framework/api/pkg/operators/v1alpha1"
@@ -28,8 +29,8 @@ import (
 	"k8s.io/apimachinery/pkg/types"
 	"k8s.io/klog"
 	"sigs.k8s.io/controller-runtime/pkg/client"
-	"sigs.k8s.io/controller-runtime/pkg/manager"
 
+	"github.com/IBM/ibm-common-service-operator/controllers/bootstrap"
 	util "github.com/IBM/ibm-common-service-operator/controllers/common"
 )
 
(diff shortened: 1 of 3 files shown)


Daniel-Fan:
@horis233 @ZhuoxiLi 
The PR has been tested, ready for reviewing.

controllers/check/check_iam.go
```diff
@@ -120,56 +118,104 @@ func getDeploymentStatus(r client.Reader, name string) string {
...
 		}
 		return err
 	}
-	if cm.Data["iamstatus"] != status {
-		klog.Infof("IAM status is %s", status)
+	isUpdate := false
+	for _, requestNs := range requestNsSlice {
+		statusKey := requestNs + "-iamstatus"
+		// check the iamstatus of cloud pak
+		if val, ok := cm.Data[statusKey]; ok {
```
horis233, line 153: What if ok is false? I believe we need to insert `requestNs + "-iamstatus"` into the configmap anyway.

Daniel-Fan: Nice catch!

I was thinking only insert the status when we create the this status Configmap.

But I ignored that the user will update the cs-mapping ConfigMap after the installation and it will introduce new namespaced cloud pak prefix `iamstatus`.

Thank you! It is updated.

controllers/check/check_iam.go
```diff
@@ -120,56 +118,107 @@ func getDeploymentStatus(r client.Reader, name string) string {
...
+	requestNsSlice := util.GetRequestNs(bs.Reader)
+	isUpdate := false
+	for _, requestNs := range requestNsSlice {
+		// update the iam status for each cloud pak which is using this common service
+		statusKey := requestNs + "-iamstatus"
+		if val, ok := cm.Data[statusKey]; ok {
+			if val != status {
+				klog.Infof("IAM status for namespace %s is %s", requestNs, status)
+				cm.Data[statusKey] = status
+				cm.Data["iamstatus"] = status
```
horis233, line 211: Should we update the `cm.Data["iamstatus"]` every time in a loop? That will cause it will always be the last status, isn't it?

horis233: Oh is it because "status" is always not ready?

Could we abstract this part into another function and call the function here?

```
			// iamstatus in this CS is ready, check iamstatus for all other CS
			reg, _ := regexp.Compile(`^(.*)\-iamstatus`)
			statusSlice := make([]string, 0)
			for key, status := range cm.Data {
				if reg.MatchString(key) {
					statusSlice = append(statusSlice, status)
				}
			}
			for _, status := range statusSlice {
				if status != "Ready" {
					cm.Data["iamstatus"] = status
					break
				}
			}
```

Daniel-Fan: Yes, the function `updateConfigMap` will always get `NotReady` parameter, indicating that if cloud pak `iamstatus` is setting to `NotReady`, the overall `iamstatus` will be `NotReady` anyway.

But I think it will be better to generalize the logic of `updateConfigMap`, in order to make it follow the same logic with `CreateOrUpdateConfigMap`.

> Could we abstract this part into another function and call the function here?

Sure, it is updated, abstracting it into another function.

Please review the above changes again, thank you!

horis233 (approved):
/lgtm

PR #504, opened 2021-05-02
Add pre-check of whether storageClass existing

ZhuoxiLi:
### why we need this
Since MongoDB requires StorageClass and many other common services require MongoDB, so we add storageClass pre-check before installation.

### test results
logs with 2 StorageClass:
Screen Shot 2021-05-02 at 6 38 16 PM

logs with No StorageClass (report error & stop deployment
Screen Shot 2021-05-02 at 6 39 40 PM

### issue
https://github.ibm.com/IBMPrivateCloud/roadmap/issues/44307

### DIFF
--- a/controllers/bootstrap/init.go
+++ b/controllers/bootstrap/init.go
@@ -180,6 +180,15 @@ func (b *Bootstrap) InitResources(instance *apiv3.CommonService) error {
 		}
 	}
 
+	// Check storageClass
+	csStorageClass, err := util.GetStorageClass(b.Reader)
+	if err == nil {
+		if err := util.ValidateStorageClass(csStorageClass); err != nil {
+			klog.Errorf("StorageClass is not found in current cluster")
+			return err
+		}
+	}
+
 	// Install Namespace Scope Operator
 	if err := b.installNssOperator(manualManagement); err != nil {
 		return err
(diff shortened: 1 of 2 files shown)


controllers/bootstrap/init.go
```diff
@@ -180,6 +180,15 @@ func (b *Bootstrap) InitResources(instance *apiv3.CommonService) error {
 		}
 	}
 
+	// Check storageClass
+	csStorageClass, err := util.GetStorageClass(b.Reader)
+	if err == nil {
```
Daniel-Fan, line 185: I think we should handle the situation that the error happens when we get storage class, for example
```
if err != nil {
     return fmt.Errorf("Failed to get StorageClass")
}
```
and after the error checking, we could validate the storage class

controllers/bootstrap/init.go
```diff
@@ -180,6 +180,15 @@ func (b *Bootstrap) InitResources(instance *apiv3.CommonService) error {
...
+		if err := util.ValidateStorageClass(csStorageClass); err != nil {
+			klog.Errorf("StorageClass is not found in current cluster")
+			return err
```
Daniel-Fan, line 188: I think it will be better if we could combine those two lines together to reduce the duplicated info in log, for example directly return the error, and in func `ValidateStorageClass`
```
if size <= 0 {
    return fmt.Errorf("StorageClass is not found in current cluster")
}
```

then the output of log will be
```
StorageClass Number: 0
Failed to initialize resources: StorageClass is not found in current cluster
Failed to reconcile ibm-common-services/common-service: StorageClass is not found in current cluster
```

Daniel-Fan (commented):
Also a suggestion that it would be better for us to not include the IBM Private Repo link in the public repo.

ZhuoxiLi:
### logs of new commit
Screen Shot 2021-05-03 at 12 58 15 PM

Daniel-Fan (approved):
/lgtm

PR #506, opened 2021-05-03
remove CatalogSource in workflow

Daniel-Fan:


Daniel-Fan:
Will close this PR.

When we meet the issue when running the github workflow, there is workaround in troubleshooting sheet inside installer folder.

bitscuit:
> 
> 
> Will close this PR.
> 
> When we meet the issue when running the github workflow, there is workaround in troubleshooting sheet inside installer folder.

It would be better if we can fix it automatically. I've been wanting to fix a couple issues with the workflow, but haven't gotten to it

PR #513, opened 2021-05-13
Add service_account.yaml under config

ZhuoxiLi:
we removed `ibm-common-service-operator/deploy/` in after v3.6.x. Thus, we need to relocate service_account.yaml under `config`.


Daniel-Fan (approved):
/lgtm

PR #514, opened 2021-05-14
feat: bootstrap cs operator in the manual approval mode

horis233:
When cloudpak operator is installed in the approval mode, then the common service operator in master namespace will be deployed in the manual approval mode.


horis233:
@Daniel-Fan @ZhuoxiLi This PR is available for reviewing.

controllers/bootstrap/init.go
```diff
@@ -813,7 +819,40 @@ func (b *Bootstrap) renderTemplate(objectTemplate string, data interface{}, alwa
 // 	return ""
 // }
 
-func (b *Bootstrap) updateOperandRegistry(installPlanApproval olmv1alpha1.Approval) error {
+func (b *Bootstrap) UpdateCsOpApproval() error {
+	sub := &olmv1alpha1.Subscription{}
+	subKey := types.NamespacedName{
+		Name:      "ibm-common-service-operator",
```
Daniel-Fan, line 825: I noticed that we define `IBMCSPackage = "ibm-common-service-operator"` in `constant.go`
Maybe we should this variable to replace the string here?

horis233: This one should be fine, it is the name of the subscription. I will replace CS packagenames with IBMCSPackage in the following comments.

Daniel-Fan (approved):
/lgtm

PR #515, opened 2021-05-21
feat: 1. bootstrap fail will exist operator. 2. remove registry from …

horis233:
…OperandRegistry, ODLM will handle the catalog finding works


ZhuoxiLi (approved):
/lgtm

PR #519, opened 2021-05-27
[WIP] sync up NSS CR for common service and ODLM

Daniel-Fan:
It creates an additional go routine to sync up the `common-service` and `nss-odlm-scope` NSS CR.

Currently, we use `common-service` NSS CR to control the namespace scope for individual common service operator and use `nss-odlm-scope` to control the namespace scope for ODLM.

When user manually extend their operator's permission to other namespaces, they will be instructed about adding namespace in `common-service` NSS CR.

Therefore, we need to sync up the namespace member in both `common-service` and `nss-odlm-scope` NSS CR, to ensure that ODLM's permission is be extended as well.

The code part is finished, need to do some test before merging it.

### DIFF
--- a/controllers/namespacescope/namespacescope.go
+++ b/controllers/namespacescope/namespacescope.go
@@ -20,6 +20,7 @@ import (
 	"context"
 	"time"
 
+	gset "github.com/deckarep/golang-set"
 	"k8s.io/apimachinery/pkg/types"
 	utilwait "k8s.io/apimachinery/pkg/util/wait"
 	"k8s.io/klog"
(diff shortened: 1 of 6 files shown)


controllers/namespacescope/namespacescope.go
```diff
@@ -0,0 +1,117 @@
...
+
+		targetNsScope := &nssv1.NamespaceScope{}
+		targetNsScopeKey := types.NamespacedName{Name: targetCR, Namespace: bs.CSData.MasterNs}
+		if err := bs.Reader.Get(ctx, targetNsScopeKey, targetNsScope); err != nil {
+			klog.Errorf("Failed to get NSS CR %s: %v, retry again", targetNsScopeKey.String(), err)
+			continue
+		}
+
+		// sync up when namepsace in source CR is different from target CR
+		if len(sourceNsScope.Spec.NamespaceMembers) != len(targetNsScope.Spec.NamespaceMembers) {
```
horis233, line 80: Using the length to check the namespace item is not very accurate.

Could we leverage "github.com/deckarep/golang-set"  to do the comparison?

It has a function ["Difference"](https://pkg.go.dev/github.com/deckarep/golang-set@v1.7.1#Set.Difference) to get the difference of two sets.

main.go
```diff
@@ -164,6 +165,8 @@ func main() {
 		go check.IamStatus(bs)
 		// Generate Issuer and Certificate CR
 		go certmanager.DeployCR(bs)
+		// Sync up NSS CR
+		go nss.SyncUpCR(bs)
```
horis233, line 169: I am wondering if we can enable the nss sync up after creating the nss CRs.

controllers/namespacescope/namespacescope.go
Daniel-Fan: It sounds good, by using the set, we will have a precise comparison.

main.go
Daniel-Fan: In the routine, we will wait to sync up until the CRD and the CR itself is ready.
https://github.com/IBM/ibm-common-service-operator/pull/519/files#diff-1b166bbb2715d4ce512c9bea4ae40667c8db17192f7e04d8db3afacfea5755a6R53

I am thinking that if we move this part into `init.go`, it may be not obvious for what additional routine we created in the future code review.

controllers/namespacescope/namespacescope.go
Daniel-Fan: @horis233 @ZhuoxiLi 

The PR is ready for reviewing.

In order to keep the consistency of two NSS CRs, I choose to overwrite the entire `nss-odlm-scope` namespace members by `common-service` members once they are different.

In this case, the changes in `nss-odlm-scope` CR will not take effect and the `common-service` CR is the only one which will be changed by users.

controllers/namespacescope/namespacescope.go
```diff
@@ -0,0 +1,122 @@
...
+			sourceNsSet.Add(ns)
+		}
+		for _, ns := range targetNsScope.Spec.NamespaceMembers {
+			targetNsSet.Add(ns)
+		}
+
+		// sync up when namepsace in source CR is different from target CR
+		if !sourceNsSet.Equal(targetNsSet) {
+			sourcenNsMems := sourceNsSet.ToSlice()
+			var targetNsMems []string
```
horis233, line 93: Why we create an empty slice here? Could we add `sourcenNsMems` into set `targetNsSet`?

controllers/namespacescope/namespacescope.go
horis233: Will the overwriting the entire nss-odlm-scope cause they are always the same? It will impact the multiple instances installation, right?

controllers/namespacescope/namespacescope.go
Daniel-Fan: The return type of `sourceNsSet.ToSlice()` is `[]interface{}`, but `targetNsScope.Spec.NamespaceMembers` has the type `[]string`, so we have to convert each item to `string` first.

`targetNsMems = append(targetNsMems, ns.(string))`

Daniel-Fan: Also, I made some changes to the PR to add the namespaces in `common-service` to `odlm-scope` instead of overwriting it.

horis233 (approved):
/lgtm

PR #520, opened 2021-05-31
fetch the namespace from namespace-scope ConfigMap for iam status

Daniel-Fan:
Previously, we fetch the namespaces from CS Mapping, but it will contain some namespaces that have not been created.

Currently, we choose to fetch the namespaces from `namespace-scope` ConfigMap whose namespaces have been validated by namespace-scope operator.

Fyre cluster is down, so I haven't tested the functionality yet.


horis233 (approved):
/lgtm

PR #521, opened 2021-05-31
rm ibm-monitoring-exporters-operator and ibm-monitoring-prometheusext…

ZhuoxiLi:
…-operator

### DIFF
--- a/common/scripts/upgrade_to_continuous_delivery.sh
+++ b/common/scripts/upgrade_to_continuous_delivery.sh
@@ -84,6 +84,8 @@ function hash_operator() {
         "ibm-cert-manager-operator") eval "$2='3.8.0'";;
         "ibm-mongodb-operator") eval "$2='1.2.0'";;
         "ibm-iam-operator") eval "$2='3.8.0'";;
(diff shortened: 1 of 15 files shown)


common/scripts/upgrade_to_continuous_delivery.sh
```diff
@@ -84,8 +84,6 @@ function hash_operator() {
         "ibm-cert-manager-operator") eval "$2='3.8.0'";;
         "ibm-mongodb-operator") eval "$2='1.2.0'";;
         "ibm-iam-operator") eval "$2='3.8.0'";;
-        "ibm-monitoring-exporters-operator") eval "$2='1.9.0'";;
-        "ibm-monitoring-prometheus-operator-ext") eval "$2='1.9.0'";;
```
Daniel-Fan, line 88: This is for the upgrade from EUS to continuous delivery like 3.7 or 3.8.

It is used to check whether the operator has been upgraded to version 3.6. So we should not delete it.

ZhuoxiLi: Thanks, I will revert them.

horis233 (approved):
/lgtm

PR #523, opened 2021-06-02
filter the template size when the CR is out of scope

Daniel-Fan:
We will always return the size template in function `applySizeTemplate()`.

Therefore, when it is out of scope, we should remove those namespace-scoped operator from template, and then return it as the `newConfigs`


ZhuoxiLi (approved):
/lgtm

PR #524, opened 2021-06-02
update the channel of the db2u operator

horis233:
update the channel of the db2u operator from `v1.0` to `v1.1`


Daniel-Fan (approved):
/lgtm

PR #525, opened 2021-06-02
add additinal parameter for MultiInstancesEnable

Daniel-Fan:


controllers/common/util.go
```diff
@@ -368,6 +368,12 @@ func CheckSaas(r client.Reader) (enable bool) {
 	return true
 }
 
+// CheckMultiInstance checks whether it is a MultiInstances including SaaS and on-prem MultiInstances
+func CheckMultiInstances(r client.Reader) (enable bool) {
+	controlNs := GetControlNs(r)
+	return len(controlNs) > 0
```
ZhuoxiLi, line 374: I think for single instance, we will assign controlNs = MasterNs, so the length will also greater than 0?

Daniel-Fan: Yes, but not in this case.

We get the controlNs directly from cs-mapping, it will remain `empty` in ConfigMap
https://github.com/IBM/ibm-common-service-operator/blob/master/controllers/common/util.go#L372

ZhuoxiLi (approved):
/lgtm

PR #526, opened 2021-06-02
enable `multipleInstancesEnabled` in the management-ingress & ingress-nginx CR

ZhuoxiLi:
Tested in single-instance env & multi-instance env, both worked as expected.

### DIFF
--- a/controllers/multipleInstancesEnabled/multipleInstancesEnabled.go
+++ b/controllers/multipleInstancesEnabled/multipleInstancesEnabled.go
@@ -1,0 +1,28 @@
+//
+// Copyright 2021 IBM Corporation
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+// http://www.apache.org/licenses/LICENSE-2.0
+//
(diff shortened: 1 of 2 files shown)


Daniel-Fan (approved):
It looks good to me.

But I am wondering whether we could gather all the templates into one package like `constant`. Currently we already have several directories that only contain the constant template.

controllers/multipleInstancesEnabled/multipleInstancesEnabled.go
```diff
@@ -0,0 +1,28 @@
...
+
+const MultipleInstancesEnabledTemplate = `
+- name: ibm-management-ingress-operator
+  spec:
+    managementIngress:
+      multipleInstancesEnabled: placeholder
+- name: ibm-ingress-nginx-operator
+  spec:
+    nginxIngress:
+      multipleInstancesEnabled: placeholder
```
horis233, line 27: We don't need to set it for `ibm-ingress-nginx-operator`

ZhuoxiLi: removed

horis233 (approved):
/lgtm

PR #528, opened 2021-06-04
update comments and add Dependabot to update base image

Daniel-Fan:
In order to trigger a rebuild to fetch the latest base image


horis233 (approved):
/lgtm

PR #530, opened 2021-06-04
find nssConfigMap in masterNs instead of operatorNs

Daniel-Fan:
To adopt the all namespace mode, the NSS ConfigMap will always be in `MasterNs` instead of `OperatorNs`


horis233 (approved):
/lgtm

PR #532, opened 2021-06-11
pre-check storage class, if nothing exists, give warning messages

ZhuoxiLi:
verified with:
1. cluster with sc
2. cluster without sc

### DIFF
--- a/controllers/bootstrap/init.go
+++ b/controllers/bootstrap/init.go
@@ -196,11 +196,13 @@ func (b *Bootstrap) InitResources(instance *apiv3.CommonService) error {
 	// Check storageClass
 	csStorageClass, err := util.GetStorageClass(b.Reader)
 	if err != nil {
-		return fmt.Errorf("failed to get StorageClass")
+		klog.Info("failed to get StorageClass")
+		// return fmt.Errorf("failed to get StorageClass")
 	}
-	if err := util.ValidateStorageClass(csStorageClass); err != nil {
-		return err
-	}
+	util.ValidateStorageClass(csStorageClass)
+	// if err := util.ValidateStorageClass(csStorageClass); err != nil {
+	// 	return err
+	// }
 
 	// Install Namespace Scope Operator
 	if err := b.installNssOperator(manualManagement); err != nil {
--- a/controllers/bootstrap/init.go
+++ b/controllers/bootstrap/init.go
@@ -196,7 +196,7 @@ func (b *Bootstrap) InitResources(instance *apiv3.CommonService) error {
 	// Check storageClass
 	csStorageClass, err := util.GetStorageClass(b.Reader)
 	if err != nil {
-		klog.Info("failed to get StorageClass")
+		return err
 	}
 	util.ValidateStorageClass(csStorageClass)
 
--- a/controllers/bootstrap/init.go
+++ b/controllers/bootstrap/init.go
@@ -194,11 +194,9 @@ func (b *Bootstrap) InitResources(instance *apiv3.CommonService) error {
 	}
 
 	// Check storageClass
-	csStorageClass, err := util.GetStorageClass(b.Reader)
-	if err != nil {
+	if err := util.CheckStorageClass(b.Reader); err != nil {
 		return err
 	}
-	util.ValidateStorageClass(csStorageClass)
 
 	// Install Namespace Scope Operator
 	if err := b.installNssOperator(manualManagement); err != nil {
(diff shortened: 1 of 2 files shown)


controllers/bootstrap/init.go
```diff
@@ -196,11 +196,9 @@ func (b *Bootstrap) InitResources(instance *apiv3.CommonService) error {
 	// Check storageClass
 	csStorageClass, err := util.GetStorageClass(b.Reader)
 	if err != nil {
-		return fmt.Errorf("failed to get StorageClass")
-	}
-	if err := util.ValidateStorageClass(csStorageClass); err != nil {
-		return err
+		klog.Info("failed to get StorageClass")
```
Daniel-Fan, line 199: I think we should keep `return fmt.Errorf("failed to get StorageClass")` here. I think it is the error that we failed to List storageclass from the cluster, not because there is no Storageclass avaliable.

Also because we will only inform the user, not stop the reconcile.

I think we need to add a separate the error check here
one is `errors.IsNotFound(err)` and the other is the remaining errors
https://github.com/IBM/ibm-common-service-operator/blob/master/controllers/common/util.go#L225

/cc @horis233 Please correct me if it is not correct.

ZhuoxiLi: I see. So `return fmt.Errorf("failed to get StorageClass")` this error indicates `not found StorageClass attribute` right? not about StorageClass instance.

Daniel-Fan: yeah, kind of.

The error may happen when there is issue to communicate with kube apiserver. it could be permission error, network err or something else. It will be better to return that error as well.
`return fmt.Errorf("failed to get StorageClass: %v", err)`

ZhuoxiLi: make sense, if those kinds of error happens, do we want to stop the reconcile or continue? I think if we `return fmt.Errorf("failed to get StorageClass: %v", err)` in controllers/bootstrap/init.go, since **return**, it will stop the reconcile.

Daniel-Fan: No, we wanna continue the reconciliation.

If the reconciler receive the return error, it will requeue the request automatically.

ZhuoxiLi: Thanks for the explanation! Updated with we discussed above, please review.

controllers/common/util.go
```diff
@@ -222,21 +222,25 @@ func GetCmOfMapCs(r client.Reader) (*corev1.ConfigMap, error) {
...
-		return nil, err
+	if err == nil {
+		return csStorageClass, nil
 	}
-	return csStorageClass, nil
+
+	if errors.IsNotFound(err) {
+		return nil, fmt.Errorf("storageClassList is not found")
+	}
+
```
Daniel-Fan, line 232: Sorry for the misleading here, I found that `errors.IsNotFound(err)` is not accurate here.

Usually, we will check `errors.IsNotFound(err)` when we want to use `r.Get()` to get some specific resource. It will have the case that the specific resource is not found. But this does not apply to `r.List()` here. Therefore, we could remove it.


And a little advice here.

I am wondering whether we could combine the `ValidateStorageClass` and `GetStorageClass` together called `CheckStorageClass`.

In this case, we don't need to return the `StorageClass`,
and after we List the `storageClass`, check its length directly.
```
err := r.List(context.TODO(), csStorageClass)
if err != nil {
    // hitting error when we try to List storageclass
    // and we indicate what exactly error is
    return fmt.Errorf("fail to list storageClass: %v", err)
}

// after list the storageClass, we check the length and print the proper info
return nil
```

ZhuoxiLi: Updated, please review, thanks!

Daniel-Fan (approved):
/lgtm

controllers/common/util.go
```diff
@@ -218,23 +218,19 @@ func GetCmOfMapCs(r client.Reader) (*corev1.ConfigMap, error) {
...
-}
 
-// ValidateStorageClass validates whether the StorageClass exist
-func ValidateStorageClass(csStorageClass *storagev1.StorageClassList) error {
 	size := len(csStorageClass.Items)
 	klog.Info("StorageClass Number: ", size)
 
 	if size <= 0 {
-		return fmt.Errorf("storageClass is not found in current cluster")
+		klog.Info("storageClass is not found in current cluster")
```
horis233, line 233: It's better to use `klog.Warning`

pgodowski: Can we update the message to indicate problem statement rather then the fact no storage is available?
I.e. message saying that storage might be required by other components for proper functioning, thus refer Cloudpak's documentation against prerequisites?

PR #534, opened 2021-06-11
update operator-sdk to 1.8.0

ZhuoxiLi:


bundle/manifests/ibm-common-service-operator.clusterserviceversion.yaml
```diff
@@ -327,5 +327,8 @@ spec:
   maturity: alpha
   provider:
     name: IBM
+  relatedImages:
+    - image: quay.io/opencloudio/common-service-operator:latest
+      name: ibm-common-service-operator.v0.0.0
```
horis233, line 332: Could we use a real image version, like
```
quay.io/opencloudio/common-service-operator@sha256:sha256:e7f42050d54530c708f162abcf04aae1b63ce6f5ca6e97af81a4a83f6ae708f6
ibm-common-service-operator.v3.8.0
```

ZhuoxiLi: sure, we could use this. The question is if we are using a specific version image, we will need to update the tag/sha every release. That will be some duplicated work. Any suggestions on that?

Daniel-Fan: Could we use tag instead of digest?

So it will be updated every time when we bump the version.

ZhuoxiLi: that sounds good to me.

horis233: We need to use digest here: https://docs.openshift.com/container-platform/4.7/operators/operator_sdk/osdk-generating-csvs.html#olm-enabling-operator-for-restricted-network_osdk-generating-csvs

Is there any difference from using digest and tag? We still need to update it every release.

Daniel-Fan: If we use digest here, we need to update it before closing to GA, after the image has been published on quay.io.
But it seems that we does not many choices

ZhuoxiLi: Updated with the digest, thank you all for the discussion here. Please review again,

horis233: I get what you mean. Then I believe we can hold this PR until we can get some help from the CICD

horis233: @ZhuoxiLi 
Please revert the changes relevant to related images and merge this PR

ZhuoxiLi:
@horis233 @Daniel-Fan this pr now only contains the change of `update operator-sdk to 1.8.0`, after we merged this pr, please also upgrade your local operator-sdk to 1.8.0. thanks!

Daniel-Fan (approved):
/lgtm

PR #536, opened 2021-06-14
iam-status enhancement

Daniel-Fan:
No IAM subscriptions:
* cs-mapping is not found(pure single mode)
     * clean up cs.data
* cs-mapping is found(both single or multi)
     * remove namespace of the scope(get cs-mapping) from iam-status

Exist IAM subscription:
1. Create ConfigMap if not found
2. Update ConfigMap
    * cs-mapping is not found(pure single mode)
         * overwrite the cs.data with current nss ConfigMap
    * cs-mapping found
         * Update Ns in Status with ns in cs mapping


horis233 (approved):
/lgtm

PR #537, opened 2021-06-16
handle interface {} is nil

Daniel-Fan:
`newConfigForOperator` and `opService` may be `nil`, not `map[string]interface{}`


ZhuoxiLi (approved):
/lgtm

PR #539, opened 2021-06-16
bump master branch to 4.0.0 for future release

Daniel-Fan:
Master branch will be used for the future pipeline going forward.

We can put future dev code into these builds as long as your operator is set to a higher version than any of the versions in `eus`, `efix`, and `cd` branches.


horis233 (approved):
/lgtm

PR #556, opened 2021-06-28
prioritize the dev catalogsource

Daniel-Fan:
The dev build with `opencloud-operator` CatalogSource has the highest priority.

When there is no dev build in the cluster, it will follow the previous strategy that we will choose `ibm-operator-catalog` if it exists.

But there is a permission escalation happening, `cs-operator` requires the cluster permission for getting the CatalogSource from `openshift-marketplace` namespace in order to check the image used by `opencloud-operator`. The `packageManifest` of `cs-operator` in its own namespace is not really helpful.


Daniel-Fan:
@horis233 @ZhuoxiLi 
It is ready for reviewing.

horis233 (approved):
/lgtm

PR #558, opened 2021-06-29
uninstall cert-manager after channel switch

Daniel-Fan:
Currently when we upgrade from CS 3.6 or below to CS 3.8, the version 3.10 cert-manager-operator in CS 3.8 will use the version 3.9 cert-manager-operator in CS 3.7 as a bridge, it is not a direct upgrade.

In AirGap situation, it will cause the `ImagePullBackOff` error for version 3.9 cert-manager-operator, because we did not include the images about CS 3.7 for CS 3.8 bundle.

Therefore, we will delete the cert-manager-operator before the upgrade, and ODLM will re-install the version 3.10 cert-manager-operator directly after the upgrade.


Daniel-Fan:
/hold

Hold the PR until the SERT is fully tested.

horis233 (approved):
/lgtm

PR #559, opened 2021-07-05
remove checking installedcsv with currentcsv for manual approval strategy

Daniel-Fan:
Avoid the situation that the user does not wanna upgrade when there is a newer version coming, but cs-operator continue waiting for others' upgrading because the currentCSV is larger than installedCSV.


ZhuoxiLi (approved):
/lgtm

PR #576, opened 2021-07-19
Make probe more loose

horis233:
CS operator expects its container can be ready within 1 second, but it can’t in the machine with bad performance.
We need to make to probe more loose to prevent this error


Daniel-Fan (approved):
/lgtm

PR #578, opened 2021-07-20
allow all cs cr in any namespaces can enable crossplane operator

ZhuoxiLi:
Here is my design idea, feel free to give any suggestions.

### DIFF
--- a/controllers/bootstrap/init.go
+++ b/controllers/bootstrap/init.go
@@ -157,6 +157,25 @@ func NewBootstrap(mgr manager.Manager) (bs *Bootstrap, err error) {
 	return
 }
 
+// Crossplane install crossplane operator when bedrockshim is true
+func (b *Bootstrap) Crossplane(instance *apiv3.CommonService) error {
+	// Install Crossplane Operator
+	bedrockshim := false
+	if instance.Spec.Features != nil {
+		if instance.Spec.Features.Bedrockshim != nil {
+			bedrockshim = instance.Spec.Features.Bedrockshim.Enabled
+		}
+	}
+
+	if bedrockshim {
+		if err := b.installCrossplaneOperator(); err != nil {
+			return err
+		}
+	}
+
+	return nil
+}
+
 // InitResources initialize resources at the bootstrap of operator
 func (b *Bootstrap) InitResources(instance *apiv3.CommonService) error {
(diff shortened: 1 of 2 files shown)


Daniel-Fan (commented):
The overall code structure looks great.

But I am confused that since `instance.Spec.Features.Bedrockshim` is controlling the installation of both `crossplane operator` and `cloud operator`. Should we include `cloud operator` installation in the reconciliation of all CRs?
If `Bedrockshim=true && SaasEnable=true` happens in any one of the CRs, not only in the `Master` CR, then we should install `cloud operator`.

The other point is we could find that in both `ReconcileMasterCR` and `ReconcileGeneralCR`, there is a parameter called `inScope`.
`inScope` is used to indicate whether the current reconciled CR is in the scope of cs-operator. Because in SaaS or multi-instance deployment, each cs-operator will still reconcile all the CommonService CRs in the cluster in order to get the configuration of cluster-scope resources in each CR like cert-manager and update its own namespace-scoped `OperandConfig`.

Therefore, when we check the `Bedrockshim` in the current CR, we should first check whether the CR is in the scope. If it is not in the scope, we should skip the installation.

horis233:
> The overall code structure looks great.
> 
> But I am confused that since `instance.Spec.Features.Bedrockshim` is controlling the installation of both `crossplane operator` and `cloud operator`. Should we include `cloud operator` installation in the reconciliation of all CRs?
> If `Bedrockshim=true && SaasEnable=true` happens in any one of the CRs, not only in the `Master` CR, then we should install `cloud operator`.
> 
> The other point is we could find that in both `ReconcileMasterCR` and `ReconcileGeneralCR`, there is a parameter called `inScope`.
> `inScope` is used to indicate whether the current reconciled CR is in the scope of cs-operator. Because in SaaS or multi-instance deployment, each cs-operator will still reconcile all the CommonService CRs in the cluster in order to get the configuration of cluster-scope resources in each CR like cert-manager and update its own namespace-scoped `OperandConfig`.
> 
> Therefore, when we check the `Bedrockshim` in the current CR, we should first check whether the CR is in the scope. If it is not in the scope, we should skip the installation.

Thanks for your comment. I think so.

ZhuoxiLi:
Thank you for the suggestions! Please review again @horis233 @Daniel-Fan

Daniel-Fan (approved):
/lgtm

PR #580, opened 2021-07-23
Reduce subscription update

ZhuoxiLi:


controllers/constant/commonservice.go
```diff
@@ -22,8 +22,6 @@ kind: Subscription
 metadata:
   name: ibm-common-service-operator
   namespace: {{ .MasterNs }}
-  annotations:
```
horis233, line 25: @ZhuoxiLi Please notice we need to remove the version from ODLM and namespacescope operator as well

controllers/common/util.go
```diff
@@ -60,6 +60,15 @@ type nsMapping struct {
 	CsNs      string   `json:"map-to-common-service-namespace"`
 }
 
+type OperatorSub struct {
+	Channel string `json:"channel"`
+	// InstallPlanApproval olmv1alpha1.Approval `json:"installPlanApproval,omitempty"`
+	InstallPlanApproval string `json:"installPlanApproval,omitempty"`
+	Name                string `json:"name"`
+	Source              string `json:"source,omitempty"`
+	SourceNamespace     string `json:"sourceNamespace,omitempty"`
```
horis233, line 69: Other than these specs, we need to compare the subscription config as well.

ZhuoxiLi:
@horis233 @Daniel-Fan this pr is ready to review, thanks!

controllers/bootstrap/init.go
```diff
@@ -438,8 +438,37 @@ func (b *Bootstrap) CreateOrUpdateFromYaml(yamlContent []byte, alwaysUpdate ...b
...
+			opt.Source = util.GetNestedString(obj, "spec", "source").(string)
+			opt.SourceNamespace = util.GetNestedString(obj, "spec", "sourceNamespace").(string)
+			optEnv := util.GetNestedSlice(obj, "spec", "config", "env")
+			var optEnvSlice []map[string]interface{}
+			for _, s := range optEnv {
+				optEnvSlice = append(optEnvSlice, s.(map[string]interface{}))
+			}
+			if len(optEnvSlice) > 0 {
+				opt.InstallScope = optEnvSlice[0]["value"].(string)
+				opt.IsolatedMode = optEnvSlice[1]["value"].(string)
```
horis233, line 460: It is good for now, but if the length of optEnvSlice is 1, then we will get out of range issue.

ZhuoxiLi: Good catch! but `optEnvSlice` is from the subscription template. for now, there are either 2 items or none.
I can add an extra length check or add a TODO comment for the future if the template is changed. 
which way do we prefer?

controllers/bootstrap/init.go
```diff
@@ -438,8 +438,37 @@ func (b *Bootstrap) CreateOrUpdateFromYaml(yamlContent []byte, alwaysUpdate ...b
...
+			}
+
+			if compareSub(sub, opt) {
+				update = true
```
Daniel-Fan, line 464: Should we add an `else` here? Because subscription of NSS operator is forced updated. the initial value of `update` is `true`

Daniel-Fan: Or we could return the value from `compareSub(sub, opt)` and assign it to `update`

controllers/bootstrap/init.go
Daniel-Fan: I would prefer the first one. Because in subscription of NSS operator, we do not have those kind of Config Variables.

horis233:
\lgtm @Daniel-Fan please review again

Daniel-Fan (approved):
/lgtm

PR #587, opened 2021-07-29
Uds onboarding

Shrivastava-Varsha:
This PR is to onboard User Data Services (UDS) as a common service for Bedrock.
We are planning to onboard in Sept release.


controllers/constant/odlm.go
```diff
@@ -240,7 +245,14 @@ spec:
...
     installPlanApproval: {{ .ApprovalMode }}
-`
+  - channel: {{ .Channel }}
+    name: user-data-services-operator
+    namespace: {{ .MasterNs }}
+    packageName: user-data-services-operator-certified
+    scope: public
+    installPlanApproval: {{ .ApprovalMode }}
+    sourceName: {{ .CatalogSourceName }}
+    sourceNamespace: {{ .CatalogSourceNs }}
```
horis233, line 255: @shikhasriva 

I believe `user-data-services` has its own catalogsource in the airgap installation. Then we don't need to add 
```
    sourceName: {{ .CatalogSourceName }}
    sourceNamespace: {{ .CatalogSourceNs }}
```

horis233:
Since it's for September release, adding hold label to this PR
/hold

controllers/constant/odlm.go
Shrivastava-Varsha: Hi @horis233 I didn't get your comment. Why other services in registry are having catalogsource ?

Daniel-Fan: Hi @Shrivastava-Varsha 

For the other default Common Service services, they are sharing the same CatalogSource, for instance `opencloud-operator` CatalogSource.

When the code is running, the template `sourceName: {{ .CatalogSourceName }}` will be replace by `sourceName: `opencloud-operator`. We add it for convenience.

Therefore, if other services are using their own CatalogSource, we should leave it blank here. And cs-operator will look at their `PackageManifest` based on services name, then find out the CatalogSource containing those operators.

horis233: Hi @Shrivastava-Varsha
catalog source fields are only required for the services using the operator catalog pipeline as ibm common service operator.  We add these fields for compatibility consideration. 

They are not the required fields for new onboarded services.

controllers/constant/odlm.go
```diff
@@ -240,7 +245,14 @@ spec:
     packageName: db2u-operator
     scope: public
     installPlanApproval: {{ .ApprovalMode }}
-`
```
horis233, line 243: @Shrivastava-Varsha Please add the back quote at the end of your update

Varsha-Shrivastava: I added the changes to my base branch "Shrivastava-Varsha:uds-onboarding" but those are not reflecting here. Do I need to create a new PR @horis233 ?

Shrivastava-Varsha: fixed it. its so tiny that I overlooked it.

horis233 (approved):
/lgtm

PR #596, opened 2021-08-10
Enrich cs cr status - display subscription status & installplan

ZhuoxiLi:


controllers/goroutines/operator_status.go
```diff
@@ -82,37 +83,64 @@ func UpdateCsCrStatus(bs *bootstrap.Bootstrap) {
...
+		currentCSV := sub.Status.CurrentCSV
+		installedIsLarger := util.CompareVersion(installedCSV, currentCSV)
+		currentIsLarger := util.CompareVersion(currentCSV, installedCSV)
+		if !installedIsLarger && !currentIsLarger {
+			// installedCSV == currentCSV
+			opt.SubscriptionStatus = "Succeeded"
+		} else if currentIsLarger {
+			opt.SubscriptionStatus = "Upgrade available"
+		} else {
+			klog.Warning("installedCSV version is larger than currentCSV")
```
Daniel-Fan, line 136: Could we indicator which operator hits the error here?

controllers/goroutines/operator_status.go
```diff
@@ -82,37 +83,64 @@ func UpdateCsCrStatus(bs *bootstrap.Bootstrap) {
...
+		}
+	}
+
 	if len(csv.Status.Conditions) > 0 {
```
Daniel-Fan, line 140:
```
if err := bs.Reader.Get(ctx, csvKey, csv); err != nil {
	klog.Warningf("%s CSV %s", name, err)
}
```

Since we did not return error above when we try to get csv, I am wondering whether we will hit `nil` pointer issue when we check `len(csv.Status.Conditions)`?

I think maybe we could leave `opt.Status` blank if we failed to get CSV instead of directly return the error?

controllers/goroutines/operator_status.go
```diff
@@ -82,37 +83,64 @@ func UpdateCsCrStatus(bs *bootstrap.Bootstrap) {
...
+	opt.InstallPlanName = installplanName
+
+	// determinate subscription status
+	if installplanName == "" {
+		opt.SubscriptionStatus = "Failed"
+		opt.InstallPlanName = "Not Found"
+	} else {
+		currentCSV := sub.Status.CurrentCSV
+		installedIsLarger := util.CompareVersion(installedCSV, currentCSV)
+		currentIsLarger := util.CompareVersion(currentCSV, installedCSV)
```
horis233, line 129: @ZhuoxiLi 

It is not accurate to compare the version of installedCSV and currentCSV to decide the operator status since numerically, operator can be upgraded to a CSV with a smaller number. OLM doesn't manage the upgrade by CSV number.

Could get the `opt.SubscriptionStatus` directly from subscription status. For example, if installedCSV equal to currentCSV then it is Succeeded, otherwise `opt.SubscriptionStatus=sub.status.state`

controllers/goroutines/operator_status.go
horis233: csv is pointer in `csv.Status.Conditions`. We will hit nil pointer panic.

ZhuoxiLi: good catch, I will update the checking

horis233: Sorry, I was wrong, csv will never be nil, since it's initialized in `csv := &olmv1alpha1.ClusterServiceVersion{}`.
Thus we won't get nil pointer issue, like https://play.golang.org/p/D876UJFcoYi
But it's better to make it more clear in the code.

ZhuoxiLi:
@horis233 @Daniel-Fan updated according to the above comments, thanks!

controllers/goroutines/operator_status.go
Daniel-Fan: @ZhuoxiLi 
`installedCSV` and `currentCSV` are strings.
Could we use `installedCSV == currentCSV` and bypass the `util.CompareVersion()`?

controllers/goroutines/operator_status.go
```diff
@@ -82,37 +83,62 @@ func UpdateCsCrStatus(bs *bootstrap.Bootstrap) {
...
 	csv := &olmv1alpha1.ClusterServiceVersion{}
 	csvKey := types.NamespacedName{
-		Name:      installCSV,
+		Name:      installedCSV,
 		Namespace: namespace,
 	}
-	if csvErr := bs.Reader.Get(ctx, csvKey, csv); csvErr != nil {
-		return opt
+	if err := bs.Reader.Get(ctx, csvKey, csv); err != nil {
+		klog.Warningf("%s CSV %s", name, err)
```
Daniel-Fan, line 112: Also a little advice here, we could use `klog.Warningf("Failed to get %s CSV: %s", name, err)` to be consistent with other error msg.

controllers/goroutines/operator_status.go
ZhuoxiLi: ah, that's true, since we used need to know whether currentCSV > installedCSV, so need util.CompareVersion(). I will update it with compare the string.
other than that, I prefer to keep CompareVersion() in util, more flexible for future use.

Daniel-Fan (approved):
It looks great to me!

PR #607, opened 2021-08-19
use default path for two probes

Daniel-Fan:
The default Readiness probe endpoint name is `readyz` and default Liveness probe endpoint name is "healthz"
https://github.com/kubernetes-sigs/controller-runtime/blob/release-0.6/pkg/manager/manager.go#L187

Also try to align it with the example in kubebuilder
```
    // +kubebuilder:scaffold:builder

    if err := mgr.AddHealthzCheck("healthz", healthz.Ping); err != nil {
        setupLog.Error(err, "unable to set up health check")
        os.Exit(1)
    }
    if err := mgr.AddReadyzCheck("readyz", healthz.Ping); err != nil {
        setupLog.Error(err, "unable to set up ready check")
        os.Exit(1)
    }
```
https://book.kubebuilder.io/cronjob-tutorial/empty-main.html

### DIFF
--- a/bundle/manifests/ibm-common-service-operator.clusterserviceversion.yaml
+++ b/bundle/manifests/ibm-common-service-operator.clusterserviceversion.yaml
@@ -330,7 +330,7 @@ spec:
                     livenessProbe:
                       failureThreshold: 10
                       httpGet:
-                        path: /readyz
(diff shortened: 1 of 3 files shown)


horis233 (approved):
/lgtm

PR #608, opened 2021-08-19
uninstall crossplane operator & cloud operator when delete cs cr

ZhuoxiLi:


controllers/bootstrap/init.go
```diff
@@ -185,11 +185,33 @@ func (b *Bootstrap) CrossplaneCloudOperator(instance *apiv3.CommonService) error
...
+		}
 	}
 
 	return nil
 }
 
+// DeleteCrossplaneCloudSubscription deleted crossplane & cloud operator subscription when bedrockshim set to false or CS CR is removed
+func (b *Bootstrap) DeleteCrossplaneCloudSubscription(namespace string) error {
+	klog.Infof("Trying to delete ibm-crossplane-operator in %s", namespace)
+	if err := b.deleteSubscription("ibm-crossplane-operator-app", namespace); err != nil {
```
Daniel-Fan, line 200: I think before we delete the crossplane-operator, we should remove its CRs as well.

controllers/commonservice_controller.go
```diff
@@ -89,6 +89,9 @@ func (r *CommonServiceReconciler) Reconcile(req ctrl.Request) (ctrl.Result, erro
 			if err := r.handleDelete(); err != nil {
 				return ctrl.Result{}, err
 			}
+			if err := r.Bootstrap.DeleteCrossplaneCloudSubscription(r.Bootstrap.CSData.MasterNs); err != nil {
```
Daniel-Fan, line 92: Is it possible that we have more than one CommonService CRs which configure the `bedrockshim` enabled?

In this case, if we delete one of the CommonService CR, and the Crossplane and Cloud operator will be deleted anyway even there are still CommonService CR which has `bedrockshim` enabled remaining in the cluster.

Daniel-Fan: We may follow the same strategy in `r.handleDelete()` and `r.getMinimalSizes()` to decide whether we should delete those operators.
https://github.com/IBM/ibm-common-service-operator/blob/master/controllers/operandconfig.go#L363
https://github.com/IBM/ibm-common-service-operator/blob/master/controllers/operandconfig.go#L317

controllers/bootstrap/init.go
ZhuoxiLi: `Configuration` & `Lock`? they are created after the subscription: https://github.com/IBM/ibm-common-service-operator/blame/f550579f88dd93ee3e5ac3b13661222c1d3c7f3a/controllers/bootstrap/init.go#L621
do we need to remove all its CRs?

Daniel-Fan: @horis233 could you mind confirming it? Should we uninstall `Configuration` & `Lock`?

I thought we should remove all the stuff installed by ourselves during the uninstallation.

And yes, because they are created after subscription(after the operator is installed), we should delete them before removing the operator.

controllers/commonservice_controller.go
ZhuoxiLi: yea.. I was thinking about this issue, which design they are actually looking for? or they will only have `bedrockshim` in one ns, for example IAF ns.

controllers/bootstrap/init.go
horis233: Yes, We should delete them before removing the operator, Otherwise crossplane operands will still be there.

controllers/commonservice_controller.go
horis233: We should consider the multiple adopters scenario. Because IAF will be deployed into each cloudpak namespace. We need to make sure there is no CommonService CR within the scope that enables the crossplane before uninstalling them.

controllers/bootstrap/init.go
```diff
@@ -172,21 +173,70 @@ func (b *Bootstrap) CrossplaneCloudOperator(instance *apiv3.CommonService) error
...
+		return err
 	}
 
+	klog.Infof("Trying to delete ibm-crossplane-operator CR in %s", namespace)
+	resourceCrossConfiguration := constant.CrossConfiguration
+	if err := b.DeleteFromYaml(resourceCrossConfiguration, b.CSData); err != nil {
+		return err
+	}
+	resourceCrossLock := constant.CrossLock
+	if err := b.DeleteFromYaml(resourceCrossLock, b.CSData); err != nil {
```
horis233, line 229: We need to delete CR before we delete operator

controllers/bootstrap/init.go
```diff
@@ -466,6 +516,43 @@ func (b *Bootstrap) CreateOrUpdateFromYaml(yamlContent []byte, alwaysUpdate ...b
...
+			continue
+		} else if err != nil {
+			errMsg = err
+			continue
+		}
+
+		klog.Infof("Deleting resource with name: %s, namespace: %s, kind: %s, apiversion: %s/%s\n", obj.GetName(), obj.GetNamespace(), gvk.Kind, gvk.Group, gvk.Version)
+		if err := b.DeleteObject(obj); err != nil {
+			errMsg = err
+		}
```
horis233, line 550: Could add a wait function to make sure the object is deleted?

ZhuoxiLi: sure, should I create a wait function to check whether the object is deleted? something similar to [WaitResourceReady](https://github.com/IBM/ibm-common-service-operator/blob/c870a4fd53f75a36f910d0a077ad68251771167f/controllers/bootstrap/init.go#L584:21)

ZhuoxiLi: maybe we don't need wait function since no following operations need these deletions has to be done

horis233: We need to make sure CRs are deleted before deleting the operator.

controllers/bootstrap/init.go
```diff
@@ -544,10 +545,30 @@ func (b *Bootstrap) DeleteFromYaml(objectTemplate string, data interface{}) erro
 			continue
 		}
 
-		klog.Infof("Deleting resource with name: %s, namespace: %s, kind: %s, apiversion: %s/%s\n", obj.GetName(), obj.GetNamespace(), gvk.Kind, gvk.Group, gvk.Version)
+		resourceVersion := objInCluster.GetResourceVersion()
+		obj.SetResourceVersion(resourceVersion)
+		obj.SetFinalizers(make([]string, 0))
+		if err := b.UpdateObject(obj); err != nil {
```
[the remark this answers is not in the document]
Daniel-Fan: @ZhuoxiLi Could we update the logic here to not patch the finalizer? and wait for the PR in crossplane to merge.
https://github.com/IBM/ibm-crossplane/pull/40

ZhuoxiLi:
@horis233 @Daniel-Fan this pr is ready to review, thanks!

controllers/bootstrap/init.go
```diff
@@ -172,21 +173,71 @@ func (b *Bootstrap) CrossplaneCloudOperator(instance *apiv3.CommonService) error
...
+
+	return nil
+}
+
+// DeleteCrossplaneCloudSubscription deleted crossplane & cloud operator subscription when bedrockshim set to false or CS CR is removed
+func (b *Bootstrap) DeleteCrossplaneCloudSubscription(namespace, NamespacedName string) error {
+	if b.CSData.CrossplaneEnabledCR[NamespacedName] {
+		delete(b.CSData.CrossplaneEnabledCR, NamespacedName)
+	} else {
+		return nil
```
Daniel-Fan, line 207: I am wondering that if we remove `bedrockshim ` from CR and restart the cs-operator at the same time, then because of the `CrossplaneEnabledCR` is empty, it seems that we could never delete the Crossplane.

But if we remove the `else, return nil`. I am worried that if we have two CRs which one has `bedrockshim` and the other does not have `bedrockshim`.
After restarting cs-operator, if the CR without `bedrockshim` is reconciled first, then the crossplane will be removed and re-installed when reconciling the other CR.

ZhuoxiLi: I think the key to solving this problem is how to get all the `bedrockshim` values from all cs cr after restart.  There are two possible ways come. 
1. could we save the `b.CSData.CrossplaneEnabledCR` in a permanent place? such as write to file or some permanent memory
2. after the cs-operator restarted, it will re-generate all cs cr. Do we know how many cs cr will be re-generated OR what cs cr will be re-generated? (I think we know since the new generated cs cr will have the same data as before). could we fetch each `bedrockshim` value of all cs cr when we re-generate them?

Daniel-Fan: > Could we save the b.CSData.CrossplaneEnabledCR in a permanent place? such as write to file or some permanent memory

Technically, we could store the information into the ConfigMap, and read it when we need the data.

> after the cs-operator restarted, it will re-generate all cs cr. Do we know how many cs cr will be re-generated OR what cs cr will be re-generated? (I think we know since the new generated cs cr will have the same data as before). could we fetch each bedrockshim value of all cs cr when we re-generate them?

A possible solution is that we go through all the CS CRs when those resources are determined by all the CS CRs in the cluster.
In this case, we don't need to worry about whether the operator will be restarted, because no matter which CR is reconciled first, we always fetch every `bedrockshim` values to determine the uninstallation.

We have a example here when we merge the size configuration across the CRs to OperandConfig.

When there is one CS CR deleted in the cluster, we have the function called `handleDelete()`
https://github.com/IBM/ibm-common-service-operator/blob/master/controllers/operandconfig.go#L364
After the deletion, the size configuration in that CR should not apply to OperandConfig anymore.

Therefore, we fetch all the CRs which remain in the cluster, then get a new set of configuration from remaining CR and apply it to the OperandConfig.
https://github.com/IBM/ibm-common-service-operator/blob/36da1ebbc69d8450629463b7ffecee1789ee1dda/controllers/operandconfig.go#L318

I think we could apply to same logic to the uninstallation of crossplane.

When a CR is deleted or its `bedrockshim` value is removed, we fetch remaining CRs and determined whether the crossplane should keep in the cluster or not.

PR #610, opened 2021-08-22
set priority for catalogsource

horis233:
Set OCP default catalogsource have lower priority than other catalogsource


Daniel-Fan (approved):
/lgtm

PR #613, opened 2021-08-24
update permission

horis233:
Revert #611 

It doesn't work on OCP4.6. We need more tests on it


Daniel-Fan (approved):
/lgtm

bundle/manifests/ibm-common-service-operator.clusterserviceversion.yaml
```diff
@@ -249,7 +243,7 @@ spec:
                 - create
                 - get
             - apiGroups:
-                - pkg.crossplane.io
+                - pkg.ibm.crossplane.io
```
Daniel-Fan, line 246: We should not revert this~

PR #622, opened 2021-09-01
UDS : fixing the operator name and packageName

Shrivastava-Varsha:
adding this fix to avoid conflict with RedHat catalog packageName.


horis233 (approved):
/lgtm

PR #624, opened 2021-09-01
require the operator-sdk v1 and yq version 4

Daniel-Fan:
If you have operator-sdk v1 or yq version 3 installed, it will still download newer version of tools to make it running.


horis233 (approved):
/lgtm

PR #627, opened 2021-09-09
add comparison for slices

Daniel-Fan:
When there are slices defined in OperandConfig, we will simply compare those values to check whether they are equal, but we will **not apply any changes insides the slices** for now.

This is for api-catalog boarding task

### DIFF
--- a/controllers/rules/resource_comparison.go
+++ b/controllers/rules/resource_comparison.go
@@ -93,8 +93,20 @@ func ResourceEqualComparison(resourceA interface{}, resourceB interface{}) bool 
 
 	isEqual := true
 	switch resourceA := resourceA.(type) {
-	// TODO: consider the slices
-	// case []interface{}:
+	case []interface{}:
+		if resourceB == nil {
+			isEqual = false
+		} else if resourceB, ok := resourceB.([]interface{}); ok {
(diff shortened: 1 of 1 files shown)


controllers/rules/resource_comparison.go
```diff
@@ -93,8 +93,20 @@ func ResourceEqualComparison(resourceA interface{}, resourceB interface{}) bool
...
+	case []interface{}:
+		if resourceB == nil {
+			isEqual = false
+		} else if resourceB, ok := resourceB.([]interface{}); ok {
+			if len(resourceA) != len(resourceB) {
+				isEqual = false
+			} else {
+				// TODO: need to find a better way to compare when the order of slice is not fixed
+				for index := range resourceA {
+					isEqual = isEqual && ResourceEqualComparison(resourceA[index], resourceB[index])
```
horis233, line 105:
```suggestion
					if !ResourceEqualComparison(resourceA[index], resourceB[index]) {
					   return false
				         }
```

Then we can cut some unnecessary branches in the loop check

Daniel-Fan: Great Idea, Updated!

horis233 (approved):
/lgtm

PR #633, opened 2021-09-14
get operator list from OperandRegistry

ZhuoxiLi:


controllers/goroutines/operator_status.go
```diff
@@ -40,23 +40,16 @@ func UpdateCsCrStatus(bs *bootstrap.Bootstrap) {
...
-			"ibm-ingress-nginx-operator",
-			"ibm-licensing-operator",
-			"ibm-management-ingress-operator",
-			"ibm-mongodb-operator",
-			"ibm-monitoring-grafana-operator",
 			"ibm-namespace-scope-operator",
-			"ibm-platform-api-operator",
-			"ibm-zen-operator",
-			"operand-deployment-lifecycle-manager-app"}
+			"operand-deployment-lifecycle-manager-app"}...)
```
Daniel-Fan, line 52: Would we include the `ibm-cloud-operator` as well?

ZhuoxiLi: actually, I am wondering will we add crossplane operator and cloud operator to operandregistry in the futurn? maybe @horis233

horis233: No, they won't be added into operandregistry. We can add both of them to this list.

ZhuoxiLi: thank you all! added to the list

horis233 (approved):
/lgtm

PR #635, opened 2021-09-16
display cs cr status in ocp console

ZhuoxiLi:


bundle/manifests/ibm-common-service-operator.clusterserviceversion.yaml
```diff
@@ -50,11 +53,22 @@ spec:
           - description: The size of deployment profile for IBM Cloud Pak foundational services
             displayName: Size
             path: size
-            x-descriptors:
```
Daniel-Fan, line 53: Wondering if we remove the `x-descriptors` for `size`. Would user be able to select one of sizes in creation view instead of input the text?

bundle/manifests/ibm-common-service-operator.clusterserviceversion.yaml
```diff
@@ -17,6 +17,9 @@ metadata:
           },
           "spec": {
             "size": "starterset"
+          },
```
Daniel-Fan, line 20: May I ask that is this added by operator-sdk automatically?

ZhuoxiLi: I added it, just tested and found it could be removed

api/v3/commonservice_types.go
```diff
@@ -64,7 +64,7 @@ type Bedrockshim struct {
 type BedrockOperator struct {
 	Name               string `json:"name,omitempty"`
 	Version            string `json:"version,omitempty"`
-	Status             string `json:"status,omitempty"`
+	OperatorStatus     string `json:"operatorStatus,omitempty"`
```
horis233, line 67: Will this Change in this field cause an upgrade issue?

bundle/manifests/ibm-common-service-operator.clusterserviceversion.yaml
horis233: We probably need to remove it. It's a little bit strange to include a status when creating CR

api/v3/commonservice_types.go
ZhuoxiLi: it should be fine. I tested the upgrade path 3.11 -> my dev build:
1. fresh installed 3.11 from quay.io
2. replace installed catalogsource with my catalogsource includes this change
3. run `oc apply -f bundle/manifests/operator.ibm.com_commonservices.yaml` on testing cluster

Daniel-Fan: It looks good from the code side. The new operator slice will totally overwrite the previously status. So it would be fine during the upgrade.

Daniel-Fan (approved):
/lgtm

PR #637, opened 2021-09-20
update certmanager APIGroup version to v1

Daniel-Fan:


horis233 (commented):
I believe we can handle this PR so far until other services adopter v1 cert-manager

Daniel-Fan:
Good idea, I will add the `hold` label for that!

bitscuit:
> 
> 
> I believe we can handle this PR so far until other services adopting v1 cert-manager

Technically it should not matter what apiVersion is being used because only v1 Issuers and Certificates are resolved. This might actually be a good test to see if there are any issues that come from mixing different apiVersions because at some point, we want all services to be creating resources with `cert-manager.io/v1`. If we can mix them, then that means we do not need to coordinate for all services to switch at the same time, reducing our work.

horis233:
@Daniel-Fan @bitscuit 

If we want to do this update in this release, we should do it soon than later.
Also please test the upgrade case. My concern is that certificate v1 generated by common service will overwrite the certificate v1 converted by the cert-manager operator.

Daniel-Fan:
I will remove the `hold` label.

For the upgrade scenario,

If the cert-manager get upgraded first, and yes, the cs-operator will update the v1 certificate which is converted by cert-manager-operator. Because the annotation version is the larger than the previous one. In this case, could cert-manager controller handle it correctly?

If the cs-operator get upgraded first, then the new v1 certificate and issuer will be generated, and I am wondering how would cert-manager-operator handle this situation? It would not convert the old certificate because the v1 version is already there, right?

bitscuit:
> 
> 
> I will remove the `hold` label.
> 
> For the upgrade scenario,
> 
>     1. If the cert-manager get upgraded first, and yes, the cs-operator will update the v1 certificate which is converted by cert-manager-operator. Because the annotation version is the larger than the previous converted one.
>        In this case, could cert-manager controller handle it correctly?
> 
>     2. If the cs-operator get upgraded first, then the new v1 certificate and issuer will be generated, and I am wondering how would cert-manager-operator handle this situation?
>        It would not convert the old certificate because the v1 version is already there, right?

1. A v1 Certificate being updated is just part of the normal workflow for cert-manager-controller. If there are no spec changes, then the certificate secret should remain the same.

2. Right, no conversion will happen https://github.com/IBM/ibm-cert-manager-operator/blob/master/pkg/controller/certificate/certificate_controller.go#L164-L165. 

The cs-operator should delete the v1alpha1 Certificate and Issuer if it's going to create the v1 versions. It's not necessary, but good to cleanup unused resources. However, I am concerned that deleting v1alpha1 Certificate will also delete the v1 version in both scenarios. We might have to change the deletion logic in cert-manager-operator.

horis233:
@bitscuit 

I believe it is a channel we can think of how to guide the best practice when they switching to v1 cert-manager.

Daniel-Fan:
@bitscuit @horis233 

I think it won't be necessary to clean up the v1alpha1 Certificate and Issuer by cs-operator.
* If we only delete cs-ca-certificate and something else that are created by cs-operator, then  there are still lots of other v1alpha1 resources that will remain in the cluster.
* if we remove all the v1alpha1 certificate and issuer in the cluster by cs-operator, it is too risky that most of them are not created by cs-operator.

> A v1 Certificate being updated is just part of the normal workflow for cert-manager-controller. If there are no spec changes, then the certificate secret should remain the same.

the only concern that I am worried is when cs-operator update(apply the template to) the v1 certificate, it will update the renew time and duration. Other fields may remain still if it is not updated by cert-manager-controller before.
```
spec:
  secretName: cs-ca-certificate-secret
  issuerRef:
    name: cs-ss-issuer
    kind: Issuer
  commonName: cs-ca-certificate
  isCA: true
  duration: 17520h0m0s
  renewBefore: 720h0m0s
```

horis233:
> the only concern that I am worried is when cs-operator update(apply the template to) the v1 certificate, it will update the renew time and duration. Other fields may remain still if it is not updated by cert-manager-controller before.

@Daniel-Fan Could you do a quick verification for your concern?

Daniel-Fan:
Oh, I think I misunderstand it.

I thought the `duration` and `renewBefore` are the specific time points when the certificates will be renewed.
But actually, it is time periods that will not be changed, which means when cs-operator updates the certificate, its `specs` section will not change.

I don't have other concerns. Do we wanna merge it now?

horis233 (approved):
/lgtm

Daniel-Fan:
@horis233 
May I ask how to restart the prow job from beginning? We try to delete the pods in test namespace, but it does not work here.

Daniel-Fan:
Solve the issue by ssh into the s390 cluster and delete all the pods to trigger the restart.

PR #647, opened 2021-09-28
Revert "update certmanager APIGroup version to v1 (#637)" so that CICD

bitscuit:
can produce latest-validated build earlier. Will add back in once
cert-manager can handle mixing v1 and v1alpha1 certs

This reverts commit b032cf32b85af87a1d9472778b7ce640b5ca0f91.


horis233 (approved):
/lgtm

PR #648, opened 2021-09-29
add EDB template & job to create the license key

ZhuoxiLi:


controllers/constant/odlm.go
```diff
@@ -104,6 +104,96 @@ spec:
...
+                hostNetwork: false
+                hostPID: false
+                restartPolicy: Never
+                securityContext:
+                  runAsNonRoot: true
+                serviceAccountName: edb-license-sa
+      - apiVersion: v1
+        kind: ServiceAccount
+        metadata:
+          name: edb-license-sa
```
Daniel-Fan, line 172: We could use `name: edb-license-sa` directly, no need for `metadata`

controllers/constant/odlm.go
```diff
@@ -104,6 +104,96 @@ spec:
...
+        kind: RoleBinding
+        metadata:
+          name: edb-license-rolebinding
+        subjects:
+        - kind: ServiceAccount
+          name: edb-license-sa
+        roleRef:
+          kind: Role
+          name: edb-license-role
+          apiGroup: rbac.authorization.k8s.io
```
Daniel-Fan, line 183: The same here for `name: edb-license-rolebinding`, other fields should be included in `data` section

controllers/constant/odlm.go
```diff
@@ -104,6 +104,96 @@ spec:
...
+      - apiVersion: rbac.authorization.k8s.io/v1
+        kind: Role
+        metadata:
+          name: edb-license-role
+        rules:
```
Daniel-Fan, line 188: The same here

controllers/constant/odlm.go
```diff
@@ -356,6 +446,96 @@ spec:
     spec:
       grafana: {}
       operandRequest: {}
+  - name: cloud-native-postgresql
```
Daniel-Fan, line 449: I don't think we need to cover the SaaS scenario for EDB.
@horis233 Could you please confirm this?

controllers/constant/odlm.go
```diff
@@ -431,6 +611,12 @@ spec:
     installPlanApproval: {{ .ApprovalMode }}
     sourceName: {{ .CatalogSourceName }}
     sourceNamespace: {{ .CatalogSourceNs }}
+  - channel: stable
```
Daniel-Fan, line 614: Could we move this part into `const CSV3OperandRegistry`?

ZhuoxiLi: `cloud-native-postgresql` already added in `const CSV3OperandRegistry`. I think EDB will support multi ns in the future, but if we prefer not to include them in saas at this time, I will remove it.

Daniel-Fan: Ah I see, sorry that I did not notice it.

For `CSV3SaaSOperandRegistry`, it really depends on whether EDB will be SaaS scenario.

For `CSV3OperandRegistry`, it already supports both on-prem single and multi instances installation, so the multi ns for EDB will be covered in this case.

Let's confirm with @horis233 about the SaaS

controllers/constant/odlm.go
horis233: No, we don't need EDB in the SaaS mode

controllers/constant/odlm.go
horis233: We don't need to add it in SaaS until we get any further notification.

controllers/constant/odlm.go
```diff
@@ -104,6 +104,92 @@ spec:
       apicatalogmanager: {}
       operandBindInfo: {}
       operandRequest: {}
+  - name: cloud-native-postgresql
+    spec:
+      operandBindInfo: {}
```
horis233, line 109: We don't need `operandBindInfo: {}` or  `operandRequest: {}`

controllers/constant/odlm.go
```diff
@@ -104,6 +104,89 @@ spec:
...
+                    cat << EOF | kubectl apply -f -
+                    apiVersion: v1
+                    kind: Secret
+                    type: Opaque
+                    metadata:
+                      name: postgresql-operator-controller-manager-config
+                    data:
+                      EDB_LICENSE_KEY: $(base64 /license_keys/edb/EDB_LICENSE_KEY | tr -d '\n')
+                    EOF
+                  image: cp.icr.io/cp/cpd/wd-postgres-license@sha256:4d8d0ecd31d04e15f757ffda74101ef5bd7fb5789db1ffcef892e961ef312ebf
```
Daniel-Fan, line 142: @horis233 Do we need to create pull secret and link it to new SA?

horis233: Yes we need to set up imagepullsecret `ibm-entitlement-key` in the job

controllers/constant/odlm.go
```diff
@@ -104,6 +104,93 @@ spec:
...
+        data:
+          rules:
+          - apiGroups:
+            - ""
+            resources:
+            - secret
+            verbs:
+            - create
+            - update
+            - patch
```
Daniel-Fan, line 193: Could we change the position of `Role` and `Rolebinding`?

We wanna create the role first and then we are able to create the rolebinding.

Daniel-Fan (approved):
/lgtm

PR #651, opened 2021-09-30
skip edb checking for size

ZhuoxiLi:


controllers/operandconfig.go
```diff
@@ -276,6 +276,10 @@ func (r *CommonServiceReconciler) updateOperandConfig(newConfigs []interface{})
 	isEqual := true
 	for _, opService := range opconServices {
 		existingOpService := getItemByName(existingOpconServices.([]interface{}), opService.(map[string]interface{})["name"].(string))
+        if opService.(map[string]interface{})["name"].(string) == "cloud-native-postgresql" {
+			// skip sizing compare for cloud-native-postgresql
+            continue
+        }
```
Daniel-Fan, line 282: Wondering whether it is aligned with the above line, maybe it is a Github UI error.

Also may I ask could we try not to hardcode the skip logic?
Is it possible that we could check that `opService.(map[string]interface{})["spec"]` does not exist or `nil`, then skip it?

ZhuoxiLi: for the aligned issue, that's weird... since it looks good on the vscode, will check why that happened

for the second one, I think the general check should work. It was just a quick fix, I will update it.

PR #656, opened 2021-10-01
set restartPolicy as OnFailure

ZhuoxiLi:
set the job `restartPolicy` as `OnFailure`, so if the job will retry when failed


horis233 (approved):
/lgtm

PR #671, opened 2021-10-18
Update crossplane's Lock API version

PawelKopel:
Update Crossplane's Lock API version from v1alpha1 to v1beta1.
Remove code creating Lock during installation - Crossplane will create Lock on its own when necessary.


horis233:
@ZhuoxiLi 

Since the `lock` creation logic is removed, let's remove the `lock` delete logic and rbac permissions for `lock` as well.

PR #687, opened 2021-10-28
get nss scope from odlm nss

horis233:


horis233:
@Daniel-Fan 

Thanks for the notification. I will update it to get the common namespaces from two NSS CRs.

PR #697, opened 2021-11-05
check dev image for individual operator package

horis233:
It is for solving the following problem.

If a developer is using opencloudio with dev image (artifactory) for common services, then other bedrock operators out of opencloudio, like postgresal, can't be installed from the ibm-operator-catalog caatalogsource.  

This PR moves the dev image checking logic into catalogsource comparison, then we can analyze the case individually without impact crossing different catalogsources


horis233:
@Daniel-Fan @ZhuoxiLi Could you help review this PR?

Daniel-Fan (approved):
/lgtm

PR #698, opened 2021-11-06
increase cert manager resources for EUS

Daniel-Fan:
@horis233 

May I ask that

Should we increase the resources for **all the three profiles** to make all of them exactly same?

Should we only increase the CPU limit of the `cert-manager CA injector`? Or other resources should be changed as well.

### DIFF
--- a/controllers/size/large_amd64.go
+++ b/controllers/size/large_amd64.go
@@ -47,10 +47,10 @@ const Large = `
       configMapWatcher:
         resources:
           limits:
-            cpu: 60m
+            cpu: 20m
             memory: 60Mi
           requests:
-            cpu: 30m
+            cpu: 20m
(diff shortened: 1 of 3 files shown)


horis233:
@Daniel-Fan 
Let's only update the resource of the cert-manager operands because the resource problems of the cert-manager are frequently reported by customers. We can keep others as before since they are working well.

Daniel-Fan:
@horis233 
Sure, I remove the changes about configwater.

Also, currently the resource values for cert-manager operands are the same in all three profiles.

Daniel-Fan:
@horis233 

The PR is ready for reviewing~

horis233 (approved):
/lgtm

horis233:
Thanks for your reminder

PR #702, opened 2021-11-12
update compare resource string version logic

ZhuoxiLi:
Let review the logic first, will need test on Fyre before merge


ZhuoxiLi:
removed testing msg, please review again @horis233 thanks!

Daniel-Fan (approved):
/lgtm

PR #703, opened 2021-11-15
add rollback script from CD to EUS

Daniel-Fan:
The current rollback procedure cannot fully pass the SERT regression test.

I have not got the chances to run the second round of test due to the maintenance of Fyre. But we could review the PR first.


Daniel-Fan:
@horis233 @ZhuoxiLi 
The script is ready for merging.

It has been added into `scripts` branch~

horis233 (approved):
/lgtm

PR #729, opened 2022-01-18
loop checking clustertype instead exit 1

ZhuoxiLi:


main.go
```diff
@@ -122,16 +122,19 @@ func main() {
...
-	typeCorrect, err := bootstrap.CheckClusterType(mgr, util.GetMasterNs(mgr.GetAPIReader()))
-	if err != nil {
-		klog.Errorf("Failed to verify cluster type  %v", err)
-		os.Exit(1)
-	}
+	for {
+		typeCorrect, err := bootstrap.CheckClusterType(mgr, util.GetMasterNs(mgr.GetAPIReader()))
+		if err != nil {
+			klog.Errorf("Failed to verify cluster type  %v", err)
+			os.Exit(1)
```
horis233, line 129: @ZhuoxiLi Let's continue the loop if we get the error as well. It will help us reduce the pod crash.

horis233 (approved):
/lgtm

PR #734, opened 2022-01-26
Update Copyright to 2022

Daniel-Fan:


controllers/operandconfig.go
```diff
@@ -0,0 +1,469 @@
+//
```
horis233, line 1: Why this new file is created?

Daniel-Fan: Yeah, I think it is introduced when I rebased some commits.
Updated~

horis233 (approved):
/lgtm

PR #751, opened 2022-02-18
create secretshare CR to share namespace-scope config to openshift-re…

horis233:
…dhat-marketplace namespace


Daniel-Fan (approved):
/lgtm

PR #752, opened 2022-02-21
add turbo profile controller for common service

Daniel-Fan:
#### New Paratemers
1. Add `profileController` as the overall management strategy
2. Add `managementStrategy` as the management controller for individual service

#### Options for new parameters
Options for `profileController` and `managementStrategy` are `default`, `commonservice`, `turbo`, `turbonomic` and `vpa`.
1. `default` and `commonservice` options will keep the original logic where the `OperandConfig` is totally managed by common service CR.
2. `turbo` is the short name of  `turbonomic`.
3. `turbo`, `turbonomic` and `vpa` share the same logic where the resource section including `limit` and `request` will be cleaned up in `OperandConfig`. But it will keep other configuration unchanged, for example, `storageclass` and `routehost`.

#### Relationship between paramater
1. `profileController` will be co-existed with `size` parameter, but `profileController` could overwrite `size` setting.
For example, if `size = large` and `profileController = turbo`, then resource section in `OperandConfig` will still be cleaned up.
2. `managementStrategy` has the higher priority then `profileController` for individual services. For example, `cert-manager-operator` has `managementStrategy = default(or commonservice)` and `profileController = turbo`. then all other services is managed by `turbonomic` which `OpandConfig` is empty for those corresponding services, and `cert-manager-operator` is still managed by common service, which means size profile or `spec` section in CS CR for `cert-manager-opperator` is applied to `OperandConfig`.
3. When `size = as-is`, we will not apply any size template and keep `OperandConfig` as usual. The `OperandConfig` will be updated if we define individual service configuration(includes either specific value or `managementStrategy`) in CS CR(it is the same as before), also the `OperandConfig` will be emptied if `profileController = turbo`

Common Service CR example that `profileController = turbo`, `size = medium` and `managementStrategy = commonservice` for `ibm-mongodb-operator`:
```
kind: CommonService
metadata:
  labels:
    app.kubernetes.io/instance: ibm-common-service-operator
    app.kubernetes.io/managed-by: ibm-common-service-operator
    app.kubernetes.io/name: ibm-common-service-operator
  name: example-commonservice
  namespace: ibm-common-services
spec:
  profileController: turbo
  services:
    - managementStrategy: commonservice
      name: ibm-mongodb-operator
      spec:
        mongoDB:
          metrics:
            resources:
              limits:
                cpu: 1000m
                memory: 350Mi
              requests:
                cpu: 100m
                memory: 300Mi
          replicas: 3
          resources:
            limits:
              cpu: 3000m
              memory: 3Gi
            requests:
              cpu: 500m
              memory: 3Gi
  size: medium
status:
  phase: Succeeded
```

OperandConfig Example where `ibm-mongodb-operator` keeps `medium` profile and other services is emptied and managed by `turbonomic`:
```
apiVersion: operator.ibm.com/v1alpha1
kind: OperandConfig
  name: common-service
  namespace: ibm-common-services
spec:
  services:
    - name: ibm-licensing-operator
      spec:
        IBMLicensing:
          datasource: datacollector
          resources:
            limits: {}
            requests: {}
        operandBindInfo: {}
    - name: ibm-mongodb-operator
      spec:
        mongoDB:
          replicas: 3
          resources:
            limits:
              cpu: 3000m
              memory: 3Gi
            requests:
              cpu: 500m
              memory: 3Gi
        operandRequest: {}
```

### DIFF
--- a/config/manifests/bases/ibm-common-service-operator.clusterserviceversion.yaml
+++ b/config/manifests/bases/ibm-common-service-operator.clusterserviceversion.yaml
@@ -44,6 +44,14 @@ spec:
         - urn:alm:descriptor:com.tectonic.ui:select:medium
         - urn:alm:descriptor:com.tectonic.ui:select:large
         - urn:alm:descriptor:com.tectonic.ui:select:production
+      - description: The profile controller for IBM Cloud Pak foundational services
+        displayName: ProfileController
+        path: profileController
+        x-descriptors:
+        - urn:alm:descriptor:com.tectonic.ui:select:default
+        - urn:alm:descriptor:com.tectonic.ui:select:commonservice
+        - urn:alm:descriptor:com.tectonic.ui:select:turbonomic
+        - urn:alm:descriptor:com.tectonic.ui:select:vpa
       statusDescriptors:
       - description: Installed Bedrock Operator Name
         displayName: Name
--- a/controllers/constant/odlm.go
+++ b/controllers/constant/odlm.go
@@ -30,7 +30,6 @@ spec:
     spec:
       IBMLicensing:
         datasource: datacollector
-      IBMLicenseServiceReporter: {}
       operandBindInfo: {}
   - name: ibm-mongodb-operator
     spec:
@@ -460,7 +459,6 @@ spec:
         datasource: datacollector
         routeEnabled: false
         logLevel: VERBOSE
-      IBMLicenseServiceReporter: {}
       operandBindInfo: {}
   - name: ibm-mongodb-operator
     spec:
--- a/controllers/constant/odlm.go
+++ b/controllers/constant/odlm.go
@@ -30,6 +30,7 @@ spec:
     spec:
       IBMLicensing:
         datasource: datacollector
+      IBMLicenseServiceReporter: {}
       operandBindInfo: {}
   - name: ibm-mongodb-operator
     spec:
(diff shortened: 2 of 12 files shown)


config/manifests/bases/ibm-common-service-operator.clusterserviceversion.yaml
```diff
@@ -44,6 +44,14 @@ spec:
         - urn:alm:descriptor:com.tectonic.ui:select:medium
         - urn:alm:descriptor:com.tectonic.ui:select:large
         - urn:alm:descriptor:com.tectonic.ui:select:production
+      - description: The profile controller for IBM Cloud Pak foundational services
```
pgodowski, line 47: It is an optional parametry for backwardancompatibility, right?

pgodowski (commented):
See my comment about optional field clarification

config/manifests/bases/ibm-common-service-operator.clusterserviceversion.yaml
Daniel-Fan: Yes, `profileControleller` is optional. It could be empty when creating Common Service CR, and will not cause issue during the upgrade.

controllers/constant/odlm.go
```diff
@@ -30,6 +30,7 @@ spec:
     spec:
       IBMLicensing:
         datasource: datacollector
+      IBMLicenseServiceReporter: {}
```
horis233, line 33: Why we add them into the OperandConfig?

Daniel-Fan: I found it in size template, but now notice that we don't actually create this CR in the installation.

Will remove it.

horis233 (approved):
/lgtm

PR #801, opened 2022-06-27
update installplan

qpdpQ:
for this issue https://github.ibm.com/IBMPrivateCloud/roadmap/issues/53818

### DIFF
--- a/controllers/bootstrap/init.go
+++ b/controllers/bootstrap/init.go
@@ -1376,17 +1376,17 @@ func (b *Bootstrap) updateApprovalMode() error {
 		return err
 	}
 
-	if err = b.UpdateOpApproval("ibm-common-service-operator"); err != nil {
+	if err = b.UpdateOpApproval(constant.IBMCSPackage); err != nil {
 		klog.Errorf("Failed to update common service operator subscription: %v", err)
 		return err
 	}
 
-	if err = b.UpdateOpApproval("operand-deployment-lifecycle-manager-app"); err != nil {
+	if err = b.UpdateOpApproval(constant.IBMODLMSPackage); err != nil {
 		klog.Errorf("Failed to update operand-deployment-lifecycle-manager-app subscription: %v", err)
 		return err
 	}
 
-	if err = b.UpdateOpApproval("ibm-namespace-scope-operator"); err != nil {
+	if err = b.UpdateOpApproval(constant.IBMNSSPackage); err != nil {
 		klog.Errorf("Failed to update namespace-scope-operator subscription: %v", err)
 		return err
 	}
@@ -1396,7 +1396,7 @@ func (b *Bootstrap) updateApprovalMode() error {
 			klog.Errorf("Failed to update %s subscription: %v", constant.ICPOperator, err)
 			return err
 		}
-		klog.Infof("%s not installed, skipping", constant.ICPOperator)
+		klog.Infof("%s not installed, skipping updating approval strategy", constant.ICPOperator)
 
 	}
 
@@ -1405,7 +1405,7 @@ func (b *Bootstrap) updateApprovalMode() error {
 			klog.Errorf("Failed to update %s subscription: %v", constant.ICPPICOperator, err)
 			return err
 		}
-		klog.Infof("%s not installed, skipping", constant.ICPPICOperator)
+		klog.V(2).Infof("%s not installed, skipping", constant.ICPPICOperator)
 
 	}
 
@@ -1414,7 +1414,7 @@ func (b *Bootstrap) updateApprovalMode() error {
 			klog.Errorf("Failed to update %s subscription: %v", constant.ICPPKOperator, err)
 			return err
 		}
-		klog.Infof("%s not installed, skipping", constant.ICPPKOperator)
+		klog.V(2).Infof("%s not installed, skipping", constant.ICPPKOperator)
 
 	}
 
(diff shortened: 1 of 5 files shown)


controllers/bootstrap/init.go
```diff
@@ -1374,11 +1376,48 @@ func (b *Bootstrap) updateApprovalMode() error {
...
+		klog.Infof("%s not installed, skipping", constant.ICPPICOperator)
+
+	}
+
+	if err = b.UpdateOpApproval(constant.ICPPKOperator); err != nil {
+		if !errors.IsNotFound(err) {
+			klog.Errorf("Failed to update %s subscription: %v", constant.ICPPKOperator, err)
+			return err
+		}
+		klog.Infof("%s not installed, skipping", constant.ICPPKOperator)
```
Daniel-Fan, line 1417: Hi @qpdpQ 

Could you try to use lower verbose level for the information log?
I think it would be better to reduce some logs for debug convenience.

Here is an example of lower verbose level
https://github.com/IBM/ibm-common-service-operator/blob/master/controllers/bootstrap/init.go#L1394
```
klog.V(2).Infof("waiting for resource ready with kind: %s, apiGroupVersion: %s", kind, apiGroupVersion)
```

controllers/bootstrap/init.go
```diff
@@ -1374,11 +1376,48 @@ func (b *Bootstrap) updateApprovalMode() error {
...
+		klog.Errorf("Failed to update namespace-scope-operator subscription: %v", err)
+		return err
+	}
+
+	if err = b.UpdateOpApproval(constant.ICPOperator); err != nil {
+		if !errors.IsNotFound(err) {
+			klog.Errorf("Failed to update %s subscription: %v", constant.ICPOperator, err)
+			return err
+		}
+		klog.Infof("%s not installed, skipping", constant.ICPOperator)
```
Daniel-Fan, line 1399: Also may I ask that if we could be more specific about the information?
For example, we could say `skipping updating approval strategy`.

controllers/bootstrap/init.go
```diff
@@ -1374,11 +1376,48 @@ func (b *Bootstrap) updateApprovalMode() error {
...
 		klog.Errorf("Failed to update common service operator subscription: %v", err)
 		return err
 	}
 
+	if err = b.UpdateOpApproval("operand-deployment-lifecycle-manager-app"); err != nil {
+		klog.Errorf("Failed to update operand-deployment-lifecycle-manager-app subscription: %v", err)
+		return err
+	}
+
+	if err = b.UpdateOpApproval("ibm-namespace-scope-operator"); err != nil {
```
Daniel-Fan, line 1389: I am wondering if we are able to use constant variable instead of hard-coded string for Operator Name.

We could find `ibm-common-service-operator` here
https://github.com/IBM/ibm-common-service-operator/blob/master/controllers/constant/constant.go#L68

Daniel-Fan (commented):
Hi @qpdpQ 

Please check the above comments.

controllers/bootstrap/init.go
qpdpQ: sorry, I just saw it. I will fix them now, thanks a lot for advice! 😄

controllers/bootstrap/init.go
```diff
@@ -1374,11 +1376,48 @@ func (b *Bootstrap) updateApprovalMode() error {
 		return err
 	}
 
-	if err = b.UpdateCsOpApproval(); err != nil {
+	if err = b.UpdateOpApproval(constant.IBMCSPackage); err != nil {
 		klog.Errorf("Failed to update common service operator subscription: %v", err)
```
Daniel-Fan, line 1380: Hi @qpdpQ 

Could we have some hard-coded operator name replaced by variables? e.g. `constant.IBMCSPackage`

main.go
```diff
@@ -214,7 +214,7 @@ func main() {
 			klog.Errorf("Failed to create common service operator subscription: %v", err)
 			os.Exit(1)
 		}
-		if err = bs.UpdateCsOpApproval(); err != nil {
+		if err = bs.UpdateOpApproval("ibm-common-service-operator"); err != nil {
```
Daniel-Fan, line 217: Also, could we have the variable instead of the hard-coded string?

controllers/bootstrap/init.go
```diff
@@ -1374,11 +1376,48 @@ func (b *Bootstrap) updateApprovalMode() error {
...
+	}
+
+	if err = b.UpdateOpApproval(constant.ICPPKOperator); err != nil {
+		if !errors.IsNotFound(err) {
+			klog.Errorf("Failed to update %s subscription: %v", constant.ICPPKOperator, err)
+			return err
+		}
+		klog.V(2).Infof("%s not installed, skipping", constant.ICPPKOperator)
+
+	}
```
Daniel-Fan, line 1419: I think we need to move this part to the place where crossplane is installed.
Crossplane is installed after we initialize the resources. Therefore, if we update subscription of crossplane here, it will usually skip the update because it has not been installed yet.
https://github.com/IBM/ibm-common-service-operator/blob/master/controllers/commonservice_controller.go#L147
https://github.com/IBM/ibm-common-service-operator/blob/master/controllers/commonservice_controller.go#L234

qpdpQ: Yeah, I agree, will change it.

Daniel-Fan (approved):
/lgtm

Great Job! Thank you @qpdpQ

PR #808, opened 2022-07-11
compatibility clause to CS CRDs

qpdpQ:


qpdpQ:
for this ticket: https://github.ibm.com/IBMPrivateCloud/roadmap/issues/54203

bundle/manifests/operator.ibm.com_commonservices.yaml
```diff
@@ -56,6 +56,7 @@ spec:
                         type: boolean
                     type: object
                 type: object
+                x-kubernetes-preserve-unknown-fields: true
```
Daniel-Fan, line 59: Is this clause applying to `spec.features` instead of `spec`?

controllers/constant/webhook.go
```diff
@@ -1633,6 +1633,7 @@ spec:
                   type: object
                 type: array
             type: object
+            x-kubernetes-preserve-unknown-fields: true
```
Daniel-Fan, line 1636: From the template
```
spec:
            description: |+
              * documentation of parameters available, but not validated by OpenAPI / CRD schema
            type: object
            x-kubernetes-preserve-unknown-fields: true
```

I think the clause should be aligned with line 1637 `type: object`

bundle/manifests/operator.ibm.com_commonservices.yaml
qpdpQ: Oh nice catch, I put it in wrong place thanks

Daniel-Fan (approved):
/lgtm

PR #814, opened 2022-07-13
MultiInstance singleton version handling

bluzarraga:
ensure crossplane operator only installs/upgrades if the attempted install version is later than the existing version


bitscuit:
if you're using vs code, install the go extension so that it automatically formats your code when saving

bluzarraga:
this pr is replaced by https://github.com/IBM/ibm-common-service-operator/pull/818/

PR #827, opened 2022-07-20
make createOrUpdateFromYaml generic to include control ns

bluzarraga:

Open to suggestions on handling the namespace is nil error

### DIFF
--- a/controllers/bootstrap/init.go
+++ b/controllers/bootstrap/init.go
@@ -25,7 +25,6 @@ import (
 	"sync"
 	"text/template"
 	"time"
-	"reflect"
 
 	olmv1 "github.com/operator-framework/api/pkg/operators/v1"
 	olmv1alpha1 "github.com/operator-framework/api/pkg/operators/v1alpha1"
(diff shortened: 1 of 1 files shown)


bluzarraga:
changed error handling based on underlying k8s function ([here](https://github.com/kubernetes/apimachinery/blob/v0.24.3/pkg/apis/meta/v1/unstructured/unstructured.go#L233) and [here](https://github.com/kubernetes/apimachinery/blob/97e5df2d0258ac077093867fabf508c1fc31f53b/pkg/apis/meta/v1/unstructured/helpers.go#L277))

controllers/bootstrap/init.go
```diff
@@ -717,9 +717,13 @@ func (b *Bootstrap) CreateOrUpdateFromYaml(yamlContent []byte, alwaysUpdate ...b
...
-			sub, err := b.GetSubscription(ctx, obj.GetName(), b.CSData.MasterNs)
+			sub, err := b.GetSubscription(ctx, obj.GetName(), obj.GetNamespace())
 			if err != nil {
-				klog.Errorf("Failed to get subscription %s/%s", b.CSData.MasterNs, obj.GetName())
+				if obj.GetNamespace() == "" {
+					klog.Errorf("Failed to get subscription for %s. Namespace not found.", obj.GetName())
+				} else {
+					klog.Errorf("Failed to get subscription %s/%s", obj.GetNamespace(), obj.GetName())
+				}
 			}
```
Daniel-Fan, line 727: Hi @bluzarraga 

When the `err != nil` happens, the `sub` would be `nil`.
Therefore, if it continues to run into `update = !equality.Semantic.DeepEqual(sub.Object["spec"], obj.Object["spec"])`, we will hit the `nil` pointer issue for `sub` and crash the pod. Therefore, we should `return err` directly.

Daniel-Fan: Also, I think the additional check for `obj.GetNamespace() == ""`  is good. It will help to provide more information if the error happens.

bluzarraga: Ok I see what you're saying now, the nil pointer is from the deep compare

Daniel-Fan (approved):
/lgtm

PR #830, opened 2022-07-21
Upgrade script for 3.20 release

Daniel-Fan:
Will change the name of script later

```
./upgrade_common_services_draft.sh -h                                                                   
Usage: upgrade_common_services_draft.sh [OPTION]...
Upgrade Common Services

Options:
Mandatory arguments to long options are mandatory for short options too.
      -h, --help                    display this help and exit
      -csNS                         specify the namespace where common service is installed. By default it is namespace ibm-common-services.
      -cloudpaksNS                  specify the namespace where cloud paks is installed. By default it would be same as csNS.
      -controlNS                    specify the namespace where singleton services are installed. By default it it would be same as csNS.
      -c                            specify the subscription channel where common services switch. By default it is channel v3
Note:
If there is no namespace defined through arguments, all Common Service in the cluster will be upgraded.
```

### DIFF
--- a/upgrade_common_services_draft.sh
+++ b/upgrade_common_services_draft.sh
@@ -246,6 +246,10 @@ function switch_channel() {
     success "Updated ibm-common-service-operator subscriptions successfully."
     msg ""
 
+    # scale down ibm-common-service-operator deployment
+    msg "scaling down ibm-common-service-operator deployment in ${csNS} namespace"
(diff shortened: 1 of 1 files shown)


Daniel-Fan:
close this one and replace it by https://github.com/IBM/ibm-common-service-operator/pull/831

PR #831, opened 2022-07-22
script for upgrading common service to 3.20

Daniel-Fan:
```
./upgrade_common_services.sh -h                                                                   
Usage: upgrade_common_services_draft.sh [OPTION]...
Upgrade Common Services

Options:
Mandatory arguments to long options are mandatory for short options too.
      -h, --help                    display this help and exit
      -csNS                         specify the namespace where common service is installed. By default it is namespace ibm-common-services.
      -cloudpaksNS                  specify the namespace where cloud paks is installed. By default it would be same as csNS.
      -controlNS                    specify the namespace where singleton services are installed. By default it it would be same as csNS.
      -c                            specify the subscription channel where common services switch. By default it is channel v3
Note:
If there is no namespace defined through arguments, all Common Service in the cluster will be upgraded.
```

### DIFF
--- a/upgrade_common_services.sh
+++ b/upgrade_common_services.sh
@@ -1,0 +1,388 @@
+#!/bin/bash
+#
+# Copyright 2021 IBM Corporation
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+# http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+
+# counter to keep track of installation steps
+STEP=0
+
+# script base directory
+BASE_DIR=$(dirname "$0")
+
+CS_LIST_CSNS=("operand-deployment-lifecycle-manager-app"
+        "ibm-common-service-operator"
+        "ibm-cert-manager-operator"
+        "ibm-mongodb-operator"
+        "ibm-iam-operator"
+        "ibm-monitoring-grafana-operator"
+        "ibm-healthcheck-operator"
+        "ibm-management-ingress-operator"
+        "ibm-licensing-operator"
+        "ibm-commonui-operator"
+        "ibm-ingress-nginx-operator"
+        "ibm-auditlogging-operator"
+        "ibm-platform-api-operator"
+        "ibm-namespace-scope-operator"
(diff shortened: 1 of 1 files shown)


upgrade_common_services.sh
```diff
@@ -0,0 +1,388 @@
...
+
+	Options:
+	Mandatory arguments to long options are mandatory for short options too.
+      -h, --help                    display this help and exit
+      -csNS                         specify the namespace where common service is installed. By default it is namespace ibm-common-services.
+      -cloudpaksNS                  specify the namespace where cloud paks is installed. By default it would be same as csNS.
+      -controlNS                    specify the namespace where singleton services are installed. By default it it would be same as csNS.
+      -c                            specify the subscription channel where common services switch. By default it is channel v3
+	Note:
+	If there is no namespace defined through arguments, all Common Service in the cluster will be upgraded.
```
bluzarraga, line 70: I don't think this has to be changed now, but if we were to continue with the script approach down the line, I could see this causing some headaches in multi instance clusters where specific versions of cs are installed

upgrade_common_services.sh
```diff
@@ -0,0 +1,388 @@
...
+    else
+        success "OpenShift Command Line tool oc is available."
+    fi
+
+    # checking oc command logged in
+    user=$(oc whoami 2> /dev/null)
+    if [ $? -ne 0 ]; then
+        error "You must be logged into the OpenShift Cluster from the oc command line."
+    else
+        success "oc command logged in as ${user}"
```
bluzarraga, line 149: Do we need a specific kind of user for this script? Would the user need to be a cluster admin to run upgrades?

upgrade_common_services.sh
```diff
@@ -0,0 +1,388 @@
...
+#         "ibm-zen-operator") eval "$2='1.0.0'";;
+#         "ibm-zen-cpp-operator") eval "$2='1.0.0'";;
+#         "ibm-crossplane-operator-app") eval "$2='1.0.0'";;
+#         "ibm-crossplane-provider-ibm-cloud-operator-app") eval "$2='1.0.0'";;
+#         "ibm-crossplane-provider-kubernetes-operator-app") eval "$2='1.0.0'";;
+#         *)  eval "$2='0.0.0'";;
+#     esac
+# }
+
+# --- Run ---
```
bluzarraga, line 385: can this be removed?

upgrade_common_services.sh
bitscuit: It should be changed to by default, the namespace you are currently one will be selected instead of all namespaces. All namespaces should be very explicit.

bitscuit (approved):
/lgtm

upgrade_common_services.sh
bitscuit: Will fix in another PR

PR #845, opened 2022-08-12
bump efix version to 3.20.1

qpdpQ:


bundle/manifests/ibm-common-service-operator.clusterserviceversion.yaml
```diff
@@ -485,10 +485,10 @@ spec:
   provider:
     name: IBM
   relatedImages:
-    - image: icr.io/cpopen/common-service-operator:3.20.0
+    - image: icr.io/cpopen/common-service-operator:3.20.1
       name: COMMON_SERVICE_OPERATOR_IMAGE
     - image: icr.io/cpopen/cpfs/ibm-cs-webhook:1.16.0
       name: IBM_CS_WEBHOOK_IMAGE
     - image: icr.io/cpopen/cpfs/ibm-secretshare-operator:1.17.0
       name: IBM_SECRETSHARE_OPERATOR_IMAGE
```
Daniel-Fan, line 493: Could we add a replace here for 3.20? For example,
Screen Shot 2022-08-12 at 1 33 52 PM
1

qpdpQ: nice catch! will do now

Makefile
```diff
@@ -79,7 +79,7 @@ OPERATOR_IMAGE_NAME ?= common-service-operator
 # Current Operator bundle image name
 BUNDLE_IMAGE_NAME ?= dev-common-service-operator-bundle
 
-CHANNELS := v3.20
+CHANNELS := v3.20.1
```
Daniel-Fan, line 82: Hi @qpdpQ 

We don't change the channel for cs-operator.

It will stay at `v3.20` channel for all `v3.20.x` versions

Daniel-Fan (approved):
/lgtm

PR #870, opened 2022-09-16
update changes for online and airgap cs upgrade script

YCShen1010:
Signed-off-by: YCShen1010 

Resolved spelling mistakes and added description to `compare_channel` function for both online and airgap cs upgrade script


Daniel-Fan (approved):
/lgtm

PR #875, opened 2022-09-22
keep the newest verison of NSS in controlNS

qpdpQ:
if there is a newer version in control NS, CS-operator should skip installing the older version to control NS


qpdpQ:
for this issue https://github.ibm.com/IBMPrivateCloud/roadmap/issues/55091

Daniel-Fan (approved):
/lgtm

PR #882, opened 2022-09-28
add ignore not found flag to deleting operandregistry and config

bluzarraga:
If, for some reason, the operandregistry and operandconfig are not present in the environment during runtime, the script will exit in error with the cs operator and odlm operator pods scaled to 0. This is not a clear situation to users and could be difficult to debug in a live environment. Considering these objects are being deleted, if they are not there, they should not force the script to exit.


Daniel-Fan (approved):
/lgtm

PR #883, opened 2022-09-29
temp hard code zen cpp @ v3.21 chan for cicd builds

bluzarraga:
CICD has asked us to change the channel for zen cpp operator to 3.21 as the zen team has not pushed an update to run on 3.22 channel. This fix will ultimately be reverted once zen team is ready.


Daniel-Fan (approved):
/lgtm

PR #894, opened 2022-10-18
remove crossplane provider operator

qpdpQ:
for the issue: https://github.ibm.com/IBMPrivateCloud/roadmap/issues/55208
add a new parameter `crossplaneProviderRemoval` in CommonService CR aligned with `.features.bedrockshim.enabled` 
it is an indicator to perform gracefully uninstallation for provider operator

### DIFF
--- a/controllers/bootstrap/init.go
+++ b/controllers/bootstrap/init.go
@@ -230,19 +230,12 @@ func (b *Bootstrap) CrossplaneOperatorProviderOperator(instance *apiv3.CommonSer
 				if updateErr := b.CreateorUpdateCFCrossplaneConfigMap("'false'"); updateErr != nil {
 					return updateErr
 				}
-			}
 
-			b.CSData.CrossplaneProvider = "odlm"
-			if b.SaasEnable {
-				b.CSData.CrossplaneProvider = "ibmcloud"
-			}
+				b.CSData.CrossplaneProvider = "odlm"
+				if b.SaasEnable {
+					b.CSData.CrossplaneProvider = "ibmcloud"
+				}
 
-			// install crossplane operator
-			if err := b.installCrossplaneOperator(); err != nil {
-				return err
-			}
-			// we install provider only if we don't need to delete it
-			if !removeCrossplaneProvider {
 				switch b.CSData.CrossplaneProvider {
 				case "odlm":
 					if err := b.installKubernetesProvider(); err != nil {
@@ -254,14 +247,14 @@ func (b *Bootstrap) CrossplaneOperatorProviderOperator(instance *apiv3.CommonSer
 					}
 				}
 			}
+			// install crossplane operator
+			if err := b.installCrossplaneOperator(); err != nil {
+				return err
+			}
 		} else {
 			// delete crossplane and provider operator if exist
 			if err := b.DeleteCrossplaneAndProviderSubscription(b.CSData.ControlNs); err != nil {
 				return err
-			}
-			value := "'" + strconv.FormatBool(removeCrossplaneProvider) + "'"
-			if updateErr := b.CreateorUpdateCFCrossplaneConfigMap(value); updateErr != nil {
-				return updateErr
 			}
 		}
 	} else {
@@ -386,6 +379,10 @@ func (b *Bootstrap) DeleteCrossplaneAndProviderSubscription(namespace string) er
 				klog.Errorf("Failed to delete %s in %s: %v", constant.ICPOperator, namespace, err)
 				return err
 			}
+		}
+
+		if updateErr := b.CreateorUpdateCFCrossplaneConfigMap("'true'"); updateErr != nil {
+			return updateErr
 		}
 	}
 	return nil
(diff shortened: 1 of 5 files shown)


controllers/bootstrap/init.go
```diff
@@ -371,6 +399,81 @@ func (b *Bootstrap) DeleteCrossplaneAndProviderSubscription(namespace string) er
...
+				uninstallProvider = true
+			} else {
+				uninstallProvider = false
+				break
+			}
+		} else {
+			uninstallProvider = false
+			break
+		}
+	}
```
Daniel-Fan, line 429: I think we might need to re-consider the logic here.

If we some CommonService CR's does not say anything about Crossplane, should we just ignore it?
For example,
1. if `.spec.features.bedrockshim` is `nil`, then we may should skip this CR.
2. If `.sepc.features.bedrock.shim.enable: false`, then we should skip `crossplaneProviderRemoval` as well.
3. From my own opinion, we should only consider `uninstallProvider`/`crossplaneProviderRemoval` when `.sepc.features.bedrock.shim.enable: enable`

controllers/bootstrap/init.go
```diff
@@ -196,19 +196,37 @@ func (b *Bootstrap) CrossplaneOperatorProviderOperator(instance *apiv3.CommonSer
...
+		if removeCrossplaneProvider {
+			//delete Crossplane Provider Subscription
+			klog.Info("try to delete CrossplaneProvider")
+			if DeleteErr := b.DeleteCrossplaneProviderSubscription(b.CSData.ControlNs); DeleteErr != nil {
+				return DeleteErr
+			}
+		}
 
+		if !removeCrossplaneProvider {
+			if updateErr := b.CreateorUpdateCFCrossplaneConfigMap("'false'"); updateErr != nil {
```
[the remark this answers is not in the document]
qpdpQ: yeah

controllers/bootstrap/init.go
```diff
@@ -196,19 +196,37 @@ func (b *Bootstrap) CrossplaneOperatorProviderOperator(instance *apiv3.CommonSer
...
+				return updateErr
+			}
+		}
```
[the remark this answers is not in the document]
qpdpQ: sure

controllers/bootstrap/init.go
qpdpQ: sure!

controllers/bootstrap/init.go
```diff
@@ -196,50 +196,76 @@ func (b *Bootstrap) CrossplaneOperatorProviderOperator(instance *apiv3.CommonSer
...
+			if !removeCrossplaneProvider {
+				switch b.CSData.CrossplaneProvider {
+				case "odlm":
+					if err := b.installKubernetesProvider(); err != nil {
+						return err
+					}
+				case "ibmcloud":
+					if err := b.installIBMCloudProvider(); err != nil {
+						return err
+					}
```
Daniel-Fan, line 254: Wondering that if we could integrate this part into above section on `if removeCrossplaneProvider {} else {}`

controllers/bootstrap/init.go
```diff
@@ -196,50 +196,76 @@ func (b *Bootstrap) CrossplaneOperatorProviderOperator(instance *apiv3.CommonSer
...
-		} else {
-			isLater = false
-		}
-		//isLater value of false means we install, true means we do not install
-		if !isLater {
+
+			b.CSData.CrossplaneProvider = "odlm"
+			if b.SaasEnable {
+				b.CSData.CrossplaneProvider = "ibmcloud"
+			}
```
Daniel-Fan, line 238: This part is only about `Provider`, could we integrate them under `if removeCrossplaneProvider {} else {}`

controllers/bootstrap/init.go
```diff
@@ -196,50 +196,76 @@ func (b *Bootstrap) CrossplaneOperatorProviderOperator(instance *apiv3.CommonSer
...
 		} else {
-			klog.Infof("Crossplane operator already exists at a later version in control namespace. Skipping.")
+			// delete crossplane and provider operator if exist
+			if err := b.DeleteCrossplaneAndProviderSubscription(b.CSData.ControlNs); err != nil {
+				return err
+			}
+			value := "'" + strconv.FormatBool(removeCrossplaneProvider) + "'"
+			if updateErr := b.CreateorUpdateCFCrossplaneConfigMap(value); updateErr != nil {
+				return updateErr
+			}
```
Daniel-Fan, line 265: Wondering if we should move this part into `DeleteCrossplaneAndProviderSubscription(b.CSData.ControlNs)`, like what we did for `DeleteCrossplaneProviderSubscription`

I am considering that we should go over all CommonService CRs in the cluster and determine which value should be put into ConfigMap.

qpdpQ: Yes, we should move it to DeleteCrossplaneAndProviderSubscription

controllers/bootstrap/init.go
qpdpQ: yeah good idea!

controllers/bootstrap/init.go
qpdpQ: yeah good idea

Daniel-Fan (approved):
/lgtm

PR #905, opened 2022-10-31
check that ns in cm exist, create them if they don't

bluzarraga:
Signed-off-by: Ben Luzarraga <luzarragaben@ibm.com>

Check that all of the namespaces referred to in common-service-maps configmap exist before running the rest of the script. Try to create all namespaces listed. It _does not_ create the cs-control namespace or check that field at all. The existence of that field is already checked earlier in the code and I think if that namespace does not exist yet we should encourage the user to review their topology before moving on.

Sample output from earlier testing:
```
creating ibm-common-services
Error from server (AlreadyExists): namespaces "ibm-common-services" already exists
ibm-common-services already exists, skipping...
creating ibm-common-services2
namespace/ibm-common-services2 created
creating cpns1
namespace/cpns1 created
creating ibm-common-services
Error from server (AlreadyExists): namespaces "ibm-common-services" already exists
ibm-common-services already exists, skipping...
creating cpns2
namespace/cpns2 created
all namespaces in common-service-maps exist
```

Open to conversation and feedback on implementation

This was the cm I used in my testing for reference:
```
apiVersion: v1
kind: ConfigMap
metadata:
  name: common-service-maps
  namespace: kube-public
data:
  common-service-maps.yaml: |
    controlNamespace: cs-control
    namespaceMapping:
    - requested-from-namespace:
      - ibm-common-services
      map-to-common-service-namespace: ibm-common-services
    - requested-from-namespace:
      - ibm-common-services2
      - cpns1
      map-to-common-service-namespace: cpns2
```

PR #909, opened 2022-11-01
remove existing odlm managed nss crs

bluzarraga:

For work item: 
> script should remove namespacescope CRs managed by ODLM (nss-managedby-odlm and odlm-scope-managedby-odlm), so that original instance does not have access to new instances

Open to conversation and feedback on implementation. I am not particularly familiar with how nss works or how these nss CRs get created/recreated (in my testing I manually created them after copying an existing one). I am mostly concerned with the placement of the `removeNSS` function. Right now it is called right after cs operator and odlm are scaled down.

Output looks like:
```
[INFO] patching and deleting namespace scope nss-managedby-odlm in namespace ibm-common-services
namespacescope.operator.ibm.com/nss-managedby-odlm patched
namespacescope.operator.ibm.com "nss-managedby-odlm" deleted
[INFO] patching and deleting namespace scope nss-managedby-odlm in namespace ibm-common-services2
namespacescope.operator.ibm.com/nss-managedby-odlm patched
namespacescope.operator.ibm.com "nss-managedby-odlm" deleted
[INFO] patching and deleting namespace scope odlm-scope-managedby-odlm in namespace ibm-common-services
namespacescope.operator.ibm.com/odlm-scope-managedby-odlm patched
namespacescope.operator.ibm.com "odlm-scope-managedby-odlm" deleted
[INFO] patching and deleting namespace scope odlm-scope-managedby-odlm in namespace ibm-common-services2
namespacescope.operator.ibm.com/odlm-scope-managedby-odlm patched
namespacescope.operator.ibm.com "odlm-scope-managedby-odlm" deleted
```

This will likely have some kind of merge conflict with https://github.com/IBM/ibm-common-service-operator/pull/905, but we'll cross that bridge when we get there.

### DIFF
--- a/convert_to_multi_instance.sh
+++ b/convert_to_multi_instance.sh
@@ -90,6 +90,9 @@ function prepare_cluster() {
     ${OC} delete operandregistry -n ${master_ns} --ignore-not-found common-service 
     ${OC} delete operandconfig -n ${master_ns} --ignore-not-found common-service
 
+    # remove existing namespace scope CRs
+    removeNSS
+
     # uninstall singleton services
     "${OC}" delete -n "${master_ns}" --ignore-not-found certmanager default
     "${OC}" delete -n "${master_ns}" --ignore-not-found sub ibm-cert-manager-operator
@@ -313,6 +316,21 @@ function check_CSCR() {
         fi
     done
 
+}
+
+function removeNSS(){
+    ${OC} get nss --all-namespaces | grep nss-managedby-odlm | while read -r line; do
+        local namespace=$(echo $line | awk '{print $1}')
+        info "patching and deleting namespace scope nss-managedby-odlm in namespace $namespace"
+        ${OC} patch nss nss-managedby-odlm -n ${namespace} --type=merge -p '{"metadata": {"finalizers":null}}'
+        ${OC} delete nss nss-managedby-odlm -n ${namespace} || error "unable to delete namespace scope nss-managedby-odlm in ${namespace}"
+    done
+    ${OC} get nss --all-namespaces | grep odlm-scope-managedby-odlm | while read -r line; do
+        local namespace=$(echo $line | awk '{print $1}')
+        info "patching and deleting namespace scope odlm-scope-managedby-odlm in namespace $namespace"
+        ${OC} patch nss odlm-scope-managedby-odlm -n ${namespace} --type=merge -p '{"metadata": {"finalizers":null}}'
+        ${OC} delete nss odlm-scope-managedby-odlm -n ${namespace} || error "unable to delete namespace scope odlm-scope-managedby-odlm in ${namespace}"
+    done
 }
 
 function msg() {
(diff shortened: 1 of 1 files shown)


Daniel-Fan (approved):
/lgtm

convert_to_multi_instance.sh
```diff
@@ -315,6 +318,21 @@ function check_CSCR() {
 
 }
 
+function removeNSS(){
+    ${OC} get nss --all-namespaces | grep nss-managedby-odlm | while read -r line; do
+        local namespace=$(echo $line | awk '{print $1}')
+        info "patching and deleting namespace scope nss-managedby-odlm in namespace $namespace"
+        ${OC} patch nss nss-managedby-odlm -n ${namespace} --type=merge -p '{"metadata": {"finalizers":null}}'
+        ${OC} delete nss nss-managedby-odlm -n ${namespace} || error "unable to delete namespace scope nss-managedby-odlm in ${namespace}"
```
Daniel-Fan, line 326: Hi @bluzarraga 

The place where we delete NSS CR is great!
Those NSS CR's will be created and updated by ODLM and cs-operator. It is good that we call `removeNSS` after scaling down cs-operator and ODLM.

I am also wondering that if we could try to remove NSS CR's first?
If the normal deletion is timeout/not working, we could raise a warning msg and patch the finalizer to `null` for that CR.
The main consideration is that
1. We did not touch namespace-scope operator in the script, it should be able to remove NSS CR correctly.
2. When namespace-scope operator is deleting the NSS CR, it will delete all RBAC's related to this CR, which are used to extend permission to other namespace. If we leave those RBAC in the cluster, it still grants operators the access to unexpected namespace. 

Reference on [waiting for CR to be deleted](https://github.com/IBM/ibm-common-service-operator/blob/95843d0f3c48f669c494a0f6ea4e79dc5a5e2af5/force-uninstall.sh#L324)

bitscuit: If it times out deleting, how can we determine which RBAC resources need to be deleted? We might need a separate PR to recover from this kind of situation, because it will be very difficult to troubleshoot if after conversion, the NSS CRs say one thing, but in actuality the RBAC resources are different

bitscuit (approved):
/lgtm

bitscuit:
We will need a separate discussion and PR to handle delete timeouts for the NSS CRs and properly updating any RBAC resources

convert_to_multi_instance.sh
Daniel-Fan: I think the deletion is time out only due to one situation where NSS controller is not able to remove finalizer in NSS CR.

And the only reason why NSS controller could not remove the finalizer is that the controller is gone(NSS operator not working properly)

Given the case that NSS operator should stay in the cluster after conversion, we might never patch the finalizer and always have NSS operator to handle the deletion.

If the deletion is truly time out in conversion, should we stop the script and ask user to manually troubleshooting the NSS operator first?

> how can we determine which RBAC resources need to be deleted?

If we would like to manually clean RBAC resources, we could [list and delete all RBAC with certain label in targetNamespace](https://github.com/IBM/ibm-namespace-scope-operator/blob/30e66e340420ef5b816a70c432bf6330fc2904ad/controllers/namespacescope_controller.go#L349)

PR #913, opened 2022-11-02
migrate from v1alpha1 Certs to v1 Certs

qpdpQ:

### DIFF
--- a/controllers/bootstrap/init.go
+++ b/controllers/bootstrap/init.go
@@ -110,6 +110,14 @@ type CSOperator struct {
 	APIVersion string
 }
 
+type Resource struct {
+	Name    string
+	Version string
+	Group   string
+	Kind    string
+	Scope   string
+}
+
 // NewBootstrap is the way to create a NewBootstrap struct
 func NewBootstrap(mgr manager.Manager) (bs *Bootstrap, err error) {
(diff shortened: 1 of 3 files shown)


controllers/bootstrap/init.go
```diff
@@ -1758,3 +1771,23 @@ func (b *Bootstrap) DeployCertManagerCR() error {
...
+	}
+
+	if err := b.Client.Delete(context.TODO(), v1alpha1Resource); err != nil {
+		if !errors.IsNotFound(err) {
+			return err
+		}
+	}
+
+	return nil
+}
```
Daniel-Fan, line 1793: Hi @YuchenLi223 

May I ask that if we could have a more generic function for deleting resources based on Group, Version and Kind?
We could refer to the existing function here to see if we could combine them together.
https://github.com/IBM/ibm-common-service-operator/blob/c1ee67b6c687d2a925523618a1a87dfbb26bbefb/controllers/goroutines/cleanup_deprecated_services.go#L229

qpdpQ: Ah I didn't find it, will use `cleanup` function now

Daniel-Fan (approved):
/lgtm

PR #914, opened 2022-11-04
cleanup zen instances before conversion

bluzarraga:

Cleanup zen instances before conversion process by patching out the csNamespace key in zenservice and deleting the iam config job and the zen client in cloud pak namespace. 

Open to feedback, related to https://github.com/IBM/ibm-common-service-operator/pull/905 and https://github.com/IBM/ibm-common-service-operator/pull/909. Will likely be merge conflicts

Sample output:
```
zenservice.zen.cpd.ibm.com/iaf-zen-cpdservice patched
job.batch "iam-config-job" deleted
client.oidc.security.ibm.com "zenclient-cpns1" deleted
zenservice.zen.cpd.ibm.com/iaf-zen-cpdservice patched
[INFO] CS Namespace not defined in iaf-zen-cpdservice in cpns2. Moving on...
No zen client in cpns2. Moving on...
```

### DIFF
--- a/convert_to_multi_instance.sh
+++ b/convert_to_multi_instance.sh
@@ -316,11 +316,11 @@ function cleanupZenService(){
             info "iam-config-job not present in namespace ${namespace}. Moving on..."
         fi
 
-        #TODO may need to patch out finalizers in zen client
         # delete zen client
         return_value=$(${OC} get client -n ${namespace} || echo "failed")
         if [[ $return_value != "failed" ]]; then
             zenClient=$(${OC} get client -n ${namespace} | awk '{if (NR!=1) {print $1}}')
+            ${OC} patch client ${zenClient} -n ${namespace} --type=merge -p '{"metadata": {"finalizers":null}}'
             ${OC} delete client ${zenClient} -n ${namespace}
         else
             info "No zen client in ${namespace}. Moving on..."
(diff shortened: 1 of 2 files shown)


convert_to_multi_instance.sh
```diff
@@ -84,6 +84,8 @@ function prepare_cluster() {
     # TODO for more advanced checking
     # find all namespaces with cs-operator running
     # each namespace should be in configmap
+    
+    cleanupZenService $cm_name
```
bitscuit, line 88: move this to after the NSS CRs are removed according to discussion with Simon

bluzarraga:
Looking at error handling nss. As it is now, if the nss specified in the removeNSS function are not present, the script breaks. If it makes sense to include it in this pr I will otherwise I will open another

bitscuit:
fix in a new PR for better tracking

bluzarraga:
nss error handling https://github.com/IBM/ibm-common-service-operator/pull/917

bitscuit:
I think we can change the function to use yq, similar to `check_cm_ns_exist` function, to read the namespaces

bitscuit (approved):
/lgtm

bluzarraga:
Need to add some kind of error handling/check for zen resources on a system. If there aren't any zen resources the script will fail trying to clean them up

bluzarraga:
Yes, if a zenclient or zenservice does not exist in a namespace the return is `No resources found in ibm-common-services namespace.` but it does not trigger the `echo "failed"` so the `return_value` is `""` instead of `failed`. We would either need to grep on something and it return nothing (which could involve a little guess work) or look for an empty response.

PR #924, opened 2022-11-15
add mongodb files

kgcarr:
Signed-off-by: Kevin Carr <kgcarr@us.ibm.com>


bluzarraga:
Spoke offline with @kgcarr, we will come back to storage class name in the automation pr https://github.com/IBM/ibm-common-service-operator/pull/925. As it is now, the $STGCLASS value used in the backup script could refer to a storageclass that does not have `persistentVolumeReclaimPolicy: Retain` which would not work for our purposes but is not an issue for the velero backup process.

PR #925, opened 2022-11-15
Automate the Mongodb Backup and Restore Steps

bluzarraga:

### DIFF
--- a/backup_restore_mongo.sh
+++ b/backup_restore_mongo.sh
@@ -29,6 +29,7 @@ function main() {
     prereq
     prep_backup
     backup
+    prep_restore
 }
 
 # verify that all pre-requisite CLI tools exist and parameters set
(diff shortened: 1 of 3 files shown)


bluzarraga:
Successfully completed a full test of the script! What's left:
- possible error handling
- verification of IAM data transferred
- update mongodb-backup.sh to accept optional parameter to ensure the right storage class is used for the cs-mongodump pvc created in the script
The first two items may be better served in documentation than added to the script. For error handling, the script will fail if previous attempts are not cleaned up (need to clean up the storage class, pv, pvc, and possibly jobs/pods created by jobs) but it may make debugging more difficult if these items are automatically cleaned up.

bitscuit (approved):
/lgtm

bitscuit:
We will need to discuss about error handling more since a decision needs to be made about how to handle re-running a failed attempt. My opinion is that we try to re-use existing PV and PVC, to allow for re-running and idempotence. If it turns out that PV is the root cause of issues, then we can document deleting it and look at how to better automate after investigating those kinds of issues. At this point, we're not knowledgeable enough in the area to predict specific problems

PR #934, opened 2022-11-24
add NSS ConfigMap synchronization

Daniel-Fan:

### DIFF
--- a/main.go
+++ b/main.go
@@ -161,6 +161,8 @@ func main() {
 	go goroutines.CreateUpdateConfig(bs)
 	// Update CS CR Status
 	go goroutines.UpdateCsCrStatus(bs)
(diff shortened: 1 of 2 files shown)


main.go
```diff
@@ -161,6 +161,8 @@ func main() {
 	go goroutines.CreateUpdateConfig(bs)
 	// Update CS CR Status
 	go goroutines.UpdateCsCrStatus(bs)
+	// Sync up NSS ConfigMap
+	go goroutines.SyncUpNSSConfigMap(bs)
```
pgodowski, line 165: Question: isn't it that at the upgrade scenario (3.x -> 4.x) we need to also do the reverse, i.e. to populate the OperatorGroup based on the NSS `namespace-scope` ConfigMap?

PR #964, opened 2023-01-03
check v1alpha1 crd

qpdpQ:

### DIFF
--- a/controllers/bootstrap/init.go
+++ b/controllers/bootstrap/init.go
@@ -1015,9 +1015,22 @@ func (b *Bootstrap) DeployCertManagerCR() error {
 		}
 	}
 	klog.Info("Deploying Cert Manager CRs")
+
+	dc := discovery.NewDiscoveryClientForConfigOrDie(b.Config)
+	v1alpha1Exist := true
+
 	for _, kind := range constant.CertManagerKinds {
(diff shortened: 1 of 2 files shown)


controllers/bootstrap/init.go
```diff
@@ -1046,9 +1059,11 @@ func (b *Bootstrap) DeployCertManagerCR() error {
 		},
 	}
 
-	for _, resource := range resourceList {
-		if err := b.Cleanup(b.CSData.ServicesNs, resource); err != nil {
-			return err
+	if v1alpha1Exist {
+		for _, resource := range resourceList {
+			if err := b.Cleanup(b.CSData.ServicesNs, resource); err != nil {
```
Daniel-Fan, line 1064: I am wondering that if we could integrate the resource check into the clean up function? If the CRD does not exists, then we simply return `nil` error.

The reason I am saying this is that there are somewhere else also using the clean up function. https://github.com/IBM/ibm-common-service-operator/blob/release-future/controllers/goroutines/cleanup_deprecated_services.go#L197

Daniel-Fan: Meanwhile, wondering if we do integrate the check into clean up function. Would we follow the same logic for CD stream?

In CD stream, we only wait for `v1` CRD, and delete `v1alpha1` Certs and issuers when they exists.

Daniel-Fan (approved):
/lgtm

PR #967, opened 2023-01-04
Update conversion process to make it more fail tolerant and increase stability

bluzarraga:
Added a cleanup function to the mongodb backup/restore script that runs before and after the script to remove any old resources created. This script is very fragile if these old resources exist so it's best to have a clean slate when starting, restarting, or finishing. I have been able to error out of the script (both manually and otherwise) and re-run it successfully because of the cleanup function.

The script also checks if the files already exist in the directory before trying to download for airgap purposes.


backup_restore_mongo.sh
```diff
@@ -173,19 +201,88 @@ function prep_restore() {
...
     export CS_NAMESPACE=$TARGET_NAMESPACE
-    wget https://raw.githubusercontent.com/IBM/ibm-common-service-operator/scripts/velero/restore/mongoDB/mongodbrestore.yaml
-    wget https://raw.githubusercontent.com/IBM/ibm-common-service-operator/scripts/velero/restore/mongoDB/set_access.js
-    wget https://raw.githubusercontent.com/IBM/ibm-common-service-operator/scripts/velero/restore/mongoDB/mongo-restore.sh
+
+    #check if files are already present on machine before trying to download (airgap)
+    if [[ -f "mongodbrestore.yaml" ]]; then
+        info "mongodbrestore.yaml already present"
+    else
+        wget https://raw.githubusercontent.com/IBM/ibm-common-service-operator/scripts/velero/restore/mongoDB/mongodbrestore.yaml
```
Daniel-Fan, line 212: Do we wanna check the return code of `wget`?

The reason for asking is that if it is a truly airgap env, would it be able to download?

bluzarraga: The assumption I made for airgap was if it is an airgap environment, the files would need to already exist on the machine before the script runs which the if statement checks for. If the file does not already exist, then it grabs with wget

bluzarraga: But I checking return code is a good idea. There should be output stating whether or not the items could be downloaded with the response code.

bluzarraga:
I added a rollback process that lets the user rollback their multi-instance installation back to a shared instance installation. The caveat is that this presently only works one there is one instance installed so the common-service-maps cm would look something like this:
```
apiVersion: v1
kind: ConfigMap
metadata:
  name: common-service-maps
  namespace: kube-public
data:
  common-service-maps.yaml: |
    controlNamespace: cs-control
    namespaceMapping:
    - requested-from-namespace:
      - cpns1
      map-to-common-service-namespace: ibm-common-services
```

I will be focusing on uninstalling the added instances now so this rollback can be more widely used as I don't imagine many multi-instance scenarios with only one instance.

A couple of notes:
- I think the rollback/uninstall might be better off as their own script(s) instead of included in the conversion script
- there will need to be manual input to edit the common-service-maps cm as using yq to manipulate the `controlNamespace` field is pretty tricky given the nested yaml file. This will probably need to be the case for uninstall as well as we will need to consolidate requested-from-namespace so they map to the "original" common services instance before rollback.

bluzarraga:
Pushed a commit that creates a new multi instance uninstall script that largely repurposes the existing `force-uninstall.sh` to only uninstall a specific common services instance in a multi instance environment. I had a successful test run yesterday uninstalling one instance and leaving two others untouched. Unfortunately, this script is very messy still. There is an issue with the `wait_for_deleted` process (more specifically the `update_remaining_resources` function) that outputs confusing messages and always takes the full wait period. 

On top of additional testing, this script also needs clearer logging as well as some enhancements with error handling.

bluzarraga:
Put the rollback script into its own file. In my own testing, I have been able to go from this topology:
```
apiVersion: v1
kind: ConfigMap
metadata:
  name: common-service-maps
  namespace: kube-public
data:
  common-service-maps.yaml: |
    controlNamespace: cs-control
    namespaceMapping:
    - requested-from-namespace:
      - cpns1
      map-to-common-service-namespace: ibm-common-services
    - requested-from-namespace:
      - cpns2
      map-to-common-service-namespace: csns2
    - requested-from-namespace:
      - cpns3
      - cpns4
      map-to-common-service-namespace: csns3
```
to
```
apiVersion: v1
kind: ConfigMap
metadata:
  name: common-service-maps
  namespace: kube-public
data:
  common-service-maps.yaml: |
    namespaceMapping:
    - requested-from-namespace:
      - cpns1
      - cpns2
      - cpns3
      - cpns4
      map-to-common-service-namespace: ibm-common-services
```

without error

Daniel-Fan (approved):
/lgtm

PR #996, opened 2023-01-30
add OperatorGroup synchronization with NSS CR

Daniel-Fan:
This is a placeholder for continuing using NSS CR for RBAC projection under advanced deployment topology(separation of duties)


Daniel-Fan:
Close the PR, OperatorGroup will not be used anymore

PR #998, opened 2023-01-30
update AllNamespaces Installations Mode

Daniel-Fan:
1. User installs the CS operator cluster scoped.
2. If the `ibm-common-services` namespace is not there, the operator waits.
3. If it's there, it creates the CR in the same namespace as operator(openshift-operators) with
    ```
    servicesNamespace: ibm-common-services
    operatorNamespace: openshift-operators
    ```
4. If the user never creates `ibm-common-services` namespace, choose to create it own CommonService CR in the same namespace as operator(openshift-operators), cs-operator accept that and move on.


YCShen1010 (approved):
/lgtm

PR #1003, opened 2023-01-31
Patch 1 to add default admin username

rashmi43:


YCShen1010:
default admin username was added in `CommonService` CR, [PR](https://github.com/IBM/ibm-common-service-operator/pull/1117) merged, closing here

PR #1018, opened 2023-02-13
remove CSV dependency on ODLM, instead installing ODLM in runtime

Daniel-Fan:
### Context

Currently cs-operator has a CSV dependency on ODLM, meaning when cs-oeprator is installed in one namespace, OLM will help to automatically install the available ODLM in the same namespace as dependency of cs-oeprator.

### Necessary Update

It is possible that multiple copies of cs-operators will be installed in the same tenant(in both Cloud Pak ns and CS ns). We would not simply rely on OLM to install ODLM whenever cs-operator is installed - because only one copy of ODLM is allowed in the same tenant.
As the result, we will ask cs-operator to identify its role(identify it is in CS ns or Cloud Pak ns), then create the subscription of ODLM when it is necessary.

### How to test
1. Deploy `future` CatalogSource containing ODLM
```
apiVersion: operators.coreos.com/v1alpha1
kind: CatalogSource
metadata:
  name: opencloud-operators
  namespace: openshift-marketplace
spec:
  displayName: IBMCS Test Operators
  image: >-
    docker-na-public.artifactory.swg-devops.com/hyc-cloud-private-daily-docker-local/ibmcom/ibm-common-service-catalog:future
  publisher: IBM
  sourceType: grpc
  updateStrategy:
    registryPoll:
      interval: 45m
```
2. Deploy a CatalogSource containing new cs-operator bundle - (no CSV dependency on ODLM)
```
apiVersion: operators.coreos.com/v1alpha1
kind: CatalogSource
metadata:
  name: cs-test-operators
  namespace: openshift-marketplace
spec:
  displayName: IBMCS only Operators
  image: 'quay.io/daniel_fan/common-service-operator-catalog:525733a6-dirty'
  publisher: IBM
  sourceType: grpc
  updateStrategy:
    registryPoll:
      interval: 45m
```

3. install cs-operator from new CatalogSource `cs-test-operators`
4. We will notice that ODLM subscription is created in the namespace
   - It is not created by OLM, instead it is created by cs-operator
   - By default, cs-operator will create the ODLM subscription, pointing ODLM subscription to the same CatalogSource(`cs-test-operators`) as cs-operator. However, this CatalogSource(`cs-test-operators`) does not contain ODLM bundle information. So subscription in `unknown` status.
5. Update CommonService CR `common-service` to specify the CatalogSource used by ODLM
```
kind: CommonService
metadata:
  name: common-service
  namespace: odlm
spec:
  catalogName: opencloud-operators
```
6. ODLM subscription now is pointing to the correct CatalogSource `opencloud-operators`, and is installed successfully.


YCShen1010 (approved):
/lgtm

PR #1021, opened 2023-02-13
Update status check on deployed services

Daniel-Fan:
When any error happens on fetching OperandRegistry and OperandConfig, return `false` for `serviceDeployed` status.
It should not block any installation or namespace changing for `operatorNamespace` and `servicesNamespace`


YCShen1010 (approved):
/lgtm

PR #1041, opened 2023-02-22
set resourceVersion before updating an CR

Daniel-Fan:
Solve an error for resourceVersion is nil when updating a resource
```
metadata.resourceVersion: Invalid value: 0x0: must be specified for an update
```


YCShen1010 (approved):
/lgtm

PR #1047, opened 2023-02-28
add fail to check

bluzarraga:
I need to test this on an environment with zen installed but without zen clients created. The error from the issue seems to come from this code block. Hopefully adding the "fail" message to the check as it expects will resolve this.

### DIFF
--- a/convert_to_multi_instance.sh
+++ b/convert_to_multi_instance.sh
@@ -362,7 +362,7 @@ function cleanupZenService(){
         fi
 
         # delete zen client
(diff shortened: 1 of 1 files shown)


Daniel-Fan:
Hi @bluzarraga 

Do we wanna merge this change and have Zen team to give a try?

bluzarraga:
Yeah that's a good idea. I have not been able to recreate the issue they saw

Daniel-Fan (approved):
/lgtm

PR #1051, opened 2023-03-02
Exclude cloned CommonService CR from being copied in AllNamespace Mode

YCShen1010:
Issue: https://github.ibm.com/IBMPrivateCloud/roadmap/issues/57453
Reason: when there is only one namespace (`operatorNs`) in `WatchNamespaces`, skip copying `CommonService` CR into other namespaces.


Daniel-Fan:
Verify with simple deployment topology where `WATCH_NAMESPACE: <operator-namespace>` and AllNamespace Installation mode where `WATCH_NAMESPACE: ''`.

In both cases, there is no need to copy cloned CommonService CR to other namespace.

Daniel-Fan (approved):
/lgtm

PR #1052, opened 2023-03-02
skip common-service-maps creation in AllNamespace mode installation

YCShen1010:
verified by image: [quay.io/yuchen_shen/cs_operator:csmaps_skip_allns](http://quay.io/yuchen_shen/cs_operator:csmaps_skip_allns)

### DIFF
--- a/main.go
+++ b/main.go
@@ -167,16 +167,19 @@ func main() {
 
 		cm, err := util.GetCmOfMapCs(bs.Reader)
 		if err != nil {
-			// Create new common-service-maps
-			if errors.IsNotFound(err) {
-				klog.Infof("Creating common-service-maps ConfigMap in kube-public")
-				if err = bs.CreateCsMaps(); err != nil {
-					klog.Errorf("Failed to create common-service-maps ConfigMap: %v", err)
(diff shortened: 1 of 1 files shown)


main.go
```diff
@@ -167,16 +167,19 @@ func main() {
 
 		cm, err := util.GetCmOfMapCs(bs.Reader)
 		if err != nil {
-			// Create new common-service-maps
-			if errors.IsNotFound(err) {
-				klog.Infof("Creating common-service-maps ConfigMap in kube-public")
-				if err = bs.CreateCsMaps(); err != nil {
-					klog.Errorf("Failed to create common-service-maps ConfigMap: %v", err)
+			// Skip common-service-maps creating when installing in AllNamespace Mode
+			if bs.CSData.WatchNamespaces != "" {
```
Daniel-Fan, line 171: I am wondering if we could place this condition before getting ConfigMap?

The reason for this is to avoid unnecessary API Call given we are not worried about ConfigMap at all when it is deployed in All Namespaces Mode

Daniel-Fan (approved):
/lgtm

YCShen1010:
issue: https://github.ibm.com/IBMPrivateCloud/roadmap/issues/56488

PR #1055, opened 2023-03-06
update backup and restore with zen licensing and cert-manager

kgcarr:


bluzarraga:
Spoke to Kevin about this over slack, he's going to make sure the changes to the mongodb backup/restore are not too specific to any one use case since we currently use what's in the velero/backup/mongodb and velero/restore/mongodb folders in the backup_restore_mongo.sh script. Going to mark it as "on hold"

PR #1057, opened 2023-03-06
update script to refresh zenservice after conversion

bluzarraga:
After conversion process, I noticed whatever CP4I install was moved to the new cs instance was not working properly. Discussed it with CP4I and Zen and realized the zenservice needed to be reconciled with the new cs instance. These changes attempt to implement this.

As a part of these changes, I also added a wait function that checks IAM status. Unfortunately, we are in the midst of changing how this status is projected so the `check_IAM` function will likely need to change in the coming month or so.

Still need to test these changes as well. I tested the individual commands but not the entire script all the way through yet.

To test:
- setup environment with two cloudpaks sharing cs instance
- convert to multi instance using this version of the conversion script
- attempt to login with the new default admin credentials from new cs instance to the cloudpak instance mapped to the new cs instance

If you get the new admin credentials and after logging in get `Internal Server Error` on the `cpd-<cloudpak namespace>` route, it did not work. If after logging in with the new admin credentials you see the dashboard for `cpd-<cloudpak namespace>` then it worked.


bluzarraga:
Individual testing successful

Daniel-Fan (approved):
/lgtm

PR #1059, opened 2023-03-07
refresh operatorNs and servicesNs for each loop

Daniel-Fan:
### Context
ibm-common-serivce-operator needs to check the IAM deployments and jobs in `serivcesNamespace`.

However, ibm-common-serivce-operator is not restarted every time when `servicesNamespace` is changed.
As a result, cs-operator does not load correct `servicesNamespace` for IAM status after `serivcesNamespace` has been updated in CommonService CR.

As a workarounds, we need to refresh those namespaces values for each loop


YCShen1010 (approved):
/lgtm

PR #1066, opened 2023-03-08
Migrate License Service CMs during conversion process

bluzarraga:
We need to migrate over LicenseService data. This data was specified in 57531 and consists of multiple configmaps that may or may not be on the cluster. 
- Iterate through the list check if present
- if present
    - output cm to a temp yaml file
    - edit the namespace to match the control from the common-service-maps
    - apply the new configmap
    - delete temp file

To test:
- Setup cluster with Licensing installed, check for configmaps specified in `POSSIBLE_CONFIGMAPS` list in cs namespace
- check output for cm related messages
- verify script converts cluster successfully
- check control namespace for any of the licensing configmaps in the `POSSIBLE_CONFIGMAPS` list

I have not tested these changes myself yet. Confirming with Licensing team that these steps are appropriate as well.

### DIFF
--- a/convert_to_multi_instance.sh
+++ b/convert_to_multi_instance.sh
@@ -106,6 +106,7 @@ function prepare_cluster() {
     # This makes recovery from simple pre-requisite errors easier.
     return_value=$(("${OC}" get crd ibmlicenseservicereporters.operator.ibm.com > /dev/null && echo exists) || echo fail)
     if [[ $return_value == "exists" ]]; then
+        migrate_lic_cms $master_ns $controlNs
         "${OC}" delete -n "${master_ns}" --ignore-not-found ibmlicensing instance
     fi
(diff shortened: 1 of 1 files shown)


bluzarraga:
Tested this on my own cluster successfully. I had trouble actually deploying a a licensing instance (it requires cloud pak deployment) with any of these configmaps so I tested the `migrate_lic_cms` function using dummy configmaps in dummy namespaces (if I tried manually creating them in ibm-common-services they would be deleted). I added some of the configmaps that are created by default to the function during testing and made tweaks based on what I saw running the full conversion script with those cms. I think it would be best for SERT team to test this conversion script with a more robust deployment for highest level of confidence but it seems to be working.

Daniel-Fan (approved):
/lgtm

PR #1069, opened 2023-03-10
restart auth idp pod after restore

bluzarraga:
After investigating 57624, we decided we would need to restart the auth idp pod as a part of the backup restore script. I tested it on my environment and it is working.

To test:
- install cs and instantiate LDAP users (easiest way is through cs dev tools)
- convert cluster to multi instance
- run this version of backup restore script, ensure it completes successfully
- verify can login to new cs instance cp-console with restored LDAP user

### DIFF
--- a/backup_restore_mongo.sh
+++ b/backup_restore_mongo.sh
@@ -33,6 +33,7 @@ function main() {
     prep_restore
     restore
     cleanup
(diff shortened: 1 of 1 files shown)


Daniel-Fan (approved):
/lgtm

PR #1070, opened 2023-03-10
feat: Remove Db2U from IBM common services

emily-chan-git:
https://github.ibm.com/DB2/tracker/issues/32036

As of Cloud Pak 3.0, Db2U will no longer be included in the OperandRegistry for IBM Common Services.
* CP4D is already creating the subscription for Db2U operator instead of using Operand Requests
* CP4S (Guardium) will switch from using Operand Requests to explicitly creating a sub and using the standalone channels


YCShen1010 (approved):
/lgtm

PR #1072, opened 2023-03-12
replace supportStatus by installMode: no-op in OperandRegistry CRD

YCShen1010:


Daniel-Fan:
@YCShen1010 

Could you please [update the `golangci-lint-version` to `1.51.x` to support golang 1.20.](https://github.com/IBM/ibm-common-service-operator/blob/f779f0f3e58952105eb128506d9571602ea12f1a/common/config/.golangci.yml#L3)

It is similar to this PR: https://github.com/IBM/operand-deployment-lifecycle-manager/pull/907

PR #1078, opened 2023-03-15
set service account for permission issues

bluzarraga:
If the SCC is set to anything other than the typical default of `restricted`, the backup job could fail due to lack of permissions. Adding the same service account used by mongo resolves this problem.

Tested on the same environment where this problem was first seen where the default SCC was changed to `anyuid`.

To test:
- setup multi instance env with SCC changed from `restricted`
- run backup process, ensure job starts and completes

@kgcarr fyi


Daniel-Fan (approved):
/lgtm

PR #1079, opened 2023-03-15
change how cs op sub is found and updated

bluzarraga:
Right now, we look for cs operator subscription based on the name of the sub. Cloud paks sometimes create the cs operator subscription under a different name (ie cp4i uses `ibm-common-service-operator-v3.23-ibm-operator-catalog-openshift-marketplace` as a sub name) which has caused the script to try to install a second instance of cs operator in a namespace that results in the second cs operator instance sitting in limbo. This has been updated in this pr.

Similarly, if the cs operator already exists in the cloud pak namespace, the version of common services installed will be determined by the subscription that is already present in the cloud pak namespace. This could cause some mix-ups in what catalog source is used and consequently what versions of services are available. This change presently enforces the original cs operator's catalog source to be used cluster wide.

To test
- setup environment with two cloud paks installed
- if possible, change catalog source for cs operator installed in cloud pak namespaces so it does not match common services catalog source
- change name of cs operator subs in cloud pak namespaces
- verify there is not a cs operator in limbo in the cloud pak namespaces
- verify the catalog sources match across cs operator subscriptions

### DIFF
--- a/convert_to_multi_instance.sh
+++ b/convert_to_multi_instance.sh
@@ -25,6 +25,11 @@ YQ=${1:-yq}
 
 cs_operator_channel=
 catalog_source=
+requestedNS=
+mapToCSNS=
+
+cm_name="common-service-maps"
+    
 function main() {
     msg "Conversion Script Version v1.0.0"
     prereq
@@ -42,7 +47,6 @@ function prereq() {
     which "${OC}" || error "Missing oc CLI"
     which "${YQ}" || error "Missing yq"
     
-    local cm_name="common-service-maps"
     return_value=$("${OC}" get -n kube-public configmap ${cm_name} > /dev/null || echo failed)
     if [[ $return_value == "failed" ]]; then
         error "Missing configmap: ${cm_name}. This must be configured before proceeding"
@@ -82,12 +86,11 @@ function prereq() {
 
 function prepare_cluster() {
 
-    local cm_name="common-service-maps"
     # TODO for more advanced checking
     # find all namespaces with cs-operator running
     # each namespace should be in configmap
     # all namespaces in configmap should exist
-    check_cm_ns_exist $cm_name
+    check_cm_ns_exist
 
     ${OC} scale deployment -n ${master_ns} ibm-common-service-operator --replicas=0
     ${OC} scale deployment -n ${master_ns} operand-deployment-lifecycle-manager --replicas=0
@@ -160,11 +163,11 @@ function migrate_lic_cms() {
                 #edit the file to change the namespace to controlNs
                 yq '.metadata.namespace = "'${controlNs}'"' tmp.yaml
                 ${OC} apply -f tmp.yaml
-                rm tmp.yaml -f
                 info "Licensing configmap $cm copied from $namespace to $controlNs"
             fi
         fi
     done
+    rm tmp.yaml -f
     success "Licensing configmaps copied from $namespace to $controlNs"
 }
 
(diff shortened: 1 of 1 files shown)


bluzarraga:
I tested this and saw both issues resolve with these changes.

However, for 57770, I am a little hesitant to merge the changes in. Copy pasting comment from issue:

The pr above works in that it changes the catalog source to match the one used for the original common services but is there any reason why we would not want this? The only argument I can think of is if zen is included in other catalogs and possibly at a different version, then we could see issues like the case I observed that caused me to open this ticket (see issue statement). With some cloud paks wanting private catalog sources and possibly pinning those catalog sources, do we want to make the final determination on which catalog source should be used? What other option do we have?

convert_to_multi_instance.sh
```diff
@@ -174,13 +178,19 @@ function scale_up_pod() {
...
     info "channel:${cs_operator_channel}"   
     catalog_source=$(${OC} get sub ibm-common-service-operator -n ${master_ns} -o yaml | yq ".spec.source")
-    info "catalog_source:${catalog_source}"   
+    info "catalog_source:${catalog_source}"
+    
+    local cm_name="common-service-maps"
+    #this command gets all of the ns listed in requested from namesapce fields
+    requestedNS=$("${OC}" get configmap -n kube-public -o yaml ${cm_name} | yq '.data[]' | yq '.namespaceMapping[].requested-from-namespace' | awk '{print $2}')
+    #this command gets all of the ns listed in map-to-common-service-namespace
+    mapToCSNS=$("${OC}" get configmap -n kube-public -o yaml ${cm_name} | yq '.data[]' | yq '.namespaceMapping[].map-to-common-service-namespace' | awk '{print}')
```
bitscuit, line 192: these look like global variables, so suggest declare them at the top

convert_to_multi_instance.sh
```diff
@@ -335,6 +339,27 @@ function refresh_zen(){
...
+    for namespace in $requestedNS
+    do
+        # remove cs namespace from zen service cr
+        return_value=$(${OC} get sub -n ${namespace} | (grep ibm-common-service-operator || echo "fail"))
+        if [[ $return_value != "fail" ]]; then
+            local sub=$(${OC} get sub -n ${namespace} | grep ibm-common-service-operator | awk '{print $1}')
+            ${OC} get sub ${sub} -n ${namespace} -o yaml > tmp.yaml 
+            ${YQ} '.spec.source = "'${catalog_source}'"' tmp.yaml || error "Could not replace catalog source for CS operator in namespace ${namespace}"
+            ${OC} apply -f tmp.yaml
+            rm tmp.yaml -f
```
bitscuit, line 354:
```suggestion
```

convert_to_multi_instance.sh
```diff
@@ -335,6 +339,27 @@ function refresh_zen(){
...
+            info "Common Service Operator Subscription in namespace ${namespace} updated to use catalog source ${catalog_source}"
+        else
+            info "No Common Service Operator in namespace ${namespace}. Moving on..."
+        fi
+        return_value=""
+    done
```
bitscuit, line 360:
```suggestion
    done
    rm tmp.yaml -f
```

Small optimization. No point in removing the tmp.yaml in the loop because each iteration will overwrite the file anyways

bitscuit (changes requested):

convert_to_multi_instance.sh
```diff
@@ -335,6 +341,27 @@ function refresh_zen(){
...
+    title "Checking subs of Common Service Operator in Cloudpak Namespaces"
+    msg "-----------------------------------------------------------------------"
+    for namespace in $requestedNS
+    do
+        # remove cs namespace from zen service cr
+        return_value=$(${OC} get sub -n ${namespace} | (grep ibm-common-service-operator || echo "fail"))
+        if [[ $return_value != "fail" ]]; then
+            local sub=$(${OC} get sub -n ${namespace} | grep ibm-common-service-operator | awk '{print $1}')
+            ${OC} get sub ${sub} -n ${namespace} -o yaml > tmp.yaml 
+            ${YQ} '.spec.source = "'${catalog_source}'"' tmp.yaml || error "Could not replace catalog source for CS operator in namespace ${namespace}"
```
bitscuit, line 354:
```suggestion
            ${YQ} -i '.spec.source = "'${catalog_source}'"' tmp.yaml || error "Could not replace catalog source for CS operator in namespace ${namespace}"
```

yq will print to stdout by default and does not update the file provided. Need the inplace flag

bitscuit (approved):
/lgtm

PR #1092, opened 2023-03-20
refresh kafkaclaims

bluzarraga:
I was able to test this in a limited fashion. The only env I have access to that currently has kafka deployed is being used for cp4ba testing so could only test the relevant code change and not the full end-to-end.

Ideally, this script copies the existing kafkaclaim to a file, patches and deletes the existing kafkaclaim, then attempts to reapply the kafkaclaim based on the copy. This would hopefully resolve the kafka issue seen when setting up the cp4ba env.

To test:
- deploy shared env with kafka installed
- verify script completion
- verify kafka statefulset exists in cloud pak namespace (could be in cs namesapce but will probably be in cloud pak ns)
- verify kafkaclaim READY=True

### DIFF
--- a/convert_to_multi_instance.sh
+++ b/convert_to_multi_instance.sh
@@ -39,6 +39,7 @@ function main() {
     restart_CS_pods
     install_new_CS
     refresh_zen
+    refresh_kafka
 }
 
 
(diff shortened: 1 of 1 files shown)


Daniel-Fan (approved):
/lgtm

convert_to_multi_instance.sh
```diff
@@ -341,6 +342,66 @@ function refresh_zen(){
...
+            return_value=$(${OC} get kafkaclaim -n ${namespace} || echo "fail")
+            if [[ $return_value != "fail" ]]; then
+                if [[ $return_value != "" ]]; then
+                    kafkaClaims=$(${OC} get kafkaclaim -n ${namespace} | awk '{if (NR!=1) {print $1}}')
+                    #TODO copy kc to file, delete original kc, re-apply copied file (check for an existing of the same name)
+                    for kc in $kafkaClaims
+                    do
+                        ${OC} get kafkaclaim -n ${namespace} $kc -o yaml > tmp.yaml
+                        ${OC} patch kafkaclaim ${kc} -n ${namespace} --type=merge -p '{"metadata": {"finalizers":null}}'
+                        ${OC} delete kafkaclaim ${kc} -n ${namespace}
```
bitscuit, line 386: Nulling the finalizer is for forceful deletes, so it should happen after running `oc delete`.

Though patching finalizer now comes with a potential error, so that would have to be handled because of the way conversion script error handling works.

PR #1097, opened 2023-03-21
Add checking singleton exists or not and skipping install

mandymchu:
1. Introduced two flags `--check-cert-manager` and `--check-licensing` to implement singleton checking and skipping installation.
2. The two flags are not included in the usage so that they are not exposed to other users.
3. Olm-utils side could directly call `setup_singleton.sh --check-cert-manager` to check whether cert manager is installed or not. The scripts will exit with 1 if no operator found after one minute waiting.


mandymchu:
close it as it is for wrong branch

PR #1099, opened 2023-03-21
[WIP] make conversion script able to run multiple times on the same cluster

bluzarraga:
Right now the conversion script is only capable of being run once on a cluster. There are a few reasons why this could be a problem:
- error in setting up the cluster (ie typo in common-service-maps)
- clusters changing needs (ie adding or removing namespaces to a given cs instance)

In addition to running the script multiple times, the conversion script will also need to change slightly to fit the pattern Mike displayed for the upgrade process. The conversion script already does most of what he drew up but we need to be able to break it down by requested-from-namespace/map-to-namespace groupings (aka isolate).

I propose checking the topology first, getting the groupings that need to be updated, then running the rest of the script. At the moment, I lump in all the namespaces that need to be updated together (activeRequestedFrom & activeMapTo) but we could break it down even further and individually call all of the functions required with the specific subset of namespaces that need to be updated from the check_topology function. Breaking it down like that would let us be more specific in what we update and when. It could also let us incorporate backup/restore more seamlessly.

I opened this PR as a jumping off point for discussion. Feedback is appreciated


bluzarraga:
Update on this pr since there is a lot going on here:
- conversion script
    - updated to run more than once on a cluster
    - what's missing
        - need to remove/update nss in affected namespaces (if a cs instance loses a namespace, it needs to be removed from nss)
        - need to determine which namespaces should be updated and leave the others alone (ie don't restart or delete anything unnecessarily)
- isolate.sh
    - should be ready to test. Has a relatively limited scope but should be functional (in my experience it has been)
- backup_preload_mongo.sh
    - the process is working end to end alongside the conversion script
        - you run this before conversion and the new mongo should pickup the restored PV on startup
    - what's missing
        - two configmaps for some reason do not create properly when included in the script
            - icp-mongodb-init will hang the script indefinitely but copying the exact same file and manually creating it in the ui works
            - icp-mongodb-install will create fine but when the mongo pod starts up it will complain about syntax error in the install script. Again, creating the exact same cm in the ui avoids this issue
            - there are around a dozen resources created for this script, only these two configmaps have any issue but they are both pretty long and house start up scripts
        - the cs-mongodump PVCs need their finalizer patched manually when the script runs .  The script patches it, the output acknowledges it, but it consistently hangs when deleting the PVC.
        - similarly, the PV attached to the cs-mongodump PVC needs it's `claimRef` removed. Otherwise the script hangs or requires manual intervention


There are also a couple changes included for splitting up the conversion script into smaller pieces like the pause.sh and zen-utils.sh. I have not tested them but open to feedback there as well.

multi-instance/isolate.sh
```diff
@@ -0,0 +1,385 @@
...
+            ${OC} delete nss odlm-scope-managedby-odlm -n ${namespace} || (error "unable to delete namespace scope odlm-scope-managedby-odlm in ${namespace}")
+        done
+    else
+        info "Namespace Scope CR \"odlm-scope-managedby-odlm\" not present. Moving on..."
+    fi
+
+    success "Namespace Scope CRs cleaned up"
+}
+
+function cleanupZenService(){
```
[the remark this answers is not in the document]
bluzarraga: I guess if it is necessary for anything, it's the restart to make sure zen refreshes properly with a new cs instance. That's assuming zen does not get upgraded after isolating so probably not necessary.

bluzarraga:
Progress update:
- Backup/preload
    - consistently working
    - applies all necessary files for mongo without hiccup
    - new mongo install picks up the restored pvc/pv and can login to LDAP using restored users in new namespaces
    - Ready for review
- Conversion running multiple times
    - as mentioned in last update, conversion can be run multiple times now
- Conversion checking topology for existing deployments
    - This is working with one caveat: If a namespace is added to the original cs instance (and only the original) the common-service nss in the original namespace will update itself very quickly, likely more quickly than an individual could apply the change to the common-service-maps and then run the conversion script (it at least creates a race condition).
    - This is a limitation but one that only applies when adding a cloud pak namespace back to the original shared cs instance which we may not support anyway.
    - Otherwise, this is ready for review
- Isolate
    - removed zen cleanup, will look into other zen suggestions

multi-instance/isolate.sh
```diff
@@ -0,0 +1,527 @@
...
+}
+
+function construct_mapping() {
+    NEW_MAPPING='- requested-from-namespace:'
+
+    unique_ns_list=()
+    # Loop over each tenant namespace and add each unique namespace value to the 'unique' array
+    local namespaces="$OPERATOR_NS $SERVICES_NS $TETHERED_NS"
+
+    for ns in $namespaces; do
```
pgodowski, line 163: why just not using `sort` and `uniq`?

multi-instance/isolate.sh
```diff
@@ -0,0 +1,527 @@
...
+            ${OC} get sub ${sub} -n ${namespace} -o yaml > tmp.yaml 
+            ${YQ} -i '.spec.source = "'${catalog_source}'"' tmp.yaml || error "Could not replace catalog source for CS operator in namespace ${namespace}"
+            ${OC} apply -f tmp.yaml
+            info "Common Service Operator Subscription in namespace ${namespace} updated to use catalog source ${catalog_source}"
+        else
+            info "No Common Service Operator in namespace ${namespace}. Moving on..."
+        fi
+        return_value=""
+    done
+    rm tmp.yaml -f
```
pgodowski, line 406:
```suggestion
    rm -f tmp.yaml
```

multi-instance/isolate.sh
```diff
@@ -0,0 +1,527 @@
...
+            if [[ $return_value == $cm ]]; then
+                ${OC} get cm -n $namespace $cm -o yaml --ignore-not-found > tmp.yaml
+                #edit the file to change the namespace to CONTROL_NS
+                yq -i '.metadata.namespace = "'${CONTROL_NS}'"' tmp.yaml
+                ${OC} apply -f tmp.yaml
+                info "Licensing configmap $cm copied from $namespace to $CONTROL_NS"
+            fi
+        fi
+    done
+    rm tmp.yaml -f
```
pgodowski, line 471:
```suggestion
    rm -f tmp.yaml
```

pgodowski (changes requested):

PR #1110, opened 2023-03-28
set ControlNs to cs-control when csmaps creating

YCShen1010:
issue: https://github.ibm.com/IBMPrivateCloud/roadmap/issues/57957
CP2 ODLM will automatically extend its permission to any namespace where OperandRequest exists if `controlNamespace:""` in csmaps
#### Issue reproduce
1. Install CP3.0 in `cloudpak-ns`, will see `controlNamespace` is empty `""` when `common-service-maps` is being created.
2. Install CP2.0 in default namespace `ibm-common-services`, and create a `OperandRequest` in `cloudpak-ns`, will see the CP2.0 ODLM  permission is extented to `cloudpak-ns`

- `ibm-common-services` namespace will be included in `namespace-scope` in `cloudpak-ns`
- those tombstone services will be installed in the CP3.0 ns
   ```
    kind: ConfigMap
    apiVersion: v1
    metadata:
      name: namespace-scope
      namespace: cloudpak-ns
    data:
      namespaces: 'ibm-common-services,cloudpak-ns'
   ```
#### Verify fix
1. Apply image `quay.io/yuchen_shen/cs_operator:control_ns`
2. Delete existent configmap `common-service-maps`, and restart cs operator pod in `cloudpak-ns`
3. New csmaps will be created with `controlNamespace: cs-control`

### DIFF
--- a/controllers/bootstrap/init.go
+++ b/controllers/bootstrap/init.go
@@ -690,6 +690,7 @@ func (b *Bootstrap) CreateCsMaps() error {
 
 	newnsMapping.RequestNs = append(newnsMapping.RequestNs, strings.Split(b.CSData.WatchNamespaces, ",")...)
 	newnsMapping.CsNs = b.CSData.ServicesNs
+	cmData.ControlNs = "cs-control"
 	cmData.NsMappingList = append(cmData.NsMappingList, newnsMapping)
 	commonServiceMap, error := utilyaml.Marshal(&cmData)
 	if error != nil {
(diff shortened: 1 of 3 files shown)


Daniel-Fan:
@YCShen1010 
Could we also update the [function `UpdateCsMaps`](https://github.com/IBM/ibm-common-service-operator/blob/4f873dc1b697c4b2713e1d516e03f28f372dd318/controllers/common/util.go#L539) to add controlNamespace into ConfigMap if it is not there

Daniel-Fan (approved):
/lgtm

PR #1113, opened 2023-03-29
BYOC support without label

qpdpQ:
for ticket: https://github.ibm.com/IBMPrivateCloud/roadmap/issues/57312

It will check if `cs-ca-certificate-secret` is already there during the installation. If we find this secret but do not find the certificate, we will skip deploying `cs-ca-cert`

### DIFF
--- a/controllers/bootstrap/init.go
+++ b/controllers/bootstrap/init.go
@@ -969,17 +969,25 @@ func (b *Bootstrap) IsBYOCert() (bool, error) {
 		}
 		return false, nil
 	}
-	certName := "cs-ca-certificate"
-	cert := &certmanagerv1.Certificate{}
-	certerr := b.Client.Get(context.TODO(), types.NamespacedName{Name: certName, Namespace: b.CSData.ServicesNs}, cert)
-	if certerr != nil {
-		if errors.IsNotFound(certerr) {
-			return true, nil
-		}
+
+	certList := &certmanagerv1.CertificateList{}
+	opts := []client.ListOption{
+		client.InNamespace(b.CSData.ServicesNs),
+		client.MatchingLabels(
+			map[string]string{"app.kubernetes.io/instance": "cs-ca-certificate"}),
+	}
+	if certerr := b.Reader.List(ctx, certList, opts...); err != nil {
 		return false, certerr
 	}
 
-	return false, nil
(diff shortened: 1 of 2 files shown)


controllers/bootstrap/init.go
```diff
@@ -953,7 +955,33 @@ func CheckClusterType(mgr manager.Manager, ns string) (bool, error) {
...
+func (b *Bootstrap) IsBYOCert() (bool, error) {
+	klog.V(2).Info("Detect if it is BYO cert")
+	secretName := "cs-ca-ceritifcate-secret"
+	secret := &corev1.Secret{}
+	err := b.Client.Get(context.TODO(), types.NamespacedName{Name: secretName, Namespace: b.CSData.ServicesNs}, secret)
+	if err != nil {
+		if !errors.IsNotFound(err) {
+			return false, err
+		}
+	}
```
bitscuit, line 970: What happens if error is `IsNotFound`? Secret will be empty, but there is no code to check if secret is empty or not before checking for certificate object.

controllers/bootstrap/init.go
```diff
@@ -953,7 +955,33 @@ func CheckClusterType(mgr manager.Manager, ns string) (bool, error) {
...
+	certName := "cs-ca-certificate"
+	cert := &certmanagerv1.Certificate{}
+	certerr := b.Client.Get(context.TODO(), types.NamespacedName{Name: certName, Namespace: b.CSData.ServicesNs}, cert)
+	if certerr != nil {
+		if errors.IsNotFound(certerr) {
+			return true, nil
+		}
```
bitscuit, line 977: This is assuming that secret was found. But what if getting the secret returned `IsNotFound` error? This error is being ignored right now.

bitscuit (changes requested):

controllers/bootstrap/init.go
qpdpQ: Ah I didn't notice this, thanks for pointing it out

controllers/bootstrap/init.go
qpdpQ: Yes, if we didn't get this secret, we should return false directly.

bitscuit (approved):
/lgtm

PR #1125, opened 2023-04-05
update licensing operator pkg name

YCShen1010:


Daniel-Fan (approved):
/lgtm

closms:
Just an FYI. The cpd-scheduler reports gpgpu usage numbers to the license service. the installation of cpd-scheduler will add the scheduler namespace to the license services operator group  in the .spec.targetNamespaces.

I want to suggest that a comment be added to the code that creates the operatorgroup, that if the name `ibm-licensing-operator-app` shouldn't change without notifying the cpd-scheduler team.

Thanks

PR #1127, opened 2023-04-05
share ibm-cpp-config to all WATCH_NAMESPACE

Daniel-Fan:
`ibm-cpp-config` will be copied from `servicesNamespace` to all other namespace in the `WATCH_NAMESPACE` list.

### How to test
1. Deploy `cd` CatalogSource
2. Install cert-manager cluster wide
3. Using `setup_tenant.sh` to install CS in advanced topology including different operator and service namespace, as well as other tethered(CloudPak) namespace
4. Replace CS operator image by `quay.io/daniel_fan/common-service-operator-amd64:dev`
5. We will observe the `ibm-cpp-config` showing up in all namespaces in the tenant.
6. If we manually change the content in `ibm-cpp-config` ConfigMap, it will propagate the changes to other namespaces in 10 min.

### DIFF
--- a/controllers/bootstrap/init.go
+++ b/controllers/bootstrap/init.go
@@ -20,6 +20,7 @@ import (
 	"bytes"
 	"context"
 	"fmt"
+	"reflect"
 	"strconv"
 	"strings"
 	"text/template"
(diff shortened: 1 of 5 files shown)


YCShen1010 (approved):
/lgtm

PR #1130, opened 2023-04-06
Create backup_preload_mongo.sh

bluzarraga:
Create new script for backup and restore of mongo database into a new namespace. The difference between this script and the existing backup_restore_mongo.sh is when the restore takes place. In the original, the restore takes place after the new mongo and IAM instances are ready. This preload script is run before the new mongo is deployed, populating the pvc/pv the new mongo will be using before it is deployed. This avoids some of the problems the backup_restore_mongo.sh experiences, especially on ROKS.

Sequence of events described here https://github.ibm.com/IBMPrivateCloud/roadmap/blob/master/feature-specs/CloudPak3.0/UpgradeSequence.md under `Prepare a MongoDB backup` bullet.

Testing:
- setup environment with multiple cloud paks sharing one cs instance
- create and populate LDAP, verify login to original cs instance
- create target namespace for new cs instance
- run `./backup_preload_mongo.sh <original cs namespace> <target namespace>`
- verify script completes and there is a stand alone `mongodbdir-icp-mongodb-0` in the target namespace
- create csmaps cm with target namespace as a `map-to-namespace` value
- run conversion script
- wait for IAM to come ready
- login to cp-console for new cs instance with an LDAP user created earlier

In my testing this is in a pretty good spot but there are some tweaks left to be made:
- address pod security warnings when creating dummy mongo statefulset
- cs-mongodump pvc can be in "Lost" state in target namespace
    - in my limited experience with this issue, it has not affected the data being transferred

Part of https://github.com/IBM/ibm-common-service-operator/pull/1099 moved to its own PR for clarity.

### DIFF
--- a/backup_preload_mongo.sh
+++ b/backup_preload_mongo.sh
@@ -1,0 +1,1671 @@
+#!/usr/bin/env bash
+#
+# Copyright 2023 IBM Corporation
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+# http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+
+#set -o errexit
+set -o pipefail
+set -o errtrace
+#set -o nounset
+
+OC=${3:-oc}
+
+NUM=$#
+TEMPFILE="_TMP.yaml"
+
+#
+# main - main logic
+#
+function main() {
+  parse $*
+  cleanup
+  deploymongocopy
+  createdumppvc
+  dumpmongo
+  swapmongopvc
+  loadmongo
+  deletemongocopy
+} # main
+
+
+#
+# Help function - print out when error are found
+#
+function help() {
+  echo "SYNTAX: copymongo.sh from-namespace to-namespace"
+  echo "Where:"
+  echo "  from-namespace: is the namespace in which mongodb is running and FROM which database content should be copied"
+  echo "  to-namespace:   is the namespace in which mongodb is running and TO which database content should be copied"
+} # help
+  
+
+#
+# Parse and validate the namespaces
+#
+function parse() {
+  info "Checking parameters and namespaces..."
+  if [ $NUM -ne 2 ]; then
+    help
+    exit -1
+  fi
+  FROM_NAMESPACE=$1
+  TO_NAMESPACE=$2
+
+  info "Copying mongodb from namespace $FROM_NAMESPACE to namespace $TO_NAMESPACE"
+
+  exists=$(oc get ns $FROM_NAMESPACE --no-headers --ignore-not-found)
+  if [[ -z "$exists" ]]; then
+    error "Namespace $FROM_NAMESPACE does not exist (or oc command line is not logged in)"
+    exit -1
+  fi 
+  runningmongo=$(oc get po icp-mongodb-0 --no-headers --ignore-not-found -n $FROM_NAMESPACE | awk '{print $3}')
+  if [[ -z "$runningmongo" ]] || [[ "$runningmongo" != "Running" ]]; then
+    error "Mongodb is not running in Namespace $FROM_NAMESPACE"
+    exit -1
+  fi
+
+  exists=$(oc get ns $TO_NAMESPACE --no-headers --ignore-not-found)
+  if [[ -z "$exists" ]]; then
+    error "Namespace $TO_NAMESPACE does not exist (or oc command line is not logged in)"
+    exit -1
+  fi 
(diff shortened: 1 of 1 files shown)


bluzarraga:
I removed the changes for the security warnings as they broke the mongo statefulset's deployment. I will need to investigate why these warnings are spit out and why mongo can't have these security fields defined.

backup_preload_mongo.sh
```diff
@@ -0,0 +1,1675 @@
...
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+
+#set -o errexit
+set -o pipefail
+set -o errtrace
+#set -o nounset
+
+OC=${3:-oc}
```
bitscuit, line 23: Should just remove this since it's not being used nor parsed.

bitscuit (approved):
/lgtm

backup_preload_mongo.sh
```diff
@@ -0,0 +1,1673 @@
...
+            - name: ca
+              mountPath: /ca-readonly
+            - name: mongodbdir
+              mountPath: /data/db
+              subPath: datadir
+            - name: tmp-mongodb
+              mountPath: /tmp
+          terminationMessagePolicy: File
+          image: >-
+            icr.io/cpopen/cpfs/ibm-mongodb-install@sha256:bb236428cd36f3937d268c4475a1c62ac1a4e7cb9ca0f3de482f08817378b003
```
pgodowski, line 1256: It should be the same image as in the CP2 namespace - otherwise it might not be mirrored

backup_preload_mongo.sh
```diff
@@ -0,0 +1,1673 @@
...
+            - name: init
+              mountPath: /init
+            - name: mongodbdir
+              mountPath: /data/db
+              subPath: datadir
+            - name: tmp-mongodb
+              mountPath: /tmp
+          terminationMessagePolicy: File
+          image: >-
+            icr.io/cpopen/cpfs/ibm-mongodb@sha256:3a44fcf5656cdd3f062d3ca45d7fd0a46ff3ed90f6ed34ba260ad50938e95f57
```
pgodowski, line 1324: It must be the same image as in CP2 namespace

backup_preload_mongo.sh
```diff
@@ -0,0 +1,1673 @@
...
+            - name: configdir
+              mountPath: /data/configdb
+            - name: mongodbdir
+              mountPath: /work-dir
+              subPath: workdir
+            - name: tmp-mongodb
+              mountPath: /tmp
+          terminationMessagePolicy: File
+          image: >-
+            icr.io/cpopen/cpfs/ibm-mongodb@sha256:3a44fcf5656cdd3f062d3ca45d7fd0a46ff3ed90f6ed34ba260ad50938e95f57
```
pgodowski, line 1419: It must be the same image as in CP2

backup_preload_mongo.sh
```diff
@@ -0,0 +1,1673 @@
...
+            - name: configdir
+              mountPath: /data/configdb
+            - name: mongodbdir
+              mountPath: /work-dir
+              subPath: workdir
+            - name: tmp-metrics
+              mountPath: /tmp
+          terminationMessagePolicy: File
+          image: >-
+            icr.io/cpopen/cpfs/ibm-mongodb-exporter@sha256:f6456dc4e473295648c779cdabe1e4a40b660d69c9190fb2d5dfe7e94656ef17
```
pgodowski, line 1497: It must be the same image as in CP2

backup_preload_mongo.sh
```diff
@@ -0,0 +1,1673 @@
...
+  oc apply -f $TEMPFILE
+
+  status=$(oc get pvc cs-mongodump --no-headers | awk '{print $2}')
+  while [[ "$status" != "Bound" ]]
+  do
+    info "Waiting for pvc cs-mongodump to bind"
+    sleep 10
+    status=$(oc get pvc cs-mongodump --no-headers | awk '{print $2}')
+  done
+  success "MongoDB backup ready"
```
pgodowski, line 170: Backup is not ready - what is ready is the PVC

backup_preload_mongo.sh
```diff
@@ -0,0 +1,1673 @@
...
+  name: mongodb-backup
+spec:
+  parallelism: 1
+  completions: 1
+  backoffLimit: 20
+  template:
+    spec:
+      containers:
+      - name: cs-mongodb-backup
+        image: quay.io/opencloudio/ibm-mongodb:4.0.24
```
pgodowski, line 199: We need to use the image which matches CP2 image

backup_preload_mongo.sh
```diff
@@ -0,0 +1,1673 @@
...
+  name: mongodb-restore
+spec:
+  parallelism: 1
+  completions: 1
+  backoffLimit: 20
+  template:
+    spec:
+      containers:
+      - name: icp-mongodb-restore
+        image: quay.io/opencloudio/ibm-mongodb:4.0.24
```
pgodowski, line 369: It has to be the same image as in CP2

backup_preload_mongo.sh
bluzarraga: This image is for the mongo backup job, would it be mirrored?

backup_preload_mongo.sh
bluzarraga: same as backup job image, would there be somewhere to grab this from since it is a restore job?

PR #1131, opened 2023-04-06
run conversion multiple times, check topology

bluzarraga:
Part of https://github.com/IBM/ibm-common-service-operator/pull/1099 moved to its own PR for clarity.

Right now the conversion script is only capable of being run once on a cluster. There are a few reasons why this could be a problem:

    error in setting up the cluster (ie typo in common-service-maps)
    clusters changing needs (ie adding or removing namespaces to a given cs instance)

In addition to running the script multiple times, the conversion script will also need to change slightly to fit the pattern Mike displayed for the upgrade process. The conversion script already does most of what he drew up but we need to be able to break it down by requested-from-namespace/map-to-namespace groupings (aka isolate).

I propose checking the topology first, getting the groupings that need to be updated, then running the rest of the script. At the moment, I lump in all the namespaces that need to be updated together (activeRequestedFrom & activeMapTo) but we could break it down even further and individually call all of the functions required with the specific subset of namespaces that need to be updated from the check_topology function. Breaking it down like that would let us be more specific in what we update and when. It could also let us incorporate backup/restore more seamlessly.

Testing:
- usual conversion setup for the environment, might want at least 3 cp instances (or just namespaces) in csmaps
- run conversion once moving 2/3 cp namespaces to a new dedicated instance
- edit csmaps to move one of the 2 cp namespaces you just moved to yet another new dedicated instance
- run conversion a second time 
- verify the original namespace is not touched during runtime
- can check the nss common-service in a cs instance, it should only include namespaces that line up with the csmaps

Note:
- if you re-add namespaces to the original cs instance, the common-service nss in the original namespace will be updated immediately and the script will think the conversion has already happened. Looking into whether we should support adding namespaces back to the original. In case of error (ie a namespace was left out by accident), it should not make a difference if it is added back after the fact.
- In my own testing this has been working well. However, I found an issue with the comparison logic in the topology check when trying to use similar logic in the isolate script. I am looking into changing this but as of this writing it is not ready yet.

### DIFF
--- a/convert_to_multi_instance.sh
+++ b/convert_to_multi_instance.sh
@@ -20,20 +20,21 @@ set -o pipefail
 set -o errtrace
 set -o nounset
 
-OC=${1:-oc}
-YQ=${1:-yq}
+OC=oc
+YQ=yq
 
 cs_operator_channel=
 catalog_source=
 requestedNS=
 mapToCSNS=
-
+master_ns=$1
 cm_name="common-service-maps"
     
 function main() {
     msg "Conversion Script Version v1.0.0"
     prereq
     collect_data
+    check_topology
     prepare_cluster
     scale_up_pod
     restart_CS_pods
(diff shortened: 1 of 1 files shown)


bitscuit (approved):
/lgtm

PR #1132, opened 2023-04-06
disable webhook in cncf

qpdpQ:
for ticket: https://github.ibm.com/IBMPrivateCloud/roadmap/issues/58081
if it is not ocp, we skip deploying webhook server and skip reconciling webhook


Daniel-Fan (approved):
/lgtm

PR #1134, opened 2023-04-06
ensure isolated mode value is true in ODLM subscription

YCShen1010:
issue: https://github.ibm.com/IBMPrivateCloud/roadmap/issues/57938
#### Function:

- Get ODLM subscriptions based on given namespace after conversion
- Check and set ISOLATED_MODE value to true, support multiple runs
- Apply the updated subscription back to the cluster

Checked the ODLM `pod` and `deployment` yaml, they all have corresponding updates.

### DIFF
--- a/isolate.sh
+++ b/isolate.sh
@@ -521,6 +521,54 @@ function isolate_odlm() {
         error "Failed to update subscription ${package_name} in ${ns}"
     fi
     rm sub.yaml
+
+    check_odlm_env "${ns}" 
+}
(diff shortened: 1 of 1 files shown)


Daniel-Fan:
@YCShen1010 

Could we add an additional check and waiting for the ENV update in ODLM deployment?

Daniel-Fan (approved):

bluzarraga (approved):

PR #1138, opened 2023-04-11
delete original Licensing CMs after moved to controlNs

YCShen1010:
bug issue: https://github.ibm.com/IBMPrivateCloud/roadmap/issues/57878

### DIFF
--- a/convert_to_multi_instance.sh
+++ b/convert_to_multi_instance.sh
@@ -165,6 +165,8 @@ function migrate_lic_cms() {
                 yq -i '.metadata.namespace = "'${controlNs}'"' tmp.yaml
                 ${OC} apply -f tmp.yaml
                 info "Licensing configmap $cm copied from $namespace to $controlNs"
+                # delete the original in the cs namespace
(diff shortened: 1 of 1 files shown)


convert_to_multi_instance.sh
```diff
@@ -165,6 +165,8 @@ function migrate_lic_cms() {
                 yq -i '.metadata.namespace = "'${controlNs}'"' tmp.yaml
                 ${OC} apply -f tmp.yaml
```
Daniel-Fan, line 166: Could we check the return value of applying new ConfigMap?

I am worried that if it failed to apply new ConfigMap, and directly delete the old one, which leads to a data loss.

Daniel-Fan:
Could we update the `migrate_cp2_licensing.sh` https://github.com/IBM/ibm-common-service-operator/blob/f80edf5f49ef4428b6f6e754ba7bb2cfe449c1a4/cp3pt0-deployment/common/migrate_cp2_licensing.sh#L140

PR #1141, opened 2023-04-12
Pass in a correct source name for cert manager argument

YCShen1010:
fix for the issue: [57978](https://github.ibm.com/IBMPrivateCloud/roadmap/issues/57978#issuecomment-55822700)
When there is a value of `--cert-manager-source` passed in, but cert-manager subscription did not create with a expected source name.


Daniel-Fan (approved):
/lgtm

PR #1144, opened 2023-04-13
add checks for existing cs maps

bluzarraga:
- No longer delete csmaps
- check if `controlNamespace` defined
- check if existing `map-to` matches parameter map-to
- remove logic that could add a second set of requested-from/map-to
- initial code for checking to make sure cs meets minimum version requirement (not quite working yet)


isolate.sh
```diff
@@ -441,6 +405,28 @@ function removeNSS(){
         info "Namespace Scope CR \"odlm-scope-managedby-odlm\" not present. Moving on..."
     fi
 
+    failcheck=$(${OC} get nss --all-namespaces | grep nss-odlm-scope || echo "failed")
```
bitscuit, line 408: We only need to check the original-cs-ns for NSS CRs to delete, not all namespaces, right?

bluzarraga: yeah, this code was grabbed from the conversion script so I'll need to change it there too

isolate.sh
```diff
@@ -383,46 +383,34 @@ function removeNSS(){
     title " Removing ODLM managed Namespace Scope CRs "
     msg "-----------------------------------------------------------------------"
 
-    failcheck=$(${OC} get nss --all-namespaces | grep nss-managedby-odlm || echo "failed")
+    failcheck=$(${OC} get nss ${master_ns} | grep nss-managedby-odlm || echo "failed")
```
bitscuit, line 386: can probably skip this check all together, and just run `${OC} delete nss nss-managedby-odlm --ignore-not-found -n ${master_ns} || (error "unable to delete namespace scope nss-managedby-odlm in ${master_ns}")` to bypass any not found errors

PR #1146, opened 2023-04-13
continue to run the in-place upgrade script even the sub was updated

YCShen1010:
for the issue: https://github.ibm.com/IBMPrivateCloud/roadmap/issues/57978#issuecomment-55770004
#### Verify the fix
Given the same channel and catalogsource info as the existing one, upgrade will continue
log:
```
[INFO] v4.0 is equal to v4.0
[INFO] catalogsource opencloud-operators is the same as opencloud-operators
[INFO] ibm-common-service-operator has already updated channel v4.0 and catalogsource opencloud-operators in the subscription.
subscription.operators.coreos.com/ibm-common-service-operator configured
```


Daniel-Fan (approved):
/lgtm

PR #1150, opened 2023-04-18
scale CS and ODLM deployment 

YCShen1010:
fix for issue: https://github.ibm.com/IBMPrivateCloud/roadmap/issues/58236

-  Check if channel and CatalogSource are in correct version in cs subscription and then scale down CS, ODLM deployment and delete old OprandRegistry
    - The upgrade involves ibm-cert-manager-operator migration, including deleting old cert-manager in CS namespace
    - `ibm-cert-manager-operator` is marked as tombstone in new v4.0 OperandRegistry
-  If the `installedCSV` in sub matches designated channel version, and then scale up CS 
-  After the new OperandRegistry is applied in the cluster by cs-operator, scale ODLM back to 1 as only new v4.0 ODLM should run in the cluster,  v3 ODLM does not recognize tombstone in OperandRegistry

### DIFF
--- a/cp3pt0-deployment/common/utils.sh
+++ b/cp3pt0-deployment/common/utils.sh
@@ -773,3 +773,84 @@ function delete_operator() {
         msg ""
     done
 }
+
+function scale_deployment() {
+    local ns=$1
+    local sub=$2
+    ${OC} scale deployment -n "${ns}" "${sub}" --replicas=$3
+}
+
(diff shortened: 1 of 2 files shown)


Daniel-Fan (changes requested):

Daniel-Fan (approved):
/lgtm

PR #1156, opened 2023-04-19
isolated script idempotency

bitscuit:


csmaps_utils.sh
```diff
@@ -0,0 +1,162 @@
...
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+
+set -o nounset
+
+OC=oc
+YQ=yq
+CM_NAME="mock-cs-maps"
```
bluzarraga, line 22: assuming this should be changed back to common-service-maps

csmaps_utils.sh
```diff
@@ -0,0 +1,162 @@
...
+# echo "List of namespaces in tenant1: "
+# read_tenant_from_csmaps "tenant1"
+
+# update_tenant "tenant1" "ns1 ns2"
+# echo "List of namespaces in tenant1: "
+# read_tenant_from_csmaps "tenant1"
+
+# update_tenant "tenant2" "ns3"
+# echo "List of namespaces in tenant2: "
+# read_tenant_from_csmaps "tenant2"
```
bitscuit, line 162: keeping this here for quick testing of the script

csmaps_utils.sh
bitscuit: I am leaning towards keeping it as `mock-cs-maps` since this file is primarily for testing and storing the functions to manipulate configmaps, not actually to be used as-is. If this script is included by another script, the CM_NAME can be overwritten by the importing script anyways

csmaps_utils.sh
```diff
@@ -0,0 +1,162 @@
...
+    update_cs_maps ""
+    success "Empty common-service-maps configmap created in kube-public namespace"
+}
+
+# insert_control_ns Insert the controlNamespace field into the configmap if it
+# does not exist
+function insert_control_ns() {
+    local current_yaml=$("${OC}" get -n kube-public cm "$CM_NAME" -o yaml | "${YQ}" '.data.["common-service-maps.yaml"]')
+
+    current=$(echo "$current_yaml" | "${YQ}" .'controlNamespace')
```
bluzarraga, line 67: does the period need to be inside the ''? '.controlNamespace' instead of .'controlNamespace'

csmaps_utils.sh
```diff
@@ -0,0 +1,162 @@
...
+    update_cs_maps "$padded_yaml"
+}
+
+# read_tenant_from_csmaps Gets the list in requested-from-namespace for a given
+# map_to_cs_ns and prints it out. If map_to_cs_ns does not exist, then output is
+# empty
+function read_tenant_from_csmaps() {
+    local map_to_cs_ns=$1
+    local current_yaml=$("${OC}" get -n kube-public cm "$CM_NAME" -o yaml | "${YQ}" '.data.["common-service-maps.yaml"]')
+    local tenant_ns_list=$(echo "$current_yaml" | "${YQ}" eval '.namespaceMapping[] | select(.map-to-common-service-namespace == "'${map_to_cs_ns}'").requested-from-namespace' | awk '{ print $2 }')
```
bluzarraga, line 83: you might need to truncate newline characters here otherwise the list is passed on like:  "ns1\nns2\nns3 ...etc". You can use `| tr '\n' ' '` after the awk command

bitscuit: `tenant_ns_list` does end up being newline delimited here. I don't think whether it's newline or space delimited impacts anything.

csmaps_utils.sh
bitscuit: good catch. Weird that it still worked though

bitscuit:
The git diff looks a bit wonky, but in summary:
1. I copied the csmaps_utils function to isolate.sh
1. optimized gather_csmaps_ns
1. removed mapping_topology function since the csmaps_utils functions can be used instead to construct the common-service-maps
1. merged abort_check into prereq
1. removed many global variables and localized any variables declared in functions
1. optimized licensing configmap migration

Contrary to what the diff shows, I did not make changes to removeNSS nor did I remove cleanupCSOperators. If you look at what is in the scripts branch and what is in this branch, you can see there is no difference.

PR #1158, opened 2023-04-20
remove webhook and secretshare alongside other singletons

bluzarraga:


isolate.sh
```diff
@@ -552,6 +556,33 @@ function wait_for_condition() {
     fi
 }
 
+function cleanup_deployment() {
+    local name=$1
+    local namespace=$2
+    info "Deleting existing Deployment ${name} in namespace ${namespace}..."
+    ${OC} delete deployment ${name} -n ${namespace} --ignore-not-found
+
+    wait_for_no_pod ${namespace} ${name}
```
Daniel-Fan, line 565: We might wanna remove this additional wait condition `wait_for_no_pod`, it is usually redundant.
If the deployment is removed, their pods should be gone as well

convert_to_multi_instance.sh
```diff
@@ -686,6 +689,33 @@ function removeNSS(){
     success "Namespace Scope CRs cleaned up"
 }
 
+function cleanup_deployment() {
+    local name=$1
+    local namespace=$2
+    info "Deleting existing Deployment ${name} in namespace ${namespace}..."
+    ${OC} delete deployment ${name} -n ${namespace} --ignore-not-found
+
+    wait_for_no_pod ${namespace} ${name}
```
Daniel-Fan, line 698: ^^^

isolate.sh
bluzarraga: I'm going to remove it

bluzarraga:
tested, should be ready to go

Daniel-Fan (approved):
/lgtm

PR #1173, opened 2023-04-24
fixed isolate.sh to properly exclude namespaces

bitscuit:
- excluding namespaces made the assumption that the provided namespace(s) were already within the scope of the tenant
- bad assumption because if the provided exclude-ns namespace(s) were not within scope, then isolate would actually add in those namespaces to the scope
- fixed by intentionally duplicating EXCLUDED_NS to the scope variable before passing it though the duplicate removal pipe, so that excluded-ns will always be removed from scope, whether they exist in namespace-scope CM or not


Daniel-Fan (approved):
/lgtm

PR #1175, opened 2023-04-25
update licensing ns for migration

Daniel-Fan:
To avoid error
```
./cp3pt0-deployment/migrate_tenant.sh --operator-namespace ibm-common-services
[✔] oc command available
[✔] yq command available
[✘] Licensing Migration could only be done in controlNamespace, please set parameter '-licensingNs cs-control'
```


YCShen1010 (approved):
/lgtm

PR #1177, opened 2023-04-25
retry the subscription patch operation for 5 times

YCShen1010:
fix for patch operation failed: https://github.ibm.com/IBMPrivateCloud/roadmap/issues/57978#issuecomment-56237562
#### enhancement
`update_operator` function attempts to patch the subscription with the desired changes, and if it fails due to a conflict, it retries the patch operation with a delay of 5 seconds for 5 times. If the maximum number of retries is reached and the patch operation still fails, an error message is displayed


Daniel-Fan (approved):
/lgtm

PR #1178, opened 2023-04-25
parse install mode command-line argument

YCShen1010:
Missing `INSTALL_MODE` argument in `parse_arguments` function, and add it back to the migration script


Daniel-Fan (approved):
/lgtm

PR #1180, opened 2023-04-26
Cleanup script

qpdpQ:
for ticket: https://github.ibm.com/IBMPrivateCloud/roadmap/issues/58319
This script is used to clean up cs-control namespace in post upgrade.


qpdpQ:
skip license check for now, because the CP3 license service will be installed in the control namespace

Daniel-Fan:
Could we add an additional check on IBM License service in control namespace?

- If Licensing Service is still at 3.x level, we would
    - Error msg to ask user to execute setup_singleton.sh to migrate Licensing Service to v4.x
    - And stop the cleaning because Licensing Service 3.x reply on NSS operator v3.x to work properly.
- If Licensing Service is at 4.x level, we are good to remove other components.

To check Licensing Service version check, we could refer to [`setup_singleton.sh` here](https://github.com/IBM/ibm-common-service-operator/blob/5bf38a89ac6c17a9e6637d5d7323133d3362aeec/cp3pt0-deployment/setup_singleton.sh#L196)

Daniel-Fan (approved):
/lgtm

PR #1182, opened 2023-04-26
accept the license for cert manager, cs op, nss, and licensing based on parameter

bluzarraga:
Added function to accept and update the license for cert manager in the setup_singletons script. Especially interested in feedback on line 173.

### DIFF
--- a/cp3pt0-deployment/common/utils.sh
+++ b/cp3pt0-deployment/common/utils.sh
@@ -865,8 +865,8 @@ function scale_up() {
 }
 
 function accept_license() {
-    kind=$1
-    namespace=$2
+    local kind=$1
+    local namespace=$2
     kind_exists=$(${OC} get "$kind" || echo "fail")
     if [[ $kind_exists != "fail" ]]; then
         cr=$(${OC} get "$kind" --no-headers -n "$namespace" | awk '{print $1}')
--- a/cp3pt0-deployment/common/utils.sh
+++ b/cp3pt0-deployment/common/utils.sh
@@ -867,10 +867,11 @@ function scale_up() {
 function accept_license() {
     local kind=$1
     local namespace=$2
-    kind_exists=$(${OC} get "$kind" || echo "fail")
+    kind_exists=$(${OC} get "$kind" -n "$namespace" || echo "fail")
     if [[ $kind_exists != "fail" ]]; then
         cr=$(${OC} get "$kind" --no-headers -n "$namespace" | awk '{print $1}')
         ${OC} patch "$kind" "$cr" -n "$namespace" --type='merge' -p '{"spec":{"license":{"accept":true}}}' || warning "Failed to update license acceptance for $kind CR $cr"
+        info "License accepted for $kind $cr."
     else
         warning "Resource kind $kind not found on cluster."
         return
--- a/cp3pt0-deployment/common/utils.sh
+++ b/cp3pt0-deployment/common/utils.sh
@@ -867,40 +867,12 @@ function scale_up() {
 function accept_license() {
     kind=$1
     namespace=$2
-    if [[ $kind == "certmanagerconfig.operator.ibm.com" ]] || [[ $kind == "ibmlicensing" ]]; then
-        kind_exists=$(${OC} get $kind || echo "fail")
-        if [[ $kind_exists != "fail" ]]; then
-            cr=$(${OC} get $kind | awk '{print $1}')
-            ${OC} patch $kind $cr --type='merge' -p '{"spec":{"license":{"accept":true}}}' || warning "Failed to update license acceptance for $kind CR $cr"
-        else
-            warning "Resource kind $kind not found on cluster."
-            return
-        fi
+    kind_exists=$(${OC} get $kind || echo "fail")
+    if [[ $kind_exists != "fail" ]]; then
+        cr=$(${OC} get $kind -n $namespace --no-headers | awk '{print $1}')
+        ${OC} patch $kind $cr --type='merge' -p '{"spec":{"license":{"accept":true}}}' || warning "Failed to update license acceptance for $kind CR $cr"
     else
-        if [[ $kind == "commonservice" ]]; then
-            kind_exists=$(${OC} get $kind -n $namespace || echo "fail")
-            if [[ $kind_exists != "fail" ]]; then
-                cr=$(${OC} get $kind -n $namespace | awk '{print $1}')
-                ${OC} patch $kind $cr -n $namespace --type='merge' -p '{"spec":{"license":{"accept":true}}}' || warning "Failed to update license acceptance for $kind CR $cr"
-            else
-                warning "$kind not found in namespace $namespace"
-                return
-            fi
-        elif [[ $kind == "namespacescope" ]]; then
-            kind_exists=$(${OC} get $kind -n $namespace || echo "fail")
-            if [[ $kind_exists != "fail" ]]; then
-                nss_crs=$(${OC} get $kind -n $namespace | awk '{print $1}' | tr '\n' ' ')
-                for cr in $nss_crs
-                do
-                    ${OC} patch $kind $cr -n $namespace --type='merge' -p '{"spec":{"license":{"accept":true}}}' || warning "Failed to update license acceptance for $kind CR $cr in namespace $namespace"
-                done
-            else
-                warning "$kind not found in namespace $namespace"
-                return
-            fi
-        else
-            warning "Kind $kind specified not recognized. Accepted kinds are commonservice, namespacescope, ibmlicensing, and certmanagerconfig.operator.ibm.com"
-            return  
-        fi
+        warning "Resource kind $kind not found on cluster."
+        return
     fi
 }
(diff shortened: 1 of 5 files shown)


bitscuit:
please also update setup_tenant to handle license acceptance after installation of NSS operator and cs-operator

cp3pt0-deployment/setup_singleton.sh
```diff
@@ -227,4 +235,16 @@ function get_and_validate_arguments() {
     get_control_namespace
 }
 
+function accept_license() {
+    if [[ $LICENSE_ACCEPT == "true" ]]; then
+        certmanager_cr=$(${OC} get certmanager --no-headers | awk '{print $1}')
```
Daniel-Fan, line 240: I think it would be new API for cert-manager
```
oc get certmanagerconfig.operator.ibm.com
```

cp3pt0-deployment/setup_singleton.sh
```diff
@@ -227,4 +235,16 @@ function get_and_validate_arguments() {
     get_control_namespace
 }
 
+function accept_license() {
+    if [[ $LICENSE_ACCEPT == "true" ]]; then
```
bitscuit, line 239:
```suggestion
    if [ $LICENSE_ACCEPT -eq 1 ]; then
```

cp3pt0-deployment/setup_singleton.sh
```diff
@@ -227,4 +235,16 @@ function get_and_validate_arguments() {
     get_control_namespace
 }
 
+function accept_license() {
```
bitscuit, line 238: Need to generalize the function because licensing-operator will also have license acceptance, so this will need to be called whenever they finish implementing license acceptance.

Suggest function to look like:
```
accept_license:
  arg1 kind - e.g. certmanager or ibmlicensing
  cr := oc get $kind
  if cr is empty then warning msg and return
  oc patch cr with .spec.license.accept: true
  if error then warning msg
```

cp3pt0-deployment/setup_singleton.sh
bluzarraga: ok I'll update that

cp3pt0-deployment/setup_singleton.sh
```diff
@@ -163,6 +170,7 @@ function install_cert_manager() {
     create_operator_group "ibm-cert-manager-operator" "${CERT_MANAGER_NAMESPACE}" "{}"
     create_subscription "ibm-cert-manager-operator" "${CERT_MANAGER_NAMESPACE}" "$CHANNEL" "ibm-cert-manager-operator" "${CERT_MANAGER_SOURCE}" "${SOURCE_NS}" "${INSTALL_MODE}"
     wait_for_operator "${CERT_MANAGER_NAMESPACE}" "ibm-cert-manager-operator"
+    accept_license #may need to rethink this, if license is false, wait for operator could fail, if we try to set it before waiting, the cert manager cr may not exist yet
```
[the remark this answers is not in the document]
bluzarraga: cool, I'll remove the comment then

bitscuit (changes requested):

cp3pt0-deployment/setup_singleton.sh
bitscuit, line 239:
```suggestion
    if [ $LICENSE_ACCEPT -ne 1 ]; then
        return
    fi
```

invert it so that function returns early if license was not accepted

bitscuit: move this check to prereq and error out of script

cp3pt0-deployment/common/utils.sh
```diff
@@ -862,4 +862,17 @@ function scale_up() {
         msg "Scaling up operand-deployment-lifecycle-manager deployment in ${operator_ns} namespace back to 1"
         scale_deployment ${operator_ns} operand-deployment-lifecycle-manager 1
     fi
+}
+
+function accept_license() {
+    kind=$1
+    namespace=$2
```
bitscuit, line 869:
```suggestion
    local kind=$1
    local namespace=$2
```
localize new variables in functions

bitscuit: suggest also add a third variable `local cr_name=$3`, since we already know the names of CRs to check from installation and upgrade perspective. No point in running an unnecessary get. 

Just try to patch the CR, and if it fails, then return error, so that wrappers can handle and re-run script if needed

bluzarraga:
settup_singleton and migrate_tenant both tested successfully. The license is being accepted as expected when using these scripts

cp3pt0-deployment/setup_singleton.sh
```diff
@@ -194,6 +200,7 @@ EOF
     create_operator_group "ibm-licensing-operator-app" "${LICENSING_NAMESPACE}" "$target"
     create_subscription "ibm-licensing-operator-app" "${LICENSING_NAMESPACE}" "$CHANNEL" "ibm-licensing-operator-app" "${LICENSING_SOURCE}" "${SOURCE_NS}" "${INSTALL_MODE}"
     wait_for_operator "${LICENSING_NAMESPACE}" "ibm-licensing-operator"
+    # accept_license "ibmlicensing" ""
```
bitscuit, line 203: should be able to uncomment this, doesn't matter if licensing hasn't implemented it yet, we can still patch the CR with it.

cp3pt0-deployment/setup_tenant.sh
```diff
@@ -202,6 +211,7 @@ function install_nss() {
     fi
 
     wait_for_operator "$OPERATOR_NS" "ibm-namespace-scope-operator"
+    accept_license "namespacescope" "$OPERATOR_NS"
```
bitscuit, line 214: need to move this to after configure_nss_kind because the NSS CR isn't created until then

cp3pt0-deployment/setup_tenant.sh
```diff
@@ -139,6 +144,10 @@ function pre_req() {
         error "Cert-manager is not found or having more than one\n"
     fi
 
+    if [[ $LICENSE_ACCEPT != 1 ]]; then
```
bitscuit, line 147:
```suggestion
    if [ "$LICENSE_ACCEPT" -ne 1 ]; then
```
simpler form of checking

cp3pt0-deployment/setup_singleton.sh
```diff
@@ -207,6 +214,10 @@ function pre_req() {
         success "oc command logged in as ${user}"
     fi
 
+    if [[ $LICENSE_ACCEPT != 1 ]]; then
```
bitscuit, line 217:
```suggestion
    if [ "$LICENSE_ACCEPT" -ne 1 ]; then
```

cp3pt0-deployment/migrate_tenant.sh
```diff
@@ -210,6 +216,10 @@ function pre_req() {
         success "oc command logged in as ${user}"
     fi
 
+    if [[ $LICENSE_ACCEPT != 1 ]]; then
```
bitscuit, line 219:
```suggestion
    if [ "$LICENSE_ACCEPT" -ne 1 ]; then
```

cp3pt0-deployment/common/utils.sh
```diff
@@ -862,4 +862,18 @@ function scale_up() {
         msg "Scaling up operand-deployment-lifecycle-manager deployment in ${operator_ns} namespace back to 1"
         scale_deployment ${operator_ns} operand-deployment-lifecycle-manager 1
     fi
+}
+
+function accept_license() {
+    local kind=$1
+    local namespace=$2
```
bitscuit, line 869:
```suggestion
    local namespace=$2
    local cr_name=$3
```

Suggest adding new cr_name variable since we know all the names of the CRs when running the scripts. Then the function can just patch the CR unconditionally, no need to run a get anymore.

Cert-manager creates CR named `default`
cs-operator creates CR named `common-service`
licensing-operator creates CR named `instance`
script creates NSS CR named `common-service`

cp3pt0-deployment/migrate_tenant.sh
```diff
@@ -210,6 +216,10 @@ function pre_req() {
...
+        error "License not accepted. Rerun script with --license-accept flag set. See https://ibm.biz/integration-licenses for more details"
```
bitscuit, line 220:
```suggestion
        warning "License not accepted. Rerun script with --license-accept flag set. See https://ibm.biz/integration-licenses for more details"
```

Let's just use warning since license-accept does not break operator functionality. Same for other files

bitscuit (changes requested):

cp3pt0-deployment/common/utils.sh
```diff
@@ -917,3 +917,16 @@ function scale_up() {
     fi
 }
 
+function accept_license() {
+    local kind=$1
+    local namespace=$2
+    local cr_name=$3
+    kind_exists=$(${OC} get "$kind" -n "$namespace" || echo "fail")
+    if [[ $kind_exists != "fail" ]]; then
```
bitscuit, line 925:
```suggestion
```
No point checking, just patch

bitscuit (changes requested):

bitscuit (approved):
/lgtm

PR #1184, opened 2023-04-26
Revert "cleanup based on feedback"

bluzarraga:
Reverts IBM/ibm-common-service-operator#1151

This change caused preload to fail with an empty storageClassName in the cs-mongodumps PVC. It was supposed to be a fix for ROKS environments but it is affecting our other test environments. We will need to investigate this further before reimplmenting


bitscuit (approved):
/lgtm

PR #1186, opened 2023-04-27
preload config initial commit

bluzarraga:
change the backup_preload_mongo.sh to a symlink to the new preload_custom_config.sh script. This new script includes all of the logic from backup_preload_mongo as well as an additional function to copy the auth idp secret from the target namespace.

I am setting up an environment to test it but right now the only difference in calling the script is using `--original-cs-ns` and `--target-ns` to pass through namespace parameters instead of just passing in two strings. If this should be changed to be the exact same as the backup_preload_mongo.sh for the short term I can do that.


bitscuit:
Suggest change the script name to `preload_data.sh`

bluzarraga:
CPD confirmed they clone the entire branch, symlink should be effective

bluzarraga:
had a successful test running this yesterday, I do not think the text updates from this morning will change that

bluzarraga:
spoke too soon, working out a bug

preload_data.sh
```diff
@@ -0,0 +1,1735 @@
...
+
+# set -o errexit
+set -o pipefail
+set -o errtrace
+# set -o nounset
+
+OC=oc
+YQ=yq
+FROM_NAMESPACE=$1
+TO_NAMESPACE=$2
```
bitscuit, line 26:
```suggestion
FROM_NAMESPACE=""
TO_NAMESPACE=""
```

Cannot just set it to positional args $1 and $2 because there are other optional flags, so if executed like this `./preload_data.sh --oc oc/path ns1 ns2`, then FROM_NAMESPACE will actually equal `--oc` and TO_NAMESPACE will actually equal `oc/path`, which is not correct

bitscuit (changes requested):

bitscuit (approved):
/lgtm

PR #1187, opened 2023-04-28
tweaks based on CPD feedback

bluzarraga:
Remove q in favor of yq since this was the only occurrence of jq in the script and we don't check to make sure it is installed. I tested the command manually and the output is the exact same.


Daniel-Fan (approved):
/lgtm

PR #1188, opened 2023-04-28
move uninstall singletons to after csmaps created, check for cert manager pods to be running

bluzarraga:


isolate.sh
```diff
@@ -73,10 +73,10 @@ function main() {
     prereq
     local ns_list=$(gather_csmaps_ns)
     pause
```
bitscuit, line 75: the pause function calls removeNSS, but we should not be deleting NSS CRs so early. The removeNSS should be run after cs-maps handling

isolate.sh
```diff
@@ -216,12 +216,12 @@ function update_tenant() {
...
     if [ -z "$isExists" ]; then
         info "The provided map-to-common-service-namespace: $map_to_cs_ns, does not exist in common-service-maps"
-        info "Adding new map-to-commn-service-namespace"
+        info "Adding new map-to-common-service-namespace"
         updated_yaml=$(echo "$current_yaml" | "${YQ}" eval 'with(.namespaceMapping; . += [{"map-to-common-service-namespace": "'$map_to_cs_ns'"}])')
     fi
 
     local tmp="\"$map_to_cs_ns\","
-
+    # echo "map $map_to_cs_ns namespace $namespaces tmp $tmp"
```
bitscuit, line 224: suggest add a debug function like in setup_tenant to handle debug logs

isolate.sh
```diff
@@ -237,7 +237,7 @@ function update_tenant() {
 # and namesapces from arguments, to output a unique sorted list of namespaces
 # with excluded namespaces removed
 function gather_csmaps_ns() {
-    local ns_scope=$("${OC}" get cm -n "$MASTER_NS" namespace-scope -o json | jq -r '.data.namespaces')
+    local ns_scope=$("${OC}" get cm -n "$MASTER_NS" namespace-scope -o yaml | yq -r '.data.namespaces')
```
bitscuit, line 240:
```suggestion
    local ns_scope=$("${OC}" get cm -n "$MASTER_NS" namespace-scope -o yaml | yq '.data.namespaces')
```

bitscuit (changes requested):

bluzarraga:
will also be including changes for 58487, adding do not merge label

isolate.sh
```diff
@@ -526,6 +539,89 @@ function cleanup_webhook() {
...
+    local name=$(${OC} get pod -n $namespace | (grep ibm-cert-manager-operator || echo "fail") | awk '{print $1}')
+    while [[ $name == "fail" ]] && [ $check_retries -gt 0 ]; do
+        sleep 15
+        name=$(${OC} get pod -n $namespace | (grep ibm-cert-manager-operator || echo "fail") | awk '{print $1}')
+        debug1 "cert manager operator pod present: $name"
+        check_retries=$(( check_retries - 1 ))
+        if [ $check_retries -eq 0 ]; then
+            error "Cert manager operator pod not found in namespace $namespace"
+        fi
+    done
```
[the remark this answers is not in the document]
bluzarraga: The name var is not set yet so if we do not wait for that, the condition gets sent through with an empty name variable. We have to wait for the cert manager pods to start before we can grep them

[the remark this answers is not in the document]
bluzarraga: I agree it seems unnecessary but it's either wait for a successful grep on the pod name or set an arbitrary sleep time to wait for cs operator to create the new cert manager deployment in control namespace and hope the following grep to set the name var works

isolate.sh
```diff
@@ -28,6 +28,8 @@ ADDITIONAL_NS=""
 CONTROL_NS=""
 CS_MAPPING_YAML=""
 CM_NAME="common-service-maps"
+cert_manager_migrated="false"
```
Daniel-Fan, line 31: Could we use all capitals for global variables.

isolate.sh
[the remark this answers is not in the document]
bitscuit: why does name have to be exact? The wait_for_condition does a grep anyways, so wouldn't hard coding the variable to `ibm-cert-manager-operator` be good enough?

isolate.sh
```diff
@@ -526,6 +539,89 @@ function cleanup_webhook() {
...
+        check_retries=$(( check_retries - 1 ))
+        if [ $check_retries -eq 0 ]; then
+            error "Cert manager cainjector pod not found in namespace $namespace"
+        fi
+    done
+    condition="oc -n ${namespace} get po --no-headers --ignore-not-found | egrep 'Running|Completed|Succeeded' | grep ^${name}"
+    wait_message="Waiting for pod ${name} in namespace ${namespace} to be running ..."
+    success_message="Pod ${name} in namespace ${namespace} is running."
+    error_message="Timeout after ${total_time_mins} minutes waiting for pod ${name} in namespace ${namespace} to be running."
+    wait_for_condition "${condition}" ${retries} ${sleep_time} "${wait_message}" "${success_message}" "${error_message}"
```
Daniel-Fan, line 621: I think if we already did cert-manager delegation at CP 2.0 level, It is possible that community or Red Hat is already installed in the cluster. As a result, we will not see any cert-manager operand pods in either `ibm-common-services` or control namespace.

IMHO I feel that cert-manager-operator pod check would be sufficient. But looking for other's opinion. FYI @bitscuit

Daniel-Fan (changes requested):

isolate.sh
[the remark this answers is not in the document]
Daniel-Fan: I think it is about `set -o errexit`, cmd `grep` will return non-zero status code if its value is not found.

So when `grep` is failed to find cert-manager-operator pod, it will abort the script immediately.

In order to solve the issue, we could make sure that `grep` is always returning zero status code. Here is an example
https://github.com/IBM/ibm-common-service-operator/blob/56743003b3ae99ec076b70ef2ce2db8947a347d2/isolate.sh#LL471C5-L471C178

isolate.sh
bitscuit: True, we cannot assume all the cert-manager pods will be there because of global cert-manager. We can probably add in more complex checks like checking for cert-manager pods cluster wide, but I think just checking for cert-manager-operator pod is good enough for the immediate issue

isolate.sh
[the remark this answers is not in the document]
bitscuit: Yea, and that's being done in `check_odlm_env`, so may as well follow that and remove the first loop that checks for existence, and just go straight to checking condition.

Also, `oc` is not using the variable form

isolate.sh
bluzarraga: We would only be waiting for cert manager if we remove the cert manager CR from the MASTER_NS. If cert manager is installed any other way, CERT_MANAGER_MIGRATED="false" and we would not call the wait_for_certmanager function

isolate.sh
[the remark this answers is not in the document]
bluzarraga: check_odlm_env is looking for the deployment, I guess we could probably check the deployments for cert manager as well instead of pods. That would circumvent the name var unset problem.

isolate.sh
bluzarraga: As far as what we should check, from the slack thread, cpd is specifically looking for controller deployment to be ready so they may want more than just the operator pod ready. I think they want the script to fully wait for cert manager to deploy since they run this in sequence with other steps

bluzarraga:
Incorporated the changes we discussed above. Just ran a test run and they seem to be working as expected. I'm going to remove the hold label.

isolate.sh
bitscuit: CertManager CR always exist in Bedrock, in both 2.0 and 3.0. And the cert-manager-operator pod also always exist. Whether the operand pods exist or not is dependent on whether the cluster has a global cert-manager or not.

So cannot always check for cert-manager operand pods. You can conditionally check them, but it would have to be generalized to checking cert-manager pods in all namespaces

isolate.sh
```diff
@@ -526,6 +540,50 @@ function cleanup_webhook() {
...
+    error_message="Timeout after ${total_time_mins} minutes waiting for deployment ${name} in namespace ${namespace} to be running."
+    wait_for_condition "${condition}" ${retries} ${sleep_time} "${wait_message}" "${success_message}" "${error_message}"
+    
+    #cainjector
+    name="cert-manager-cainjector"
+    condition="${OC} -n ${namespace} get deploy --no-headers --ignore-not-found | egrep '1/1' | grep ^${name} || true"
+    wait_message="Waiting for deployment ${name} in namespace ${namespace} to be running ..."
+    success_message="Deployment ${name} in namespace ${namespace} is running."
+    error_message="Timeout after ${total_time_mins} minutes waiting for deployment ${name} in namespace ${namespace} to be running."
+    wait_for_condition "${condition}" ${retries} ${sleep_time} "${wait_message}" "${success_message}" "${error_message}"
```
Daniel-Fan, line 582: Hi @bluzarraga 

Would we check the operand deployment/pod in cluster wide?
See the comment here https://github.com/IBM/ibm-common-service-operator/pull/1188#discussion_r1182012992

Daniel-Fan (approved):

bitscuit (approved):
/lgtm

PR #1193, opened 2023-05-01
BYOCA enhancement

qpdpQ:
fix an issue in the BYOCA certificate test.
We should only detect if it is BYOCA when we reconciling Master CR


Daniel-Fan:
After discussion with @qpdpQ , we will make some adjustments on this to reduce the number of function call `DeployCertManagerCR()`

Daniel-Fan (approved):
/lgtm

PR #1194, opened 2023-05-01
Patch CommonService CR before install CS operator

qpdpQ:
for ticket: https://github.ibm.com/IBMPrivateCloud/roadmap/issues/58489
If  CommonService CRD exists in the cluster, we patch CommonService CR before install CS operator
Otherwise, we patch CommonService CR after install CS operator


Daniel-Fan (approved):
/lgtm

PR #1198, opened 2023-05-02
updates for schedule

kgcarr:


velero/restore/restore.sh
```diff
@@ -0,0 +1,249 @@
...
+oc get pods -n $CS_NAMESPACE | grep icp-mongodb-0 | grep Running
+while [ $? -ne 0 ]
+do
+  echo "Waiting foundational services to be ready...Checking every minute"
+  oc get pods -n $CS_NAMESPACE
+  sleep 60
+  oc get pods -n $CS_NAMESPACE | grep icp-mongodb-0 | grep Running
+done
+echo "5 more minutes..."
+sleep 300
```
bluzarraga, line 212: A couple sleep statements that could be updated to look for something more specific to prevent possible race conditions.

bluzarraga:
Outside of the sleep statements I added a comment for, Kevin and I reviewed this and I think it looks good. Kevin will consider this on his next iteration of these scripts since they are currently not under pressure to be delivered. Going to approve based on this.

PR #1199, opened 2023-05-03
remove WebhookConfiguration to unblock CR update

Daniel-Fan:
If user execute the `migrate_tenant.sh` script after CS operator is upgraded, it will hit issue when updating CS CR because CS deployment is scaled down, but webhook configuration is looking for service endpoint.

In order to support run the script idempotent, delete CS CR validatingWebhookConfiguration when we scale down CS deployment.


YCShen1010 (approved):
/lgtm

PR #1204, opened 2023-05-03
ignore license accept in check mode

bitscuit:
No point in checking for license accept when nothing is being installed


Daniel-Fan (approved):
/lgtm

PR #1209, opened 2023-05-05
change cert manager cr to full name

bluzarraga:
It's possible if multiple cert managers are or have been installed on a cluster that simply searching for `certmanager` to find the cert manager CR will not work:
```
[root@api.alex3.cp.fyre.ibm.com ~]# oc get certmanager
No resources found
[root@api.alex3.cp.fyre.ibm.com ~]# oc get certmanagers.operator.ibm.com 
NAME      AGE
default   4h15m
```
Specifying the full name of the cr always works and should be a safer, more consistent option.

### DIFF
--- a/convert_to_multi_instance.sh
+++ b/convert_to_multi_instance.sh
@@ -107,7 +107,7 @@ function prepare_cluster() {
     cleanupZenService
 
     # uninstall singleton services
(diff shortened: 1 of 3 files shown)


Daniel-Fan (approved):
/lgtm

PR #1211, opened 2023-05-08
add flags back to parameters

bluzarraga:
Add back `--original-cs-ns` and `--services-ns` flags to denote parameters for the preload_data.sh script after discussion with cpd


Daniel-Fan (approved):

bitscuit (approved):
/lgtm

PR #1212, opened 2023-05-08
change backup image to existing mongo image

bluzarraga:
For https://github.ibm.com/IBMPrivateCloud/roadmap/issues/58594. In the function `dumpmongo` a mongodb-backup job is created using an image from quay (found https://github.com/IBM/ibm-common-service-operator/blob/scripts/preload_data.sh#L262). This image looks to be the same image that is used in mongodb so I have replaced the quay image with whatever the existing mongo in the original cs ns is using. I have run a preliminary test and the script was able to complete successfully and the data seems to have been transferred smoothly.

Testing:
- install a shared common services cluster with at least one instance of zen with iam enabled
- run preload_data.sh to preload data into target ns for new cs instance (aka services ns)
- edit common-service-maps to move whichever ns zen was installed into the new tenant
- run conversion script with zen using the new cs instance
- verify can login to zen's cpd route and that it redirects the login to `cp-console-<new cs instance ns>`

### DIFF
--- a/preload_data.sh
+++ b/preload_data.sh
@@ -246,6 +246,8 @@ function dumpmongo() {
     error "Cannot switch to $FROM_NAMESPACE"
   fi
 
+  ibm_mongodb_image=$(oc get pod icp-mongodb-0 -n $FROM_NAMESPACE -o=jsonpath='{range .spec.containers[0]}{.image}{end}')
+
   cat <<EOF >$TEMPFILE
 apiVersion: batch/v1
 kind: Job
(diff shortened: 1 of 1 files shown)


bluzarraga:
Ran this test successfully on a quick burn cluster. Was able to:
- use the existing mongo image on the cluster to run the backup job 
- complete the rest of the preload script 
- run conversion and install new cs instance where data was preloaded
- log in via LDAP to cpd-cp4i2 (the namespace w/ zen w/ iam enabled that moved to the new cs instance)

Need to verify airgap still but that will be done by the SERT team as there is an airgap specific SERT issue open already for the scripts.

bitscuit (approved):
/lgtm

PR #1213, opened 2023-05-08
edit odlm sub in place

bluzarraga:
To avoid applying the changed subscription on top of the existing and possibly running into an error due to resourceVersion mismatches, we need to update the value in place in the `isolate_odlm` function. This function is called in the isolate, conversion, and rollback script so needs to be updated in all three locations.

To do this, I have added a patch statement to edit the existing subscription. Unfortunately, patches to an array completely overwrite the array. This means I cannot just update the `ISOLATE_MODE` value without wiping the values of the other `spec.config.env` variables. To get around this, I loop through the output of `spec.config.env[]` and add the name and value back to the subscription in the order they were read in. If the variable name is `ISOLATE_MODE`, the script makes sure it is set to "true" (false in the case of the rollback script). I was able to run the script a couple times and saw that the `patch_string` variable looked in line with the command I tested by hand and that the odlm subscription and deployment were being properly updated to reflect the new value. I have only tested the isolate script so far.

Testing isolate.sh:
- setup cluster with shared common services installed 
- run isolate.sh
- verify script completes successfully
- verify odlm sub spec.config.env[ISOLATED_MODE]=true

Testing conversion script:
- setup cluster with shared common services installed
- create cs maps, does not necessarily need more than one tenant defined
- run conversion script
- verify script completes successfully
- verify odlm sub spec.config.env[ISOLATED_MODE]=true

Testing rollback script:
- probably best to use the conversion cluster after successful test since rollback is the opposite
- uninstall new common service instances with the uninstall-dedicated-cs-instance.sh script (use -f flag)
- remove the controlNamespace value from the common-service-maps configmap and consolidate tenants into one (should only be one map-to-common-services-namespace)
- run rollback script
- verify script completes successfully
- verify odlm sub spec.config.env[ISOLATED_MODE]=*false*

### DIFF
--- a/convert_to_multi_instance.sh
+++ b/convert_to_multi_instance.sh
@@ -695,7 +695,6 @@ function isolate_odlm() {
         fi
         count=$((count + 1))
     done
-    debug1 "patch string $patch_string"
     #use the patch string to apply the isolate mode patch
     ${OC} patch subscription.operators.coreos.com ${sub_name} -n ${ns} --type=merge -p '{"spec": {"config": {"env": ['"${patch_string}"']}}}'
     if [[ $? -ne 0 ]]; then
--- a/convert_to_multi_instance.sh
+++ b/convert_to_multi_instance.sh
@@ -666,25 +666,41 @@ function check_topology() {
 }
 
 function isolate_odlm() {
-    package_name=$1
-    ns=$2
+    local package_name=$1
+    local ns=$2
     # get subscription of ODLM based on namespace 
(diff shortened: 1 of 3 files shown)


bluzarraga:
Just finished testing all three scripts with the odlm change and it appears to be working as intended. Ran conversion, then rollback, then isolation. Conversion successfully changed ISOLATE_MODE to true while keeping each of the other spec.config.env variables. Rollback then did the same but changed ISOLATED_MODE to false (as it should). Isolation had the same results as the conversion script. I think this is ready for review, feedback on implementation is always appreciated

Daniel-Fan (approved):
/lgtm

PR #1224, opened 2023-05-10
sync scripts 2023-05-10

bitscuit:


bluzarraga:
I think the only unique change being merged in is 1217, the other three commits are from yesterday's sync

PR #1197, opened 2023-05-02
check for ldaps secret in backup_restore_mongo

bluzarraga:
Opened this pr for tracking the work I had locally, not ready for review yet

PR #1225, opened 2023-05-11
ensure ldaps secret exists in target namespace after restore

bluzarraga:
Replacing https://github.com/IBM/ibm-common-service-operator/pull/1197

When running either of the scripts for backing up and restoring mongo data into a new namespace, the ldaps secret can be created without the certificate filled in. This would disable the ability to log in using LDAP so we need to make sure this secret is there and matches the original.

Testing preload:
- install shared cluster with LDAP enabled and users created
- run preload script
- verify secret is copied over into new namespace and equals original
- convert cluster with new cs in target namespace from preload
- make sure LDAP login works with new cs instance

Testing backup_restore:
- install shared cluster with LDAP enabled and users created
- run conversion
- run backup restore script
- verify secret is present in new namespace and equals original
- make sure LDAP login works with new cs instance

### DIFF
--- a/backup_restore_mongo.sh
+++ b/backup_restore_mongo.sh
@@ -69,6 +69,7 @@ function main() {
     if [[ $restore == "true" ]]; then
         prep_restore
         restore
+        check_ldap_secret
         refresh_auth_idp
     fi
     if [[ $cleanup == "true" ]]; then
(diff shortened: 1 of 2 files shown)


Daniel-Fan (approved):
/lgtm

PR #1227, opened 2023-05-11
change cert manager check to just webhook, check for dupe deploys

bluzarraga:
If the cluster has a redhat cert manager deployed, the check for `cert-manager-controller` will fail because redhat names its controller pod and deployment just `cert-manager`. This would then cause the isolate.sh script to fail. Instead of checking for each of the three operands, change the check to only check for the webhook pod to be ready as well as check for no duplicate webhook deployments on the cluster.

Testing in case of open source cert manager deployed:
- install redhat cert manager on cluster
- setup shared cs environment, zen install a plus
- run isolate.sh, verify script completes successfully

~Testing in case of cert manager no longer requested:~
- ~install common services in ibm-common-services~
- ~create opreq for cert manager in a ns other than ibm-common-services (let's call it test-ns)~
    - ~make sure cert manager is only installed via this opreq~
- ~run isolate script excluding test-ns~
- ~make sure new opreq is created in control namespace for cert manager~
- ~make sure script completes successfully and cert manager is deployed~

### DIFF
--- a/isolate.sh
+++ b/isolate.sh
@@ -87,11 +87,7 @@ function main() {
     check_cm_ns_exist "$ns_list $CONTROL_NS" # debating on turning this off by default since this technically falls outside the scope of isolate
     isolate_odlm "ibm-odlm" $MASTER_NS
     restart
-    if [[ $CERT_MANAGER_MIGRATED == "true" ]]; then
-        wait_for_certmanager "$CONTROL_NS"
-    else
-        info "Cert Manager not migrated, skipping wait."
-    fi
+    wait_for_certmanager "$CONTROL_NS"
     success "Isolation complete"
 }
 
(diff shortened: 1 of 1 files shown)


bluzarraga:
I have tested both of the scenarios described above successfully. For the second scenario, it is debatable whether we need to create a new opreq for cert manager in a situation where every cp2 requesting namespace is excluded. That being said, the zen team has asked us before to check for cert manager to be ready in the isolate script so I believe they expect it to be there too. Adding the cert manager opreq is a fix that ensures our script completes and that cert manager is on the cluster but maybe more conversation can be had between zen and bedrock about expectations for cert manager, who installs it, and what the isolate script should do in a position where cert manager is not present if adding it back is not the long term answer.

That being said, PR is ready for review.

bluzarraga:
Ongoing discussion about whether the change to always install cert manager in case none is requested is necessary so I have removed it from this pr and opened a new one (https://github.com/IBM/ibm-common-service-operator/pull/1239) Sorry for the messiness in this pr.

bluzarraga:
The changes from this pr were merged in with https://github.com/IBM/ibm-common-service-operator/pull/1239. Closing this pr

PR #1237, opened 2023-05-15
make preload data roks compatible

bluzarraga:
The preload script was not roks compatible due to specifying the storageclass name in the PVC creation during the `swapmongopvc` function. I have not seen this issue myself but it was brought up by Mike Kaczmarski and this was ostensibly the only change required.

Testing:
- setup shared common services installed on a roks cluster, setup ldap users
- run the preload_data.sh script
- ensure script completes successfully, verify the  mongodbdir-icp-mongodb-0 pvc is bound in the target namespace
- run conversion using the target namespace of the preload script as the new cs instance, verify ldap can login to new cs namespace

### DIFF
--- a/preload_data.sh
+++ b/preload_data.sh
@@ -359,24 +359,25 @@ function swapmongopvc() {
     error "Volume for pvc  cs-mongodump not found in $FROM_NAMESPACE"
   fi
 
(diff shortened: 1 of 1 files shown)


Daniel-Fan (approved):
/lgtm

PR #1171, opened 2023-04-24
only check for iam when a zenservice has requested it

bluzarraga:
The conversion script waits for iam to come ready in order to reconcile zenservices with the new iam deployment. Ran into an issue during cp4d testing where a zen service did not enable iam so the script failed by timing out waiting for iam and consequently did not reconcile the zenservices that did request iam. This change makes it so we only wait for iam to come ready in the cs namespaces serving zenservices that enable iam.

Test pending

PR #1238, opened 2023-05-15
only wait for iam if it's requested during rollback

bluzarraga:
Similar to the changes made for the conversion script here https://github.com/IBM/ibm-common-service-operator/pull/1171

Low priority given current use case for rollback and the scenario required for this to be a blocker but I ran into it in my testing of https://github.com/IBM/ibm-common-service-operator/pull/1227 so opened a pr before forgetting. 

Basically, we are looking if any of the cp namespaces have a zenservice that requested iam. If they do, we wait for iam to come ready in master ns and if they do not, we skip the wait.

Testing:
- setup a converted cluster with zen requested, create zenservice that sets `iamIntegration: false`
- ensure no other zenservices enable iam
- rollback the cluster
- verify the script completes and does not wait on iam to come ready

example cs maps for start of the test:
```
apiVersion: v1
kind: ConfigMap
metadata:
  name: common-service-maps
  namespace: kube-public
data:
  common-service-maps.yaml: |
    controlNamespace: cs-control
    namespaceMapping:
    - requested-from-namespace:
      - zen
      map-to-common-service-namespace: ibm-common-services
```
example csmaps pre-rollback (just make sure to remove the control namespace before running script):
```
apiVersion: v1
kind: ConfigMap
metadata:
  name: common-service-maps
  namespace: kube-public
data:
  common-service-maps.yaml: |
    namespaceMapping:
    - requested-from-namespace:
      - zen
      map-to-common-service-namespace: ibm-common-services
```

### DIFF
--- a/convert_to_multi_instance.sh
+++ b/convert_to_multi_instance.sh
@@ -255,7 +255,8 @@ function check_IAM(){
         do
             zenservice_exists=$(${OC} get zenservice -n $cp_namespace || echo fail)
             if [[ $zenservice_exists != "fail" ]] && [[ $zenservice_exists != "" ]]; then
-                iam_enabled=$(${OC} get zenservice -n $cp_namespace -o yaml | grep iamIntegration | awk '{print $2}')
+                zenservice=$(${OC} get zenservice -n $cp_namespace --no-headers | awk '{print $1}')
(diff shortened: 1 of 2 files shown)


Daniel-Fan (approved):
/lgtm

PR #1239, opened 2023-05-16
Check if cert manager is deployed in the instance that it is not requested (excluded-ns edge case)

bluzarraga:
This pr is predicated on change made to how isolate waits on cert manager in https://github.com/IBM/ibm-common-service-operator/pull/1227 and contains some of the same changes. Copying over some of the relevant comments from that pr here:

It is debatable whether we need to create a new opreq for cert manager in a situation where every cp2 requesting namespace is excluded. That being said, the zen team has asked us before to check for cert manager to be ready in the isolate script so I believe they expect it to be there too. Adding the cert manager opreq is a fix that ensures our script completes and that cert manager is on the cluster but maybe more conversation can be had between zen and bedrock about expectations for cert manager, who installs it, and what the isolate script should do in a position where cert manager is not present if adding it back is not the long term answer.

Testing in case of cert manager no longer requested:
- install common services in ibm-common-services
- create opreq for cert manager in a ns other than ibm-common-services (let's call it test-ns)
    - make sure cert manager is only installed via this opreq
- run isolate script excluding test-ns
- make sure new opreq is created in control namespace for cert manager
- make sure script completes successfully and cert manager is deployed

After discussion with the zen team today (5/16), they determined they could make some manual changes to their testing pipeline that would circumvent this problem and downgraded its urgency. Other approaches are being explored. With this in mind, I have removed these changes from the previous pr (so the other changes in that pr could still be merged) and opened this one should we end up needing this change after all.

### DIFF
--- a/isolate.sh
+++ b/isolate.sh
@@ -87,11 +87,7 @@ function main() {
     check_cm_ns_exist "$ns_list $CONTROL_NS" # debating on turning this off by default since this technically falls outside the scope of isolate
     isolate_odlm "ibm-odlm" $MASTER_NS
     restart
-    if [[ $CERT_MANAGER_MIGRATED == "true" ]]; then
-        wait_for_certmanager "$CONTROL_NS"
-    else
-        info "Cert Manager not migrated, skipping wait."
-    fi
+    wait_for_certmanager "$CONTROL_NS"
     success "Isolation complete"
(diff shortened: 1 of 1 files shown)


Daniel-Fan (approved):
/lgtm

PR #1243, opened 2023-05-16
check necessary br files in filesystem

bluzarraga:
Adding the scripts repo to the case bundles means that we can expect the file system to be the same in the bundle (and in airgap scenarios) as what is in the scripts repo. As it is now, the checks for files to be present would fail since the necessary files are in separate directories. This would cause the script to fail ostensibly due to a lack of prereq files when they are already present. Changing the check to check the relative file path should resolve that. We'll just need to make sure to update it should the layout of the scripts branch change.


bluzarraga:
closing in favor of implementation in https://github.com/IBM/ibm-common-service-operator/pull/1270

PR #1245, opened 2023-05-17
uninstall and force uninstall script

qpdpQ:
### uninstall and force uninstall script 
For ticket: https://github.ibm.com/ibmprivatecloud/roadmap/issues/58490
Base on https://github.ibm.com/IBMPrivateCloud/cs-dev-tools/blob/master/install/cp3pt0-install/uninstall_tenant.sh

### Usage
```
Options:
   --oc string                    File path to oc CLI. Default uses oc in your PATH
   --operator-namespace string    Required. Namespace to install Foundational services operator
   -f                             Enable force delete
   -h, --help                     Print usage information
```

### How to test: 
install CP3 and run `./uninstall_tenant.sh --operator-namespace ibm-common-services -f`
After running this script, all the resources in/and the target namespaces should be removed. 
It should clean all the namespaces in the namespace-scope configmap.
It will keep control namespace/ cert-manager namespace/ licensing namespace untouched. 
(If you want, you can run `./uninstall_tenant.sh --operator-namespace cs-control -f` and it will cleanup the control namespace)

### Current issues
I tried it with in-place upgrade and isolate upgrade. It looks good to me. But it cannot handle other cluster scope resources.I think we need to add them manually.

### DIFF
--- a/cp3pt0-deployment/common/utils.sh
+++ b/cp3pt0-deployment/common/utils.sh
@@ -986,7 +986,7 @@ function delete_operand_finalizer() {
     local crds=$1
     local ns=$2
     for crd in ${crds}; do
-        if [ "${crd}" != "packagemanifests.packages.operators.coreos.com" ]; then
+        if [ "${crd}" != "packagemanifests.packages.operators.coreos.com" ] && [ "${crd}" != "events" ]; then
             crs=$(${OC} get ${crd} --no-headers --ignore-not-found -n ${ns} 2>/dev/null | awk '{print $1}')
             for cr in ${crs}; do
                 msg "Removing the finalizers for resource: ${crd}/${cr}"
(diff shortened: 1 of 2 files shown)


qpdpQ:
I think we can merge it for now since Zen team needs some time to incorporate it into olm-utils.
For cluster scope resources, we can add an enhancement for this script or a separate script for that.

bluzarraga:
I have a feeling zen will be asking to include the cluster scope resources as well, namely the webhook resources as they've been creating a lot of errors recently. We could add it in now and create the enhancement ticket, it just seems likely to me that they will hit a problem with webhooks not allowing changes to resources and ask us why

bluzarraga:
Not sure how I missed that webhook cleanup, and yeah, we'll have to have them manually cleanup the common service maps after running the script. A possible enhancement could be doing it in the script later on

bluzarraga:
uninstalled a couple tenants and discussed changes with Allen, I think this looks good

PR #1247, opened 2023-05-18
move ns existence check earlier in script

bluzarraga:
Licensing noticed if cs-control was not already created before running script, isolate would fail if it needed to move over licensing configmaps because the check for cs-control to exist was after the call to move licensing configmaps.

Test:
- setup environment for isolation with at least one licensing configmap (can be dummy, see list of possible cms in script)
- do not create control namespace before running script
- run isolate script
- ensure control namespace is created before licensing configmaps are moved over
- ensure script completes successfully


Daniel-Fan (approved):
/lgtm

PR #1250, opened 2023-05-18
don't remove IBMLicensing singleton CR which has customisations

pgodowski:
and remove bindinfo, which will be left-over


Daniel-Fan (approved):
/lgtm

PR #1255, opened 2023-05-19
preload: change restore image to existing mongo

bluzarraga:
update image used for restore job to the existing mongo image on the cluster for airgap purposes


Daniel-Fan (approved):
/lgtm

PR #1256, opened 2023-05-19
add resource limits and requests to backup restore jobs

bluzarraga:
CPD quality issue brought up an issue where they are trying to preload mongo into a namespace with an existing cpd deployment. Part of this existing deployment is a resource quota that is enforcing pods have values for their resource limits and requests:
>  Warning  FailedCreate  2s    job-controller  Error creating: pods "mongodb-restore-l7sp5" is forbidden: failed quota: cpd-quota: must specify limits.cpu,limits.memory,requests.cpu,requests.memory

This is preventing the preload script from completing and is a blocker for zen. Adding the resource values should be sufficient for resolving this issue.

Testing:
- run preload script as normal, bonus if a resource quota is created in the target namespace
- ensure script completes
- view the restore pod in the target namespace and verify resources are specified in the yaml

### DIFF
--- a/preload_data.sh
+++ b/preload_data.sh
@@ -270,6 +270,13 @@ spec:
       containers:
       - name: cs-mongodb-backup
         image: $ibm_mongodb_image
+        resources:
+          limits:
+            cpu: 500m
+            memory: 500Mi
+          requests:
+            cpu: 100m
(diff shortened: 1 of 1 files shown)


Daniel-Fan (approved):
/lgtm

PR #1258, opened 2023-05-19
retry patch commonservice cr for webhook issue

qpdpQ:
for ticket: https://github.ibm.com/IBMPrivateCloud/roadmap/issues/58771

if we fail to patch CommonService CR we should retry it after installing cs-operator


Daniel-Fan (approved):
/lgtm

PR #1259, opened 2023-05-19
Generalize copy function, add more customization resources to copy over

bluzarraga:
We need to make sure all of the possible customization options an end user may have are carried over from CP2 into the new cs namespace to make the transition as seamless as possible. So far the resources included in preload_data.sh consist of:
- backup of original mongo
- platform-auth-idp-credentials secret
- platform-auth-ldaps-ca-cert secret
- ibm-cpp-config configmap
- common-web-ui-config configmap
- common-service commonservice (CR)

Ongoing investigation as to what else needs to be included in this process

Testing:
- setup cp2 cluster with these customization resources created and values assigned (to verify they carry over)
    - for configmaps, can set a `data.test: "test"` value
        - verified common-web-ui-config customization carried over with @sgrube yesterday
    - for commonservice, change the size to anything but starterset
- verify script completes
- verify resources copied over and customizations applied

### DIFF
--- a/preload_data.sh
+++ b/preload_data.sh
@@ -33,8 +33,11 @@ function main() {
     # run backup preload
     backup_preload_mongo
     # copy im credentials
-    copy_secret "platform-auth-idp-credentials"
-    copy_secret "platform-auth-ldaps-ca-cert"
+    copy_resource "secret" "platform-auth-idp-credentials"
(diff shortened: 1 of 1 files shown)


bluzarraga:
Removing work in progress tag, we will need to merge in other customization resources once IAM has the processes for them ready.

Daniel-Fan (approved):
/lgtm

PR #1262, opened 2023-05-19
match admin pass to original

bluzarraga:
Currently, the admin user/pass is hardcoded so any mongo that gets it's data after the preload process is using the exact same admin user/pass combo regardless of cluster or which mongo its data originated from. Change this to use the existing admin user/pass combo from the mongo the data originates from instead.

test:
- run preload
- verify script completes
- check icp-mongodb-admin secret and verify it matches in original and new cs ns


Daniel-Fan (approved):
/lgtm

PR #1263, opened 2023-05-19
clean up the resources when services ns is changed

YCShen1010:
For issue: https://github.ibm.com/IBMPrivateCloud/roadmap/issues/57401
Testing image: quay.io/yuchen_shen/cs_operator:clean_resources

When the first time to deploy CP3 in the cluster, by default `servicesNamespace` is the same as `operatorNamespace`, the following resources will be generated in that tmp `serivcesNamespace`.
After servicesNamespace is changed, the resources should be cleaned up properly and create under the new `serivcesNamespace`
#### Resources
- CertManager CRs including certificates and issuers
- default OperandRegistry and OperandConfig `common-service`
#### Verify 
1. Using setup_tenant script to deploy a CP3 (could choose Topology B with 2 different tethered ns, and same ns for services ns and operator ns)
2. apply the testing image
3. add label `operator.ibm.com/managedByCsOperator: 'true'` into `cs-ca-issuer`, `cs-ss-issuer` and `cs-ca-certificate`
4. run `setuo_tenant.sh` again with a new ns for `ServicesNamespace`
5. Observation results: all the resources default OperandRegistry, OperandConfig `common-service`, CertManager CRs in original services ns are deleted and create under new services ns.

### DIFF
--- a/controllers/bootstrap/init.go
+++ b/controllers/bootstrap/init.go
@@ -206,6 +206,11 @@ func (b *Bootstrap) InitResources(instance *apiv3.CommonService, forceUpdateODLM
 	// Install/update Opreg and Opcon resources before installing ODLM if CRDs exist
 	if existOpreg && existOpcon {
 
+		klog.Info("Checking OperandRegistry and OperandConfig deployment status")
+		if err := b.ConfigODLMOperandManagedByOperator(ctx); err != nil {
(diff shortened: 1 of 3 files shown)


Daniel-Fan (approved):
/lgtm

PR #1265, opened 2023-05-23
Uninstall ODLM from servicesNs in all SoD scenario

Daniel-Fan:
Uninstall ODLM in other namespace other than operator namespace.

This is used to handle the All Namespace Upgrade Scenario when CS operator will move ODLM from `ibm-common-services` namespace to `openshift-operators` namespace.


YCShen1010 (approved):
/lgtm

PR #1266, opened 2023-05-23
redirect output to a log file

YCShen1010:
for issue: https://github.ibm.com/ibmprivatecloud/roadmap/issues/57620#issuecomment-57362560
- Check if the passed in value of DEBUG is 1 or 0, otherwise return error
- Redirect stdout and stderr to the log file, overwriting it each time
- Remove ANSI escape sequences from log file

#### log file example
```
[✔] oc command available
[✔] Channel is valid
[✔] Profile size is valid.
[INFO] Namespace test-ns already exists. Skip creating
[INFO] Namespace test-ns already exists. Skip creating
[INFO] Namespace tenant1-a already exists. Skip creating
[INFO] Namespace tenant1-b already exists. Skip creating
[INFO] Checking existing OperatorGroup in test-ns:

[INFO] OperatorGroup already exists in test-ns. Skip creating
```

### DIFF
--- a/cp3pt0-deployment/common/utils.sh
+++ b/cp3pt0-deployment/common/utils.sh
@@ -1009,3 +1009,25 @@ function delete_operand_finalizer() {
         fi
     done
 }
+
+function save_log(){
+    local LOG_DIR="$BASE_DIR/$1"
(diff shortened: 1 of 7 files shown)


Daniel-Fan (approved):
/lgtm

PR #1268, opened 2023-05-24
Bug/58774: re-fix of license service migration with customisation preserved

pgodowski:
fixes: https://github.ibm.com/IBMPrivateCloud/roadmap/issues/58774

moved restore of IBMLicensing CR just after the LicSvc operator is removed, to avoid the situation where such operator is redeployed later, while the CR restore was conditional based on the namespaces/topology configuration


Daniel-Fan (approved):
/lgtm

PR #1270, opened 2023-05-24
Update backup_restore_mongo to work on ROKS environments

bluzarraga:
Previously, because the script created it's own storage class, the backup_restore_mongo.sh script did not work on ROKS. The changes in this pr fix that by reusing the storage class the original mongo install already uses instead of creating a new one.
Other changes include changing where the file is located as well as the necessary support files to make airgap process easier. Also updated preload_data.sh to remove any instance of `oc` and replace it with `${OC}`.

Testing:
- get a roks cluster with common services installed
  - really just need iam and mongo
  - needs LDAP setup
- run conversion script to split two cp namespaces to use two different cs instances
- verify cannot login with ldap to the new instance
- run backup_restore_mongo.sh to restore mongo data into new cs instance
- log in with LDAP to the new cs instance's cp-console

I was able to get this to work using a SERT cluster

~Still working on the airgap, need to pass in the image to the scripts that run the backup and restore jobs~


bluzarraga:
changes included in #1330 closing here

PR #1271, opened 2023-05-24
Creating and updating common-service-maps when reconciling master CS CR

YCShen1010:
For issue: https://github.ibm.com/IBMPrivateCloud/roadmap/issues/57890
Testing image: `quay.io/yuchen_shen/cs_operator:csmaps_update`

### Context
`common-service-maps` ConfigMap will be updated whenever `servicesNamespace` is changed in CommonService CR even the new servicesNamespace is already in the tenant scope. Therefore, no restart is triggered.

### How to test
- Option 1 
  1. Install new IBM Cloud Pak foundational services (v4.0) by using the same namespace for `servicesNs` and `operatorNs`
  2. Apply the testing image 
  3. Patch CommonService CR with different services namespace which is already in the `WATCH_NAMESPACE`
  4. Observe `namespaceMapping` is updated with new servicesNamespace
- Option 2
  1. delete a namespace in `spec.namespaceMembers` in nss CR
  2. the namespace will be auto removed from `WATCH_NAMESPACE`, and followed by `common-service-maps` updated
- Option 3
  1. Running `setup_tenant.sh` with a different servicesNamespace, `common-service-maps` will be updated as well.

### DIFF
--- a/controllers/commonservice_controller.go
+++ b/controllers/commonservice_controller.go
@@ -19,6 +19,7 @@ package controllers
 import (
 	"context"
 	"fmt"
+	"os"
 	"reflect"
 	"strings"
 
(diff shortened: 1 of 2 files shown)


Daniel-Fan (approved):
/lgtm

PR #1274, opened 2023-05-25
check for ibmlicensing resource to exist before backup

bluzarraga:
If there is no ibmlicensing resource on a cluster, the isolate script will fail with `error: the server doesn't have a resource type "IBMLicensing"`. This pr adds a check for  the resource to exist before continuing to prevent an unnecessary script exit.

Testing
- setup cluster with common services but without licensing deployed
- run the isolate script
- verify script completes and this info line is output: "No ibmlicensing resources on cluster, skipping backup."

### DIFF
--- a/isolate.sh
+++ b/isolate.sh
@@ -338,7 +338,12 @@ function uninstall_singletons() {
 
     migrate_lic_cms $MASTER_NS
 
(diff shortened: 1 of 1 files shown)


Daniel-Fan (approved):
/lgtm

PR #1276, opened 2023-05-25
wait for namespace-scope cm to match expected value after isolation

bluzarraga:
CPD has asked us to wait on the namespace-scope cm in the original namespace to be updated before exiting from the script. This pr adds that wait as well as cleans up a few other issues.

Testing:
- setup cs environment
- run isolation script, make sure the original tenant changes in some way (either via excluded ns or insert ns)
- verify script completes

### DIFF
--- a/isolate.sh
+++ b/isolate.sh
@@ -70,8 +70,7 @@ function main() {
             CONTROL_NS=$2
             shift
             ;;
-        "-v" | "--debug")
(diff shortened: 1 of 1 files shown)


bluzarraga:
ready for review

Daniel-Fan (approved):
/lgtm

PR #1278, opened 2023-05-25
fix scale down function for csv and deployment missing

YCShen1010:
cover corner case where the cluster does not have ODLM


Daniel-Fan (approved):
/lgtm

PR #1284, opened 2023-05-26
Add label selector to filter out cloned CS CRs

Daniel-Fan:
### Issue
https://github.ibm.com/IBMPrivateCloud/roadmap/issues/58257

The issue only happens on Advance Topology involving multiple namespaces.
When we update the default CommonService CR `.spec.size` from `large` back to `starterset`, The OperandConfig still contains large profile with `replicas: 3`

### Context
In advance topology, common service operator will clone default CommonService CR into all other namespaces in this tenant. It is used to help other common service operator installed in the tenant to identify its position.

### Root cause
Those cloned CommonService CRs are identical to the default CommonService CR, except a special label
```
operator.ibm.com/common-services.cloned-from: <operator-namespace>
```
In other words, those cloned CommonService CRs contains the size profile configurations which is the same as default CS CR.

When we update the default CommonService CR `.spec.size` from `starterset` to `large`
1. It scales up services with `replica: 3` as we expected.
2. Those cloned CommonService CR will also have `.spec.size: large`

However, when we update the default CommonService CR `.spec.size` from `large` back to `starterset`, the reconciliation logic
1. Apply the size configuration based on the all available CommonService CRs in the tenant
2. Propagate the configurations in default CommonService CR into other cloned CS CRs.

**It leads to a problem that when applying the size configuration, it counts the size profile in cloned CommonService CR which has not been updated to `starterset` yet. And the `large` size configuration is preserved in the OperandConfig**
 
### Solution
When list the CommonService CR list, we will construct a label selector to filter out those cloned CS CRs, and its size configurations will not be counted anymore.

### Test
1. Install a Bedrock v4 in Separation of control and data topology - DO NOT deploy any OperandRequest
2. Replace CS operator image with `docker-na-public.artifactory.swg-devops.com/hyc-cloud-private-scratch-docker-local/ibmcom/common-service-operator-amd64:dev`
3. Switch the profile size in default CommonService from `starterset` to `large`
```
kind: CommonService
metadata:
  name: common-service
  namespace: cloudpak-control
spec:
  license:
    accept: true
  operatorNamespace: cloudpak-control
  servicesNamespace: cloudpak-data
  size: starterset  -> large
```
4. Check the OperandConfig in `cloudpak-data` namespace to verify that replica is 3 for `mongodb`
5. Switch the profile size in default CommonService from `large` back to `starterset`
```
kind: CommonService
metadata:
  name: common-service
  namespace: cloudpak-control
spec:
  license:
    accept: true
  operatorNamespace: cloudpak-control
  servicesNamespace: cloudpak-data
  size: large -> starterset
```
6. Check the OperandConfig in `cloudpak-data` namespace to verify that replica is back to 1 for `mongodb`

### DIFF
--- a/controllers/bootstrap/init.go
+++ b/controllers/bootstrap/init.go
@@ -35,6 +35,7 @@ import (
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
 	"k8s.io/apimachinery/pkg/labels"
 	"k8s.io/apimachinery/pkg/runtime/schema"
+	"k8s.io/apimachinery/pkg/selection"
 	"k8s.io/apimachinery/pkg/types"
 	utilwait "k8s.io/apimachinery/pkg/util/wait"
 	discovery "k8s.io/client-go/discovery"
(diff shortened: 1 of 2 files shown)


bitscuit (approved):
/lgtm

PR #1288, opened 2023-05-30
[Bedrock 4.1] Implement Concatenating OperandConfig YAMLs

YCShen1010:
Introduced a new function `ConcatenateConfigs` essentially combines the `Services` from two OperandConfig objects and returns the concatenated YAML representation of the modified OperandConfig.

for issue: https://github.ibm.com/IBMPrivateCloud/roadmap/issues/57539

####  The function performs the following steps:

- It creates two OperandConfig objects: `baseConfig` and `insertedConfig` by unmarshaling the YAML templates using the applyTemplate function.
- It appends the `Services` from `insertedConfig` to the existing `Services` in `baseConfig`.
- It marshals the modified base object back to YAML format using json.Marshal.
- It returns the resulting YAML string and any error that occurred during the process.

#### Verify
test image: quay.io/yuchen_shen/cs_operator:split_opcon
delete origin OprandConfig and it will create a new one with the entries of CP3 services and followed by the services with suffix.


Daniel-Fan (approved):
/lgtm

PR #1290, opened 2023-05-31
always call wait for certmanager function

bluzarraga:
In the scenario where the isolate script fails and is then run again, it is possible given current implementation that the cert manager is deleted and marked for migration on the first failed run. Then, if the script failed before cert manager was redeployed, it would not be marked for migration on the second run. Combine this situation with the possibility that operand requests requesting cert manager are now all out of scope (due to the excluded parameter) and it creates a scenario where cert manager is not installed at all.

The code change in this pr would always call the `wait_for_certmanager` function. This function calls `check_if_certmanager_deployed` which checks the scope for operand requests calling for cert manager. If none exist in scope, it will create one in the original cs ns and which then deploys cert manager to the control namespace. This way, we always have cert manager deployed. If community is already installed, then only the operator pod will start up for the ibm cert manager and the rest of the wait checks will still pass as they are looking for the webhook pod to be ready somewhere on the cluster.

Testing:
- setup cp2 cluster with only opreq for cert manager in `test-ns` and no operand requests for im
- run isolate.sh excluding `test-ns`
- verify script completes
- verify cert manager operator pod exists in control namespace

### DIFF
--- a/cp3pt0-deployment/common/utils.sh
+++ b/cp3pt0-deployment/common/utils.sh
@@ -387,17 +387,12 @@ function is_sub_exist() {
 }
 
 function check_cert_manager(){
-    csv_count=$(${OC} get csv -A | grep "cert-manager"| wc -l | tr -d " ")
+    csv_count=`$OC get csv |grep "cert-manager"|wc -l`
     if [[ $csv_count == 0 ]]; then
         error "Missing a cert-manager"
     fi
-    # if installed in all namespace mode or alongside cp2 cert manager, 
-    # csv_count will be >1, need to check for multiple deployments
     if [[ $csv_count > 1 ]]; then
-        webhook_deployments=$(${OC} get deploy -A --no-headers --ignore-not-found | grep "cert-manager-webhook" -c)
-        if [[ $webhook_deployments != "1" ]]; then
-            error "Multiple cert-managers found. Only one should be installed per cluster"
-        fi
+        error "Multiple cert-manager csv found. Only one should be installed per cluster"
     fi
 }
 
--- a/cp3pt0-deployment/common/utils.sh
+++ b/cp3pt0-deployment/common/utils.sh
@@ -391,8 +391,13 @@ function check_cert_manager(){
     if [[ $csv_count == 0 ]]; then
         error "Missing a cert-manager"
     fi
+    # if installed in all namespace mode or alongside cp2 cert manager, 
+    # csv_count will be >1, need to check for multiple deployments
     if [[ $csv_count > 1 ]]; then
-        error "Multiple cert-manager csv found. Only one should be installed per cluster"
+        webhook_deployments=$(${OC} get deploy -A --no-headers --ignore-not-found | grep "cert-manager-webhook" -c)
+        if [[ $webhook_deployments != "1" ]]; then
+            error "Multiple cert-managers found. Only one should be installed per cluster"
+        fi
     fi
 }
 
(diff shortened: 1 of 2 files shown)


bluzarraga:
I have added the check_for_certmanager function from utils.sh to isolate.sh to try to avoid the problems found when running isolate.sh with more than one cert manager on a cluster. I was having some trouble getting the function to work so I also made some updates to it

isolate.sh
```diff
@@ -112,11 +112,7 @@ function main() {
         info "Licensing not marked for backup, skipping."
     fi
     restart
-    if [[ $CERT_MANAGER_MIGRATED == "true" ]]; then
-        wait_for_certmanager "$CONTROL_NS" "${ns_list}"
-    else
-        info "Cert Manager not migrated, skipping wait."
-    fi
+    wait_for_certmanager "$CONTROL_NS" "${ns_list}"
```
Daniel-Fan, line 115: I am testing on a cluster without any CP2 workload installed(no CP2 cert-manager), and doing isolation.
I do not see CP2 cert-manager is installed in `cs-control` namespace. Not sure if we should always wait for cert-manager

bluzarraga: in wait_for_certmanager, check_if_cert_manager_deployed is called and if no opreq exists in scope, this function will create one so we should always have cert manager deployed to control namespace. https://github.com/IBM/ibm-common-service-operator/blob/3bd2a10da2c5497f66b53085c90e4c37c42adfeb/isolate.sh#L645

bluzarraga: if there is an opreq that is still in the scope of the master cs ns that requests cert manager, then cert manager will be deployed to cs control. If there is not an opreq within the scope of the master cs ns, then we create an opreq that only requests cert manager in the master cs ns so cert manager will be deployed in cs control

bluzarraga: Would it be better to not check specifically for ibm-cert-manager-operator in control ns and instead check for `cert-manager-operator` deployment across all namespaces?

bluzarraga:
The changes to the utils.sh helper script also seem relevant to recently opened cpd-quality issue 9998

isolate.sh
Daniel-Fan: I was hesitant to `always have cert manager deployed to control namespace`.
But I think it is ok to always have CP2 cert manager in control namespace.

> instead check for cert-manager-operator deployment across all namespaces?

Does this count red hat cert manager? Or it is just looking for ibm cert-manager-operator.

cp3pt0-deployment/common/utils.sh
```diff
@@ -387,12 +387,17 @@ function is_sub_exist() {
...
         error "Missing a cert-manager"
     fi
+    # if installed in all namespace mode or alongside cp2 cert manager, 
+    # csv_count will be >1, need to check for multiple deployments
     if [[ $csv_count > 1 ]]; then
-        error "Multiple cert-manager csv found. Only one should be installed per cluster"
+        webhook_deployments=$(${OC} get deploy -A --no-headers --ignore-not-found | grep "cert-manager-webhook" -c)
+        if [[ $webhook_deployments != "1" ]]; then
+            error "Multiple cert-managers found. Only one should be installed per cluster"
+        fi
```
Daniel-Fan, line 400: @bluzarraga 
> The changes to the utils.sh helper script also seem relevant to recently opened cpd-quality issue 9998

I believe it will impact `olm-utils` functionality that they only have tenant permission to get CSV, `-A` will require a cluster permission to do so.

bluzarraga:
After discussion with @Daniel-Fan removed changes to utils.sh and created a new function in isolate.sh since it has more permissions.

Daniel-Fan (approved):
/lgtm

PR #1298, opened 2023-06-02
[323] allow multiple opbi

qpdpQ:
For ticket: https://github.ibm.com/IBMPrivateCloud/roadmap/issues/58807
Use k8s resources to create licensing operandBindInfo instead of alm-example, which allows us to create multiple opbi in different namespace


Daniel-Fan (approved):
/lgtm

PR #1304, opened 2023-06-06
CS operator handles scoping changes

YCShen1010:
For issue: https://github.ibm.com/ibmprivatecloud/roadmap/issues/58293

Moving cluster type check into reconcile master CS CR
This PR partially fix for the scenario that NSS CR accidentally remove namespace from scope which will cause the cs-operator stuck in this initialization stage (0/1 pod not ready), and error msg ` Failed to verify cluster type unable to get: xxx/ibm-cpp-config because of unknown namespace for the cache`. This is because cs operator does not have permissions copy the ibm-cpp-config configmap to the namespace as it has been removed from NSS CR.

### DIFF
--- a/controllers/commonservice_controller.go
+++ b/controllers/commonservice_controller.go
@@ -173,6 +173,25 @@ func (r *CommonServiceReconciler) ReconcileMasterCR(ctx context.Context, instanc
 		}
 	}
 
+	typeCorrect, err := r.Bootstrap.CheckClusterType(util.GetServicesNamespace(r.Reader))
+	if err != nil {
+		klog.Errorf("Failed to verify cluster type  %v", err)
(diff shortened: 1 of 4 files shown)


controllers/commonservice_controller.go
```diff
@@ -181,6 +182,21 @@ func (r *CommonServiceReconciler) ReconcileMasterCR(ctx context.Context, instanc
 		}
 	}
 
+	for {
```
Daniel-Fan, line 185: I am wondering if we could remove the for loop here.

The main reason if that we would like to return error in CR reconciliation. Subsequently, it will trigger the next reconciliation, and shows `Failed` status in CommonService CR.

Daniel-Fan (approved):
/lgtm

PR #1315, opened 2023-06-14
[4.1] check operatorNS and servicesNS in cs cr

qpdpQ:
for ticket: https://github.ibm.com/IBMPrivateCloud/roadmap/issues/59007
In the upgrade, if CommonService CRD does not contain  `x-kubernetes-preserve-unknown-fields: true`, operatorNS and serviceNS will be removed by k8s, so we need to double check it.

### DIFF
--- a/cp3pt0-deployment/setup_tenant.sh
+++ b/cp3pt0-deployment/setup_tenant.sh
@@ -492,7 +492,7 @@ EOF
 }
 
 function configure_cs_kind() {
-    local retries=5
+    local retries=10
     local delay=30
 
     title "Configuring CommonService CR in $OPERATOR_NS..."
@@ -531,8 +531,9 @@ EOF
             else
                 operatorNSinCR=$(${OC} get commonservice common-service -n ${OPERATOR_NS} -o yaml | yq '.spec.operatorNamespace')
                 servicesNSinCR=$(${OC} get commonservice common-service -n ${OPERATOR_NS} -o yaml | yq '.spec.servicesNamespace')
-                info "OperatorNamespace in CommonService cr should be ${OPERATOR_NS}, but it is ${operatorNSinCR}"
-                info "ServicesNamespace in CommonService cr should be ${SERVICES_NS}, but it is ${servicesNSinCR}"
+                warning "OperatorNamespace in CommonService cr should be ${OPERATOR_NS}, but it is ${operatorNSinCR}"
+                warning "ServicesNamespace in CommonService cr should be ${SERVICES_NS}, but it is ${servicesNSinCR}"
+                retries=$((retries-1))
             fi
         else
             warning "Failed to patch CommonService CR in ${OPERATOR_NS}, retry it in ${delay} seconds..."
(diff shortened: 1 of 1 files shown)


cp3pt0-deployment/setup_tenant.sh
```diff
@@ -464,8 +464,12 @@ EOF
     
         # Check if the patch was successful
         if [[ $? -eq 0 ]]; then
-            success "Successfully patched CommonService CR in ${OPERATOR_NS}"
-            break
+            checkOperatorNS=$(${OC} get commonservice common-service -n ${OPERATOR_NS} -o yaml | yq '.spec.operatorNamespace=="'${OPERATOR_NS}'"')
+            checkServicesNS=$(${OC} get commonservice common-service -n ${OPERATOR_NS} -o yaml | yq '.servicesNamespace=="'${SERVICES_NS}'"')
```
Daniel-Fan, line 468:
```suggestion
            checkServicesNS=$(${OC} get commonservice common-service -n ${OPERATOR_NS} -o yaml | yq '.spec.servicesNamespace=="'${SERVICES_NS}'"')
```

Daniel-Fan (changes requested):

cp3pt0-deployment/setup_tenant.sh
Daniel-Fan: Could we cover some failure cases on tests?

Fr example, the `servicesNamespace` in the CS CR is not desired or this field is pruned by k8s APIServer(Manually edit CS CRD to remove `preserve-unknown-fields` marker and remove `servicesNamespace` field).

cp3pt0-deployment/setup_tenant.sh
```diff
@@ -523,8 +523,17 @@ EOF
...
+            checkOperatorNS=$(${OC} get commonservice common-service -n ${OPERATOR_NS} -o yaml | yq '.spec.operatorNamespace=="'${OPERATOR_NS}'"')
+            checkServicesNS=$(${OC} get commonservice common-service -n ${OPERATOR_NS} -o yaml | yq '.spec.servicesNamespace=="'${SERVICES_NS}'"')
+            if [ $checkOperatorNS ] && [ $checkServicesNS ]; then
+                success "Successfully patched CommonService CR in ${OPERATOR_NS}"
+                break
+            else
+                operatorNSinCR=$(${OC} get commonservice common-service -n ${OPERATOR_NS} -o yaml | yq '.spec.operatorNamespace')
+                servicesNSinCR=$(${OC} get commonservice common-service -n ${OPERATOR_NS} -o yaml | yq '.spec.servicesNamespace')
+                info "OperatorNamespace in CommonService cr should be ${OPERATOR_NS}, but it is ${operatorNSinCR}"
+                info "ServicesNamespace in CommonService cr should be ${SERVICES_NS}, but it is ${servicesNSinCR}"
```
Daniel-Fan, line 535: I am wondering that should we raise `warning` msg here, and subtract one retry times.

As a result, it will retry several times, and error out the script when retry is zero.

Daniel-Fan (approved):
/lgtm

PR #1316, opened 2023-06-14
fix bug in conversion script

bluzarraga:
variable names were mixed up causing script to fail, ran new version and it is working now


Daniel-Fan (approved):
/lgtm

PR #1318, opened 2023-06-21
creating namespaces in pre-req

YCShen1010:
Since we do the cert manager check in `pre-req` function, we need to create all namespaces before `check_cert_manager` for those which are first time passed into the script, otherwise, there will be `No resources found in operator_namespace` and failed in `[✘] Missing a cert-manager`
fix for issue: https://github.ibm.com/IBMPrivateCloud/roadmap/issues/59207

test result：
```
# Start to validate necessary functional prerequisite... 
[✔] oc command available
[✔] Channel is valid
[✔] Profile size is valid.
# Creating namespace cloudpak-control-test
namespace/cloudpak-control-test created
[INFO] Waiting for project cloudpak-control-test to be created
[✔] Project cloudpak-control-test is created

# Creating namespace cloudpak-data-test
namespace/cloudpak-data-testcreated
[INFO] Waiting for project cloudpak-data-test to be created
[✔] Project cloudpak-data-test is created

#  Checking whether Cert Manager exist...

#  Checking IBMLicensing...

```

### DIFF
--- a/cp3pt0-deployment/setup_tenant.sh
+++ b/cp3pt0-deployment/setup_tenant.sh
@@ -46,6 +46,7 @@ function main() {
     trap cleanup_log EXIT
     pre_req
     setup_topology
+    check_singleton_service
     setup_nss
     install_cs_operator
 }
@@ -142,6 +143,7 @@ function print_usage() {
 }
 
 function pre_req() {
+    title "Start to validate necessary functional prerequisite... "
     # Check the value of DEBUG
     if [[ "$DEBUG" != "1" && "$DEBUG" != "0" ]]; then
         error "Invalid value for DEBUG. Expected 0 or 1."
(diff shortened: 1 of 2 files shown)


cp3pt0-deployment/setup_tenant.sh
```diff
@@ -157,6 +157,7 @@ function pre_req() {
         success "oc command logged in as ${user}"
     fi
 
+    create_ns_list
```
Daniel-Fan, line 160: I believe we could try to move following functions out of `pre_req` function
- `create_ns_list`
- `check_cert_manager`
- `check_licensing`

IMHO, `pre_req` function could be used for validate scripts parameters only.

And we have a a new function to validate necessary functional prerequisite, after we `setup_tepology`

- `pre_req`
- `setup_topology`
-  `check_singleton_service`
    - `check_cert_manager`
    - `check_licensing`

Daniel-Fan (changes requested):
Could we use local variable to store namespace and service name?
```
function check_cert_manager(){
    local service_name=$1    
    local namespace=$2
    csv_count=`$OC get csv -n "$namespace" | grep "$service_name" | wc -l`
    if [[ $csv_count == 0 ]]; then
        error "Missing a cert-manager"
    fi
```

cp3pt0-deployment/setup_tenant.sh
```diff
@@ -142,6 +143,7 @@ function print_usage() {
 }
 
 function pre_req() {
+    title "Start to validate necessary functional prerequisite... "
```
Daniel-Fan, line 146: Could we update the title here that `pre_req` is used to validate the parameters passed into script.

functional prerequisite is something like certificate management functionality.

Daniel-Fan (approved):
/lgtm

PR #1322, opened 2023-06-22
[4.1] Make podrestart label more accurate

qpdpQ:
cherry pick from:https://github.com/IBM/ibm-cert-manager-operator/pull/340

If two certificates are mounted by one pod, and those two certificates are refreshed in the same minute, The second certificate restart will not trigger a pod restart because the time labels are the same.


Daniel-Fan (approved):
/lgtm

PR #1323, opened 2023-06-22
Update golang image to bullseye base

Daniel-Fan:
golang:1.20.5 switched to a new base os debian `bookworm`, use `1.20.5-bullseye` to keep compatibility


YCShen1010 (approved):
/lgtm

PR #1308, opened 2023-06-08
z specific changes to preload process

bluzarraga:
Implementing the workaround for cpd quality repo issue 9990. 

Workaround is as follows:

- Scale down mongo operator in ibm-common-services to 0
-  edit icp-mongodb cm in ibm-common-services . Change the net.ssl.mode value to preferSSL from requireSSL
-  restart mongo pods in ibm-common-services one by one starting with icp-mongo-0. Delete the pod, let it come all the way ready, then move on to the next pod
-  clone the github repo for scripts branch, edit the preload_data.sh to remove the ssl parameters from the backup job definition [here](https://github.com/IBM/ibm-common-service-operator/blob/scripts/preload_data.sh#L317) and the restore job definition [here](https://github.com/IBM/ibm-common-service-operator/blob/scripts/preload_data.sh#L490)
-  remove the metrics container from the statefulset yaml (remove from https://github.com/IBM/ibm-common-service-operator/blob/scripts/preload_data.sh#L1554 through https://github.com/IBM/ibm-common-service-operator/blob/scripts/preload_data.sh#L1674)
-  save edited file
-  delete the dummy mongo statefulset in the services namespace (zen in the cluster from this issue)
-  rerun the edited preload_data.sh script with the same parameters
-  after completion, revert change to icp-mongodb configmap in ibm-common-services (aka the original cs namespace) to set net.ssl.mode to requireSSL
-  delete mongo pods one by one again same as before
-  scale mongo operator back to 1

Testing:
- setup cs on z cluster
- run preload_data.sh
- verify script completes and data is transferred into new namespace

PR #1330, opened 2023-06-23
Update backup and restore mongo script to work on ROKS and Z

bluzarraga:
Includes changes from https://github.com/IBM/ibm-common-service-operator/pull/1270 because they have not been merged in yet and the changes for Z needed some of the changes for ROKS.
Copy paste text from that pr:
>Previously, because the script created it's own storage class, the backup_restore_mongo.sh script did not work on ROKS. The changes in this pr fix that by reusing the storage class the original mongo install already uses instead of creating a new one.
Other changes include changing where the file is located as well as the necessary support files to make airgap process easier. Also updated preload_data.sh to remove any instance of oc and replace it with ${OC}.
Testing:
    - get a roks cluster with common services installed
        - really just need iam and mongo
        - needs LDAP setup
    - run conversion script to split two cp namespaces to use two different cs instances
    - verify cannot login with ldap to the new instance
    - run backup_restore_mongo.sh to restore mongo data into new cs instance
    - log in with LDAP to the new cs instance's cp-console
I was able to get this to work using a SERT cluster

For Z changes:
The exact same problem that the preload script had see https://github.com/IBM/ibm-common-service-operator/pull/1308 for more details. 
Essentially, mongo on Z does not accept `--ssl` flags and will fail a job that tries to use them which both the backup and restore jobs did. These changes are the same made for preload just applied to this different situation. Biggest change is scaling mongo operator, replacing icp-mongodb cm, and restarting mongo pods in both the original and target namespace in this script instead of just the original namespace for the preload script. I have gotten this to run on a SERT Z cluster so I think this should be working.

Testing:
- setup cs on z cluster. Create LDAP 
- run conversion script so now there should be mongo and iam running in both original and target namespace
- wait for pods to come ready
- run backup_restore_mongo.sh --bns <original ns> --rns <target ns> -b -r -c
- verify script completes
- verify can login to target namespace using ldap users from original namespace


Daniel-Fan (approved):
/lgtm

PR #1331, opened 2023-06-23
Cleanup scritp enhancement

qpdpQ:
1. Allow user to input multiple namespaces
2. update common-service-maps after a tenant is uninstalled
3. when there is no `map-to-common-service-namespace` in the common-service-maps , cleanup control ns

### DIFF
--- a/cp3pt0-deployment/uninstall_tenant.sh
+++ b/cp3pt0-deployment/uninstall_tenant.sh
@@ -366,7 +366,9 @@ function cleanup_cs_control() {
         cleanup_secretshare $CONTROL_NS ""
         # cleanup crossplane    
         cleanup_crossplane
-
+        # delete common-service-maps 
+        ${OC} delete configmap common-service-maps -n kube-public
+        # delete namespace
         ${OC} delete --ignore-not-found ns "$CONTROL_NS" --timeout=30s
         if [ $? -ne 0 ] || [ $FORCE_DELETE -eq 1 ]; then
             warning "Failed to delete namespace $CONTROL_NS, force deleting remaining resources..."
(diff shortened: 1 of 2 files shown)


Daniel-Fan:
@qpdpQ 
When we delete `cs-control` namespace, we are missing to delete NamespaceScope CR `common-service`. Here is the [template](https://github.com/IBM/ibm-common-service-operator/blob/7ba99773c8cc7a3f16b3d2b61af3cee442005a85/controllers/constant/namespace.go#L109)
```
function cleanup_NamespaceScope() {
    local namespace=$1
    ${OC} delete namespacescope odlm-scope-managedby-odlm nss-odlm-scope nss-managedby-odlm -n ${namespace} --ignore-not-found
}
```

cp3pt0-deployment/uninstall_tenant.sh
```diff
@@ -265,10 +300,84 @@ function delete_tenant_ns() {
...
+data:
+  common-service-maps.yaml: |
+${yaml}
+EOF
+)"
+    echo "$object" | oc apply -f -
+}
+
+# check if we need to cleanup contorl namespace and clean it
+function cleanup_cs_contorl() {
```
bluzarraga, line 352: nitpick but a typo in the function name

bluzarraga (commented):
When deleting one tenant from my cluster (where there were two), the script cleaned up my cs-control namespace as well. @qpdpQ and I have slacked about it but will add to this pr too

bluzarraga:
Tested on amd64, successfully uninstalled multiple tenants, cleaned up control ns when it should have and left it alone when a tenant still existed, deletes csmaps when no more tenants are present. I think this looks good, thanks @qpdpQ

PR #1332, opened 2023-06-26
fix bug in conversion script

bluzarraga:
Sync up scripts-dev conversion specific change into scripts branch


Daniel-Fan (approved):
/lgtm

PR #1335, opened 2023-06-28
updated go version to 1.20.5

bitscuit:


Dockerfile
```diff
@@ -1,5 +1,5 @@
 # Build the manager binary
-FROM golang:1.20.4 as builder
+FROM golang:1.20.5 as builder
```
Daniel-Fan, line 2:
```suggestion
FROM golang:1.20.5-bullseye as builder
```

Daniel-Fan: Slack discussion: https://ibm-cloudplatform.slack.com/archives/GUFCP8AH5/p1687460912765779

Daniel-Fan (changes requested):

Daniel-Fan (approved):
/lgtm

PR #1349, opened 2023-07-07
migrate tenant adoption on singleton enhancement

Daniel-Fan:
- Add YQ variable for binary path
- Turn off Preview Mode by default
- Remove quotation mark for `arguments` when passing it into `${BASE_DIR}/setup_singleton.sh "--operator-namespace" "$OPERATOR_NS" "-c" "$CHANNEL" "--cert-manager-source" "$CERT_MANAGER_SOURCE" "--licensing-source" "$LICENSING_SOURCE" "--license-accept" $arguments`.
   - Otherwise, `"$arguments"` will be phased as a long parameter `--enable-licensing -licensingNs cs-control` as `wildcard` instead of three dedicated and valid parameters.
- Only enable Licensing Migration when it is specified in `migrate_tenant.sh` parameter
- Add wait message for licensing instance to be present
- empty directory `/tmp/preview` for scripts other than `setup_tenant.sh`

### DIFF
--- a/cp3pt0-deployment/common/delegate_cp2_cert_manager.sh
+++ b/cp3pt0-deployment/common/delegate_cp2_cert_manager.sh
@@ -14,7 +14,7 @@ OC=oc
 YQ=yq
 CONTROL_NS=""
 DEBUG=0
-PREVIEW_MODE=1
+PREVIEW_MODE=0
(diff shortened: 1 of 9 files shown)


YCShen1010 (approved):
/lgtm

PR #1370, opened 2023-07-17
don't copy oidc credential

bluzarraga:
A couple issues involving data migration with the secret copied over. This is the leading option to resolve them but there is still some discussion so will mark with a hold for now


bitscuit (approved):
/lgtm

Daniel-Fan:
Remove holding label

PR #1371, opened 2023-07-17
change channel along with subscription

bluzarraga:
For issue 57770, make sure the channel matches the original cs operator subscription value as well as the catalog source. This commit also changes the file in place instead of deleting the existing subscription by removing certain metadata fields and then applying the file on top of the existing subscription.


bitscuit (approved):
/lgtm

PR #1373, opened 2023-07-18
Add NSS restart label when patching the deployment

Daniel-Fan:
This label should be patched together with `WATCH_NAMESPACE` ENV variable in common service operator deployment

Screenshot 2023-07-18 at 2 01 07 AM


YCShen1010 (approved):
/lgtm

PR #1387, opened 2023-07-26
Update operand requests' registryNamespace value to reflect new cs instance namespace

bluzarraga:
Ran into an opportunity for enhancement while testing conversion script where an operand request was not updated to point to the new cs instance namespace for registryNamespace. This prevented ODLM in the new instance from reconciling the operandrequest as it tried to access the operandregistry in a namespace it did not have access to. Manually changing the registryNamespace value resolved the issue but I think this can be included in the conversion script. I have tested this on two different environments now (both amd64) successfully following this methodology:
- setup shared instance cluster with multiple cloudpak namespaces with operand requests
- convert cluster using this version of conversion script
- verify that operandrequests are updated correctly during script runtime
- verify script completes
- verify each operandrequest is reconciled and services deployed

### DIFF
--- a/convert_to_multi_instance.sh
+++ b/convert_to_multi_instance.sh
@@ -838,9 +838,9 @@ function update_opreqs(){
                 ${YQ} -i 'del(.metadata.uid)' tmp.yaml
                 ${YQ} -i 'del(.metadata.generation)' tmp.yaml
                 ${YQ} -i 'del(.metadata.managedFields)' tmp.yaml
-                ${YQ} -i '.spec.requests[0].registryNamespace = "'${csns}'"' tmp.yaml    
+                ${YQ} -i '.spec.requests[0].registryNamespace = "ibm-common-services"' tmp.yaml    
                 ${OC} apply -n $ns -f tmp.yaml || error "Failed to update registryNamespace value for operand request $opreq in namespace $ns."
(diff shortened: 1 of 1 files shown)


convert_to_multi_instance.sh
```diff
@@ -817,6 +818,36 @@ function cleanup_webhook() {
...
+    title "Updating Operand Requests to use Operand Registry in new CS namespace"
+    #check map to namespaces, get list of requested from ns from there
+    #update opreq in map to first
+    #update opreq in list of namespace from first line
+    for csns in $map_to_cs_ns
+    do
+        local namespaces=$(${OC} get cm common-service-maps -o yaml -n kube-public | $YQ '.data[]' | $YQ '.namespaceMapping[] | select(.map-to-common-service-namespace == "'$csns'").requested-from-namespace' | awk '{print $2}' | tr '\n' ' ')
+        namespaces="$namespaces $csns"
+        for ns in $namespaces
+        do
```
bitscuit, line 831: The loops are iterating over all namespaces in cs-maps. Why not avoid the loops entirely, and just update all OperandRequests to have `ibm-common-services` registryNamespace. It would be equivalent and faster

bluzarraga: The map_to_cs_ns variable only includes the map-to values that the topology check noted were changed so it is actually only the subset of namespaces used in this run of the conversion script. See line 664 and 665

PR #1418, opened 2023-08-16
restart the webhook pod after deleting config

bluzarraga:
Conditionally restart the webhook pod in the control namespace after deleting the webhook config. This is mostly to deal with subsequent runs as this is not a problem on the first run. Tested myself on amd64 and it is working as expected. 
Test:
- setup env with 2 cp namespaces using ibm-common-services
- convert once by moving one namespace to it's own tenant
- convert a second time by moving the remaining cp namespace to its own tenant
- verify webhook pod is deleted (look for info line "Webhook pod in control namespace, restarting.")
- verify the operandrequests' registryNamespace values have been changed from ibm-common-services to the new registry namespace for the new tenants

### DIFF
--- a/convert_to_multi_instance.sh
+++ b/convert_to_multi_instance.sh
@@ -28,6 +28,7 @@ catalog_source=
 requested_ns=
 map_to_cs_ns=
 master_ns=$1
+control_ns=
 cm_name="common-service-maps"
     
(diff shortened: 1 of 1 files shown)


Daniel-Fan (approved):
/lgtm

PR #1420, opened 2023-08-17
check for WaitForFirstConsumer bind mode

bluzarraga:
While creating a solution for a customer to change the storage class for mongo, we ran into an issue where the storage class had a `volumeBindingMode` value of `WaitForFirstConsumer` hanging the script while waiting for the `cs-mongodump` pvc to bind. This means that the storage class will not bind a pv to a pvc unless the pvc is attached to a running pod. To fix this, I added a conditional check to see if this specific `volumeBindingMode` is set. If it is, we do not wait for the pvc to bind and instead move on. Once the mongo backup pod is created, it will use the `cs-mongodump` pvc which then allows the pvc to bind to a pv provisioned by the storage class and the process continues as expected. The scripts for changing storage class for mongo were based on the preload_data script (the function `createdumppvc` is explicitly used) so I needed to update the preload_data script to make sure we support a wider variety of storage class options.

Test:
- install common services using a storage class that has `WaitForFirstConsumer` as its `volumeBindingMode`
- run preload data
- verify this info message is output: `StorageClass waits for pod to claim PVC, skipping wait for binding.`
- verify script completes successfully

### DIFF
--- a/preload_data.sh
+++ b/preload_data.sh
@@ -285,13 +285,18 @@ EOF
 
   ${OC} apply -f $TEMPFILE
 
-  status=$(${OC} get pvc cs-mongodump --no-headers | awk '{print $2}')
-  while [[ "$status" != "Bound" ]]
-  do
-    info "Waiting for pvc cs-mongodump to bind"
-    sleep 10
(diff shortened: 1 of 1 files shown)


Daniel-Fan (approved):
/lgtm

PR #1423, opened 2023-08-19
Support picking default channel from bundle

Daniel-Fan:
### Background
It is used to cover a scenario when user want to install operators from its default channel `.status.defaultChannel` set in operator packagemanifest.
Screenshot 2023-08-18 at 10 58 57 PM

And it is possible that different operator has different default channel. For example,
- User would like to install `ibm-common-service-operator` from its default `v4.3` channel.
- User would like to install `ibm-namespace-scope-operator` from its default `v4.2` channel.

### Action
Script will support to set operator subscription channel to null. e.g., `.spec.channel: null`.
When channel is set to `null` in subscription, OLM will automatically pick up the default channel in operator PackageManifest.
Screenshot 2023-08-18 at 11 09 27 PM

### How to test
1. Deploy CatalogSource for Singleton Service(Cert-Manager and Licensing) and CS Services
2. Execute the script with channel specified to `null`
  - `./cp3pt0-deployment/setup_singleton.sh --license-accept --enable-licensing -c null`
  - `./cp3pt0-deployment/setup_tenant.sh --operator-namespace cp4ba-operator --services-namespace cp4ba-service --license-accept -c null`
3. The script should run successfully and install the latest version.

NOTE: `migrate_tenant.sh` script also supports picking up default channel from packagemanifest
  - `./cp3pt0-deployment/migrate_tenant.sh --license-accept --enable-licensing --operator-namespace ibm-common-services -c null`

### DIFF
--- a/cp3pt0-deployment/common/utils.sh
+++ b/cp3pt0-deployment/common/utils.sh
@@ -449,6 +449,14 @@ function wait_for_operator_upgrade() {
     local success_message="Operator ${package_name} is upgraded to latest version in channel ${channel}"
     local error_message="Timeout after ${total_time_mins} minutes waiting for operator ${package_name} to be upgraded"
 
+    # if channel is not set, skip the wait
+    if [[ "${channel}" == "null" ]]; then
+        info "${wait_message}"
+        sleep ${sleep_time}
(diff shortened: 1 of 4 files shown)


YCShen1010 (approved):
/lgtm

PR #1436, opened 2023-08-28
updated descriptions in CRD fields

bitscuit:
Can run `make manifests` and `make bundle-manifests` from now on to properly generate manifests/bundle (except need to revert the preserve-unknown-fields` change that happens)

### DIFF
--- a/api/v3/commonservice_types.go
+++ b/api/v3/commonservice_types.go
@@ -51,24 +51,50 @@ type ServiceConfig struct {
 
 // CommonServiceSpec defines the desired state of CommonService
 type CommonServiceSpec struct {
-	// INSERT ADDITIONAL SPEC FIELDS - desired state of cluster
-	// Important: Run "make" to regenerate code after modifying this file
-
-	Features            *Features            `json:"features,omitempty"`
+	Features *Features `json:"features,omitempty"`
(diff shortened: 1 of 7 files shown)


api/v3/commonservice_types.go
```diff
@@ -127,36 +153,63 @@ type CatalogName string
...
+	OperatorDeployed bool `json:"operatorDeployed,omitempty"`
+	// ServicesNamespace is the namespace where the foundational services'
+	// operands will be created in.
+	ServicesNamespace ServicesNamespace `json:"servicesNamespace,omitempty"`
+	// ServicesDeployed indicates whether the OperandConfig has been created or
+	// not.
+	ServicesDeployed bool `json:"servicesDeployed,omitempty"`
+	// Configurable indicates whether this CommonService CR is the one that can
+	// that can be used to configure the foundational services' installer. Other
+	// CommonService CRs configurations will not take effect, except for sizing
```
Daniel-Fan, line 186:
```suggestion
	// Configurable indicates whether this CommonService CR is the one
	// that can be used to configure the foundational services' installer. Other
	// CommonService CRs configurations will not take effect, except for sizing
```

Daniel-Fan (changes requested):

Daniel-Fan (approved):
/lgtm

PR #1446, opened 2023-09-02
Verify yq version in "convert_to_multi_instance.sh" script

### LINKED ISSUE #1445: Convert to Multi-Instance Script - Bug with yq
Following documentation here: https://www.ibm.com/docs/en/cloud-paks/1.0?topic=icpfsimn-converting-single-namespace-installation-multiple-namespaces-installation-tech-preview#example

[Convert to multi instance script](https://github.com/IBM/ibm-common-service-operator/blob/scripts/convert_to_multi_instance.sh) fails due to an old version of *yq*. The script does not adequately check the user's *yq* version before proceeding and gives a poor error message as a result. In my case, I got this error with *yq* version 3.3.0. This issue was resolved after upgrading to the latest version of *yq*.

Line causing the issue:`return_value=$("${OC}" get configmap -n kube-public -o yaml ${cm_name} | yq '.data' | grep controlNamespace: > /dev/null || echo failed)`

Error message: `2023/09/01 18:32:58 unknown command ".data" for "yq"`

tylerstanczak:
Set and verify minimum *yq* version at 4.18.1 as this is the earliest version that supports the *yq* syntax in the script.

Prior to yq 4.18.1, example usage is as follows: 
`echo "yamlData: value" | yq e .yamlData`

From *yq* 4.18.1 and onwards "e" (short for eval) became the default command. This matches the usage in the script where "e" (eval) has been assumed to be the default command.

*yq* 4.18.1 release notes: https://github.com/mikefarah/yq/releases/tag/v4.18.1

Fixes #1445

### DIFF
--- a/convert_to_multi_instance.sh
+++ b/convert_to_multi_instance.sh
@@ -51,6 +51,16 @@ function prereq() {
     which "${OC}" || error "Missing oc CLI"
     which "${YQ}" || error "Missing yq"
     
+    # Verify that yq is at least version 4.18.1
+    yq_minimum_version=4.18.1
+    # Get the user's yq version and remove any leading "v" character if present
+    yq_version=$("${YQ}" --version | awk '{print $NF}' | sed 's/^v//')
+
+    # Compare yq versions using sort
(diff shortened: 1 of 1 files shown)


Daniel-Fan (approved):
/lgtm

Thank you @tylerstanczak for the PR!
It looks good to me!

PR #1448, opened 2023-09-05
delete cs ca cert secret

YCShen1010:
for issue: https://github.ibm.com/ibmprivatecloud/roadmap/issues/58293
Testing image: quay.io/yuchen_shen/cs_operator:scope_handling

When the first time to deploy CP3 in the cluster, by default servicesNamespace is the same as operatorNamespace, `cs-ca-certificate-secret` secret will be generated in that tmp serivcesNamespace.
After servicesNamespace is changed, this resource should be cleaned up properly and create under the new serivcesNamespace

#### Verify

- Using setup_tenant script to deploy a CP3 (could choose Topology B with 2 different tethered ns, and same ns for services ns and operator ns)
- apply the testing image, change imagePullPolicy to `Always`
- run setuo_tenant.sh again with a new ns for ServicesNamespace
- Observation results: `cs-ca-certificate-secret` secret in original services ns are deleted and create a new one under new services ns.

### DIFF
--- a/controllers/bootstrap/init.go
+++ b/controllers/bootstrap/init.go
@@ -1295,7 +1295,7 @@ func (b *Bootstrap) ConfigCertManagerOperandManagedByOperator(ctx context.Contex
 		client.MatchingLabels(
 			map[string]string{constant.CsManagedLabel: "true"}),
 	}
-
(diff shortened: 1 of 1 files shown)


Daniel-Fan (approved):
/lgtm

PR #1453, opened 2023-09-05
fixed make manifests to have good CRD

bitscuit:
- before it would not put preserve-unknown-fields marker in the ServiceConfigSpec


Daniel-Fan (approved):
/lgtm

bitscuit:
Will need to investigate a bit further to properly implement this. I tried the following combinations, but one of the following failed depending on the code: `make manifests`, `make generate`, `make check`.

The following will have `make generate` fail, but weirdly enough the deepcopy code is generated for the ExtensionWithMarker type

```
// +kubebuilder:pruning:PreserveUnknownFields
type ExtensionWithMarker runtime.RawExtension

type ServiceConfig struct {
	Name               string                         `json:"name"`
	Spec               map[string]ExtensionWithMarker `json:"spec"`
	ManagementStrategy string                         `json:"managementStrategy,omitempty"`
}
```

The following will have `make check` fail because if you look in the deepcopy generated code, there is no deepcopy function for ExtensionWithMarker

```
// +kubebuilder:pruning:PreserveUnknownFields
type ExtensionWithMarker *runtime.RawExtension

type ServiceConfig struct {
	Name               string                         `json:"name"`
	Spec               map[string]ExtensionWithMarker `json:"spec"`
	ManagementStrategy string                         `json:"managementStrategy,omitempty"`
}
```

The follow will have `make manifests` fail because there is missing json tag next to `runtime.RawExtension`, and we definitely do not want a json tag there because it changes the CRD

```
// +kubebuilder:pruning:PreserveUnknownFields
type ExtensionWithMarker struct {
	runtime.RawExtension
}

type ServiceConfig struct {
	Name               string                         `json:"name"`
	Spec               map[string]ExtensionWithMarker `json:"spec"`
	ManagementStrategy string                         `json:"managementStrategy,omitempty"`
}
```

bitscuit:
Finally fixed, ready for review. Hold merge for 4.3 since not needed for 4.2

Daniel-Fan (approved):
/lgtm

PR #1454, opened 2023-09-06
Fix make run by disabling webhook server

Jamstah:
make run currently does not work because there is no webhook certificate in the tmp dir. This stops the webhook server from starting when using make run, because when testing the operator locally the k8s api server won't be configured to point to your local server anyway.


Daniel-Fan (approved):
/lgtm

Thank you for the PR @Jamstah !

It looks good to me.

PR #1459, opened 2023-09-06
Adding watch verb to edb-license-role

senorhesles:
https://github.com/kubernetes/kubectl/issues/1411

The kubectl binary was updated the `edb-postgres-license-provider` image; this new version of kubectl lists and watches resources after deletion; before it just listed them. So we need additional rbac for the role.


Daniel-Fan (approved):
/lgtm

PR #1462, opened 2023-09-06
preserve new spec when it is specified

Daniel-Fan:
This is an experiment of configuring arbitrary attributes for OperandConfig for CS CR.


Daniel-Fan:
Close as this is not needed

PR #1467, opened 2023-09-07
Add necessary files to support zen 5 backup and restore

bluzarraga:
These files likely need to be added to scripts-dev branch as well to be included in the case bundle. The images need to be updated to reflect the final image used for cpfs-utils


Daniel-Fan (approved):
/lgtm

PR #1468, opened 2023-09-08
enable private CatalogSource in setup tenant script

YCShen1010:
issue enhancement: https://github.ibm.com/IBMPrivateCloud/roadmap/issues/59447

- Set private CatalogSource flag in setup tenant script
- Check and correct cs operator catalogsource in each scoped namespace

### DIFF
--- a/cp3pt0-deployment/migrate_tenant.sh
+++ b/cp3pt0-deployment/migrate_tenant.sh
@@ -96,9 +96,9 @@ function main() {
             if [ $ENABLE_PRIVATE_CATALOG -eq 1 ]; then
                 op_source_ns=$ns
             fi
+            validate_operator_catalogsource $pm $ns $op_source $op_source_ns $CHANNEL op_source op_source_ns
+            update_operator $pm $ns $CHANNEL $op_source $op_source_ns $INSTALL_MODE
         fi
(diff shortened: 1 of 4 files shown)


cp3pt0-deployment/migrate_tenant.sh
```diff
@@ -86,16 +85,20 @@ function main() {
...
+        local sub_name=$(${OC} get subscription.operators.coreos.com -n ${ns} -l operators.coreos.com/${pm}.${ns}='' --no-headers | awk '{print $1}')
+        if [ ! -z "$sub_name" ]; then
+            op_source=$SOURCE
+            op_source_ns=$SOURCE_NS
+            if [ $ENABLE_PRIVATE_CATALOG -eq 1 ]; then
+                op_source_ns=$ns
+            fi
         fi
+        validate_operator_catalogsource $pm $ns $op_source $op_source_ns $CHANNEL op_source op_source_ns
+        update_operator $pm $ns $CHANNEL $op_source $op_source_ns $INSTALL_MODE
```
Daniel-Fan, line 101: Could we move those two functions into above if condition? It will validate catalog and update operator when subscription exists.

Daniel-Fan (changes requested):

Daniel-Fan (approved):
/lgtm

PR #1484, opened 2023-09-15
LSR migration fix for labelling PV on ROKS

geforcelive:


Daniel-Fan:
Hi @geforcelive 
May I ask that is this still needed?

geforcelive:
Hey @Daniel-Fan I think it's not needed. As far as I remember LSR migration should be already properly handled.

Daniel-Fan:
Thanks @geforcelive 
Yes, LSR migration is already handled, I will close the PR.

PR #1490, opened 2023-09-19
Add common label to operator Deployment/Pod

pgodowski:
Resolves https://github.ibm.com/IBMPrivateCloud/roadmap/issues/59790


bluzarraga (changes requested):
We can't have spaces in a label or it will be invalid. We can use `.`, `_`, or `-` instead

bluzarraga (approved):

PR #1498, opened 2023-09-20
update label licensing to take namespace parameter

bluzarraga:
necessary to match what is in the docs and scripts branch


Daniel-Fan (approved):
/lgtm

PR #1509, opened 2023-09-27
add new tags to scripts dev

bluzarraga:
match scripts-dev to scripts to actually backup and restore new items


bitscuit (approved):
/lgtm

PR #1515, opened 2023-09-29
mapping between CS operator and Namespace Scope operator channel

YCShen1010:
issue: https://github.ibm.com/IBMPrivateCloud/roadmap/issues/60615

- If the user provides `v4.2` or a higher channel version when running the script, it will choose the NSS operator from the `v4.2` channel and select the CS operator from the `v4.2` channel or any version higher.
- Conversely, if the user specifies the parameter -c with `v4.1` or a lower version when executing the script, it will select the NSS operator from the same channel as the CS operator channel.

### DIFF
--- a/cp3pt0-deployment/setup_tenant.sh
+++ b/cp3pt0-deployment/setup_tenant.sh
@@ -387,7 +387,7 @@ function install_nss() {
     local cs_channel="${CHANNEL#v}"
     local nss_channel="${NSS_CHANNEL#v}"
 
-    # Check if CHANNEL_NUMERIC is greater than or equal to 4.2, if yes, use NSS channel as v4.2
+    # Check if cs_channel is less than nss_channel, if yes, let NSS channel be the same as CS channel
     if (( $(echo "$cs_channel < $nss_channel" | bc -l) )); then
         NSS_CHANNEL="$CHANNEL"
     fi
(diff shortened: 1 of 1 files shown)


cp3pt0-deployment/setup_tenant.sh
```diff
@@ -102,6 +103,7 @@ function parse_arguments() {
         -c | --channel)
             shift
             CHANNEL=$1
+            NSS_CHANNEL=$1
```
Daniel-Fan, line 106: I believe we could remove this assignment. Check below comment for details.

cp3pt0-deployment/setup_tenant.sh
```diff
@@ -382,15 +384,22 @@ function check_singleton_service() {
 function install_nss() {
     title "Checking whether Namespace Scope operator exist..."
 
+    # Extract the numeric part from the CHANNEL variable using parameter expansion
+    local channel_numeric="${CHANNEL#v}"
+    # Check if CHANNEL_NUMERIC is greater than or equal to 4.2, if yes, use NSS channel as v4.2
+    if (( $(echo "$channel_numeric >= 4.2" | bc -l) )); then
+        NSS_CHANNEL="v4.2"
+    fi
```
Daniel-Fan, line 392: I am wondering if we could avoid hardcode `4.2` and `v4.2` as much as possible?

This is what we have known:
- We already know the latest channel for NSS is `v4.2`, and this is the default value we assign to `NSS_CHANNEL` at the beginning of the script.
- We wanna only update `NSS_CHANNEL` when it has to be updated.
  - The only scenario that we will need to update the `NSS_CHANNEL` is when user using this script to install a old version of CPFS. e.g., `v4.1`.

As a result, we do not need to update `NSS_CHANNEL` when user installs CPFS v4.3+.

```suggestion
    if (( $(echo "$cs_channel < #nss_channel" | bc -l) )); then
        NSS_CHANNEL=$CHANNEL
    fi
```

Daniel-Fan (approved):
/lgtm

PR #1519, opened 2023-10-03
Zen extension update script

bluzarraga:
This script (mounted into the utility image via configmap just like backup zen5 and restore zen5 scripts) looks up the parameter configmap to determine customizations to make to existing zen5 instances. Right now, this is limited to zen extensions.

The script takes in a comma-delimited list of zen extensions from the configmap, searches the database for their ids via the zen api, then sequentially deletes each extension from the database using the zen api.

Could possibly build on this to include necessary logic for custom routes.


bluzarraga:
I have also added a step to the restore_zen5.sh script to interpret the new custom route host name from a specified configmap (the same one as the zen extension) and apply this change to the existing zenservice.

Daniel-Fan (approved):
/lgtm

Daniel-Fan (approved):
/lgtm

PR #1520, opened 2023-10-03
[WIP] update mongo velero br process to work on z arch

bluzarraga:
Mongo does not accept ssl commands on z and all of the backup and restore commands rely on ssl to connect to mongo to run these commands so as of this writing, z is not a supported architecture for backup and restore of mongo. This problem was initially addressed in multi instance and bedrock 4 migration scripts with these prs: https://github.com/IBM/ibm-common-service-operator/pull/1308 & https://github.com/IBM/ibm-common-service-operator/pull/1330. The logic used in this pr comes directly from those two previous prs as this is the same problem.

Testing
velero process
- prep a z environment for backup and restore via velero
- deploy the z specific backup resources for mongo
- proceed with velero backup restore as usual
- verify backup completes successfully (will need to check the logs of the backup to verify, it may say "completed" without backing up)
- on either the same cluster or a new cluster, restore the mongo-data according to velero steps
- verify this actually restored via logs

workaround process
- can use same cluster setup for this test
- run the backup with ./mongo-backup.sh <cs namespace> <stgclassname> true
- verify completion
- run the restore with ./restore-mongo.sh <cs namespace> true
- verify completion

These changes will also need to be applied to the scripts repo since these changes will likely need to be picked up from there for 4.0, 4.1, 4.2, and LTSR. I will open that pr after this one is completed.


bluzarraga:
replace this pr with https://github.com/IBM/ibm-common-service-operator/pull/1531 and #1530, closing

PR #1525, opened 2023-10-06
Fix for isolation LSR when it is corrupted

martabereta:


cp3pt0-deployment/migrate_tenant.sh
```diff
@@ -345,21 +345,24 @@ function isolate_license_service_reporter(){
...
+                fi
+
+                # label LSR PV as LSR PV for further LSR upgrade
+                ${OC} label pv $VOL license-service-reporter-pv=true --overwrite 
+                debug1 "License Service Reporter PV labeled with 'license-service-reporter-pv=true'"
+            
+                ${OC} patch pv $VOL -p '{"spec": { "persistentVolumeReclaimPolicy" : "Retain" }}'
+                debug1 "License Service Reporter PV reclaim policy set to 'Retain'"
+            else
+                info "No Lisense Service Reporter PVC found in $MASTER_NS or it is not in 'Bound' state, skipping isolation."
```
Daniel-Fan, line 364: Hi @martabereta 

Could we keep use `$OPERATOR_NS` here for `migrate_tenant.sh`

Daniel-Fan (changes requested):

cp3pt0-deployment/migrate_tenant.sh
martabereta: You are right. I've fixed it

Daniel-Fan (approved):
/lgtm

PR #1531, opened 2023-10-10
add necessary files for velero br on z

bluzarraga:
Mongodump and mongorestore commands cannot run on s390x without tweaks so this PR makes those necessary tweaks. We will not be able to run the velero br procedure completely automatically at the moment due to the need to scale down mongo operator and change `requiressl` connection setting to `preferssl` since the dump and restore commands do not support ssl on z.

Tested as part of https://github.com/IBM/ibm-common-service-operator/pull/1520. This pr also replaces #1520, more info can be found there.


Daniel-Fan (approved):
/lgtm

PR #1548, opened 2023-10-18
label cert-manager.io specific resources and custom route secrets

bluzarraga:
Add logic to specifically label secrets, certificates, and issuers using cert-manager.io version of the CRD to make sure no secrets fall through the cracks. Also included: labeling the custom secrets for the custom routes created for iam and zen.


bitscuit (approved):
/lgtm

PR #1549, opened 2023-10-18
label cert-manager.io specific resources and custom route secrets

bluzarraga:
same thing as #1548 but for the scripts branch


Daniel-Fan (approved):
/lgtm

PR #1550, opened 2023-10-18
add labels to zen5 permissions, add errexit to zen5 scripts

bluzarraga:
Add labels to the permissions files so they can be carried over to the new cluster, add errexit and errtrace to the scripts so if they fail for one reason or another, they fail both the script and the velero process. Otherwise the velero output may still incorrectly be "completed".


Daniel-Fan (approved):
/lgtm

PR #1551, opened 2023-10-18
add labels to zen5 permissions, add errexit to zen5 scripts

bluzarraga:
same as #1550 but for scripts branch


Daniel-Fan (approved):
/lgtm

PR #1553, opened 2023-10-20
wait for deployments to fully scale down

bluzarraga:
Possible that the subsequent command is run before the deployments are fully scaled down so we need to wait for them to be gone.


Daniel-Fan (approved):
/lgtm

PR #1559, opened 2023-10-25
update wait for deployments to scale down

bluzarraga:


bitscuit:
What is this for and why?

bluzarraga:
This is for restoring zen data. The process drops the entire database then replaces it but it can't be dropped if anything is currently using it. It's taken me a few tries to get consistent but if you don't wait for the pods from the deployments to stop before moving on, the drop will not happen which affects the restore of zen. The restore will be littered with duplicate errors and possibly not clean up data not from the target backup.

YCShen1010 (approved):
/lgtm

PR #1567, opened 2023-10-30
bump up zen restore timeout, add error handling for custom secrets

bluzarraga:
Zen restore can take a long time depending on a couple factors, better to give a timeout it won't hit prematurely. We need to label the secrets used for either the iam custom route or the zen custom route since they may or may not be labeled correctly. We also need to make sure the script can gracefully handle when these secrets/configmaps do not exist.


Daniel-Fan (approved):
/lgtm

PR #1576, opened 2023-11-01
Update OperandConfig with value reference

Daniel-Fan:
Please merge this PR ONLY AFTER ODLM's PR is merged: https://github.com/IBM/operand-deployment-lifecycle-manager/pull/990

### Update
1. KeyCloak Route is created by ODLM directly, and support value reference on following fields.
   - `spec.host`: Optional, user could specify custom hostname by updating `.data.keycloak_route_name` in `ibm-cpp-config` ConfigMap
   - `spec.tls.caCertificate`, `spec.tls.certificate`, `spec.tls.key`: Optional, user could specify custom certificate by creating secret `keycloak-custom-tls-secret`
  - `spec.tls.destinationCACertificate`: Required, it is referencing the CA cert from `cs-keycloak-tls-secret`
2. KeyCloak CR is created by ODLM, and support value reference on following fields.
  - `spec.hostname.hostname`: Required, it is reference the `.spec.host` from `keycloak` Route
  - `spec.ingress.enabled: false` to turn off KeyCloak operator creating Ingress Object, so it will use Route directly.
3. EDB Cluster CR is created by ODLM directly, and support value reference on following fields.
  - `spec.imageName`: Required.
     - The default value is `icr.io/cpopen/edb/postgresql:14.9@sha256:90136074adcbafb5033668b07fe1efea9addf0168fa83b0c8a6984536fc22264`.
     - And if the ConfigMap `cloud-native-postgresql-image-list` exists in operator namespace, it will reference `.data.ibm-postgresql-14-operand-image`
     - And user could specify `.data.edb-keycloak-operand-image` in `ibm-cpp-config` ConfigMap to override above default EDB image setting.

### DIFF
--- a/controllers/constant/odlm.go
+++ b/controllers/constant/odlm.go
@@ -800,7 +800,15 @@ spec:
                     data:
                       EDB_LICENSE_KEY: $(base64 /license_keys/edb/EDB_LICENSE_KEY | tr -d '\n')
                     EOF
-                  image: cp.icr.io/cp/cpd/edb-postgres-license-provider@sha256:e683c4bfceb5a99f7971409d4028cf326cdedb007f9cf3daf28b8141835535f1
+                  image:
+                    templatingValueFrom:
+                      default:
+                        required: true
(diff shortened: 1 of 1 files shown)


YCShen1010 (approved):
/lgtm

PR #1578, opened 2023-11-02
Update image used for mongo BR containers for SSL support on Power and Z

bluzarraga:
Power and Z mongotools did not support SSL connections so the Power and Z teams rebuilt the tools with SSL support and have included them in the mongo image used in this PR.


Daniel-Fan (approved):
/lgtm

PR #1579, opened 2023-11-02
new volumesnapshots resources permission from cloud-native-postgresql

YCShen1010:
New resources `volumesnapshots` permission was added in `cloud-native-postgresql.v1.18.7`,  need to grant this permission to NSS in order to create EDB role in service namespace


Daniel-Fan (approved):
/lgtm

PR #1587, opened 2023-11-06
add default admin creds to backup

bluzarraga:
should be the exact same as https://github.com/IBM/ibm-common-service-operator/pull/1586 just need to make it available in the scripts repo


Daniel-Fan (approved):
/lgtm

PR #1589, opened 2023-11-06
fix target ns check for all namespace mode

YCShen1010:
fix for the issue: https://github.ibm.com/IBMPrivateCloud/roadmap/issues/61091


Daniel-Fan (approved):
/lgtm

PR #1591, opened 2023-11-07
delete webhook configuration when cs pod down

YCShen1010:
When cs-operator is down, the webhook configuration block the update of CS CR, need to be deleted before updaing
issue: https://github.ibm.com/IBMPrivateCloud/roadmap/issues/61091#issuecomment-66388356


Daniel-Fan (approved):
/lgtm

PR #1595, opened 2023-11-07
add two more secrets to backup

bluzarraga:
exact same as #1594 but for the scripts branch to make it public


Daniel-Fan (approved):
/lgtm

PR #1597, opened 2023-11-08
Missing . in objectRef path

Jamstah:
Missing . means this comes out as a string instead of reading from the route as expected.


Daniel-Fan (approved):
/lgtm

PR #1598, opened 2023-11-08
update zen extension logic based on feedback

bluzarraga:
Update the logic for grabbing extension names to be more consistent based on feedback from CP4BA


Daniel-Fan (approved):
/lgtm

PR #1599, opened 2023-11-08
update zen extension logic based on feedback

bluzarraga:
Update the logic for grabbing extension names to be more consistent based on feedback from CP4BA, exact same as https://github.com/IBM/ibm-common-service-operator/pull/1598


Daniel-Fan (approved):
/lgtm

PR #1608, opened 2023-11-09
create nss configmap in cs namespace for all ns upgrade

YCShen1010:
In all namespace migration, as all nss CR will be cleared after running the migrate_tenant script, which causes fail to delete ingress operator and operand due to no `namespace-scope` ConfigMap found. We will create a new nss configmap back in `ibm-common-services` namespace only for `ingress` operator deletion.


Daniel-Fan (approved):
/lgtm

PR #1610, opened 2023-11-09
Update Zen extension logic to capture extensions before backup, standardize some values across files

bluzarraga:
See issues 60195, 61131, and 61134

### DIFF
--- a/velero/schedule/zen5-br-scripts-cm.yaml
+++ b/velero/schedule/zen5-br-scripts-cm.yaml
@@ -2,6 +2,7 @@ apiVersion: v1
 kind: ConfigMap
 metadata:
   name: zen5-br-configmap
+  namespace: <zenservice namespace>
   labels:
     foundationservices.cloudpak.ibm.com: zen5-data
 data:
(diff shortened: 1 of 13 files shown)


velero/schedule/zen5-br-scripts-cm.yaml
```diff
@@ -355,12 +353,35 @@ data:
...
+                        extensions="$ext_name,"
+                    else
+                        extensions="$extensions,$ext_name"
+                    fi
+                done
+                
+                extensions=$(echo $extensions | tr "," " ")
+                echo $extensions
+                
+                cat <<< $extensions > /tmp/extensions.txt
```
joshua-bj, line 373: @bluzarraga , Hi Ben, the `/tmp/` folder is not in the backup scope, I guess you need persist this file at `/zen5/zen-backup` or its sub folder

joshua-bj: Oh, I see there is a new "emptyDir" mounted at the backup pod, but I'm wondering why you do not use the `/zen5/zen-backup`. The emptyDir is usually for temporary persistency request like cache. Seems like it's any pattern to be considered at the backup. Just my gut feeling.

bluzarraga: changed the directory to /zen5/extensions so it is in the same parent directory as the rest of the backup

Daniel-Fan (approved):
/lgtm

PR #1612, opened 2023-11-10
fixed licensing upgrade regression caused by opreq

bitscuit:
- the graceful upgrade method by saving subscription yaml first caused a regression because the opreq-control label still existed, which causes ODLM v3 to still manage it even after v4 migration


Daniel-Fan (approved):
/lgtm

PR #1615, opened 2023-11-10
fixed upgrade_operator not deleting opreq label

bitscuit:
conditional was checking for empty instead of non-empty value


Daniel-Fan (approved):
/lgtm

PR #1617, opened 2023-11-10
match scripts dev commits

bluzarraga:
Exact same changes as here https://github.com/IBM/ibm-common-service-operator/pull/1610 to be made available publicly


YCShen1010 (approved):
/lgtm

PR #1627, opened 2023-11-14
changed opreq-control label removal implementation

bitscuit:
to use oc patch instead of oc apply because of unknown issue in oc apply causing labels to return after being removed

### DIFF
--- a/cp3pt0-deployment/common/utils.sh
+++ b/cp3pt0-deployment/common/utils.sh
@@ -1189,7 +1189,7 @@ function update_operator() {
     # something (either Open Shift or OLM), so need to use patch
     # "~1" is the escape sequence for "/" character
     if [ ! -z "$remove_opreq_label" ]; then
-        ${OC} patch sub ${sub_name} --type='json' -p='[{"op": "remove", "path": "/metadata/labels/operator.ibm.com~1opreq-control"}]' || info "Could not patch Subscription to remove label"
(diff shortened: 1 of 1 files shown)


cp3pt0-deployment/common/utils.sh
```diff
@@ -1184,14 +1184,18 @@ function update_operator() {
     fi
 
     title "Updating ${sub_name} in namesapce ${ns}..."
+
+    # there is currently a bug with oc apply where the removed labels are added back by
+    # something (either Open Shift or OLM), so need to use patch
+    # "~1" is the escape sequence for "/" character
+    if [ ! -z "$remove_opreq_label" ]; then
+        ${OC} patch sub ${sub_name} --type='json' -p='[{"op": "remove", "path": "/metadata/labels/operator.ibm.com~1opreq-control"}]' || info "Could not patch Subscription to remove label"
```
Daniel-Fan, line 1192: @bitscuit 
Would we add a namespace and use full resource name for the command
```suggestion
        ${OC} patch subscription.operators.coreos.com ${sub_name} -n ${ns} --type='json' -p='[{"op": "remove", "path": "/metadata/labels/operator.ibm.com~1opreq-control"}]' || warning "Could not patch Subscription ${sub_name} in ${ns} to remove label"
```

Daniel-Fan:
Test on patch command looks good to me
```
➜  ~ oc patch subscription.operators.coreos.com keycloak-operator -n keycloak --type='json' -p='[{"op": "remove", "path": "/metadata/labels/operator.ibm.com~1opreq-control"}]'
subscription.operators.coreos.com/keycloak-operator patched
```

Daniel-Fan (changes requested):

cp3pt0-deployment/common/utils.sh
bitscuit: Yes, updated

Daniel-Fan (approved):
/lgtm

PR #1636, opened 2023-11-14
change when housekeeping command is run

bluzarraga:
move the clean up of backups to the beginning of the next backup instead of immediately after. This way, the backup is accessible between runs and out of the way when attempting another backup.


Daniel-Fan (approved):
/lgtm

PR #1642, opened 2023-11-17
remove invalid option in script

bluzarraga:
--ignore-not-found cannot be used on the scale command as it will cause an error


Daniel-Fan (approved):
/lgtm

PR #1647, opened 2023-11-20
resolve syntax error, change check command

bluzarraga:
Same as #1646 

Needed to add a space before `--ignore-not-found` as the string was being included in the replica value for zen core api

Changed the check from directly grabbing zen-watchdog to grepping for it to prevent an error from being output.


Daniel-Fan (approved):
/lgtm

PR #1649, opened 2023-11-20
change message from error to warning for more consistent results

bluzarraga:
This message being an error will prevent the script from cleaning up the cluster after running. A warning is better suited for this situation as it denotes that the zen extensions were still not found (which may or may not be an issue) while still allowing the script to clean up after itself.


bluzarraga:
Large change added in here but it is very related to the original change in this pr. Due to the zen core api and usermgmt deployments being scaled down, the curl command to retrieve the list of IDs for the zenextensions would not return anything (since there was nothing listening at the endpoint we were trying to hit). Scaling usermgmt and zen core api deployments before running the zenextension restore steps resolved this issue. I still think the change to a warning from an error is necessary though as certain extensions do not have database IDs (like the common ui one) and shouldn't return an error.

Daniel-Fan (approved):
/lgtm

PR #1652, opened 2023-11-23
Remove multiInstance template

Daniel-Fan:
Issue: https://github.ibm.com/IBMPrivateCloud/roadmap/issues/61292

CommonService operator v4 should not change anything about any v3 components configuration.
It should no longer render multipleInstancesEnabled parameter in OperandConfig for both mgmt-ingress and commonWebUI
- mgmt-ingress CR will keep what it has before the upgrade, either `multipleInstancesEnabled: false` or `multipleInstancesEnabled: true` depending on how CS v3 instance looks like before upgrade, but it will not change during the upgrade.
- commonWebUI does not use this parameter at all, [confirming with Steve before](https://github.ibm.com/IBMPrivateCloud/roadmap/issues/60890#issuecomment-66381515)


YCShen1010 (approved):
/lgtm

PR #1655, opened 2023-11-28
Add default realm security settings

Jamstah:
Set sensible defaults for the realm security.
Addresses https://github.ibm.com/IBMPrivateCloud/roadmap/issues/61319

These can be overridden by users because the KeycloakRealmImport CR only adds the realm once and then doesn't reconcile except on delete.


pgodowski (approved):

Jamstah:
This one is ready to merge.

YCShen1010 (approved):
/lgtm

PR #1660, opened 2023-11-29
use new mongo image with new mongorestore and mongodump tools that utilize ssl on power/z clusters

bluzarraga:
With help from the Power and Z teams last month, we have a new mongo image version that includes mongorestore and mongodump binaries for Power and Z that can use the `--ssl` options. This change allows us to remove the platform specific steps used to backup and restore the mongo database and that is what this pr does.

Test:
- setup cluster with an instance of mongo (preferably on a power or z cluster)
- run preload_data.sh to a test namespace
- verify script completes
- check outputted logs to verify backup and restore performed successfully

### DIFF
--- a/preload_data.sh
+++ b/preload_data.sh
@@ -29,7 +29,7 @@ TO_NAMESPACE=""
 NUM=$#
 TEMPFILE="_TMP.yaml"
 DEBUG=0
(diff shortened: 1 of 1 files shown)


Daniel-Fan (approved):
/lgtm

PR #1663, opened 2023-11-30
Create a dedicate controller for self-isolation

Daniel-Fan:
Create a new dedicate controller to perform the self-isolation process, the reconciliation will only be triggered by `common-service-maps` ConfigMap in `kube-public` namespace.


YCShen1010 (approved):
/lgtm

PR #1666, opened 2023-11-30
Enable and Disable CS CR reconciliation in self-isolation

YCShen1010:
Issue: https://github.ibm.com/ibmprivatecloud/roadmap/issues/61346

#### Context

- Enable maintenance mode for CS CR at the beginning of self-isolation process
- Disable maintenance mode at the end of self-isolation

### DIFF
--- a/controllers/common/util.go
+++ b/controllers/common/util.go
@@ -44,6 +44,7 @@ import (
 	"k8s.io/klog"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 
+	apiv3 "github.com/IBM/ibm-common-service-operator/api/v3"
 	"github.com/IBM/ibm-common-service-operator/controllers/constant"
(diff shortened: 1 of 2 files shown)


controllers/common/util.go
```diff
@@ -771,3 +772,60 @@ func FindDifference(superset, subset []string) []string {
...
+	}
+	return nil
+}
+
+// DisableMaintenanceMode disables maintenance mode for CommonService CR
+func DisableMaintenanceMode(c client.Client, masterNs string) error {
+	// Fetch CommonService CR
+	instance := &apiv3.CommonService{}
+	if err := c.Get(context.TODO(), types.NamespacedName{
+		Name:      "common-service",
```
Daniel-Fan, line 810: Could we make CS CR name as a variable? So it would be passed into the function.

Daniel-Fan (approved):
/lgtm

PR #1667, opened 2023-12-01
Delete Crossplane, webhook, and secretshare deployment

qpdpQ:
for ticket: https://github.ibm.com/ibmprivatecloud/roadmap/issues/61347

Delete Crossplane, webhook, and secretshare deployment and resources

### DIFF
--- a/controllers/scopewatcher_controller.go
+++ b/controllers/scopewatcher_controller.go
@@ -24,6 +24,7 @@ import (
 	ctrl "sigs.k8s.io/controller-runtime"
 	"sigs.k8s.io/controller-runtime/pkg/reconcile"
 
+	"github.com/IBM/ibm-common-service-operator/controllers/bootstrap"
 	util "github.com/IBM/ibm-common-service-operator/controllers/common"
 	"github.com/IBM/ibm-common-service-operator/controllers/constant"
 )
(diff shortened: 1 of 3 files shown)


controllers/scopewatcher_controller.go
```diff
@@ -98,6 +99,93 @@ func (r *CommonServiceReconciler) ScopeReconcile(ctx context.Context, req ctrl.R
...
+		if err := r.Bootstrap.Cleanup(r.Bootstrap.CSData.ControlNs, deployment); err != nil {
+			return ctrl.Result{}, err
+		}
+	}
+
+	for _, resource := range CP2Resources {
+		if err := r.Bootstrap.Cleanup(r.Bootstrap.CSData.ControlNs, resource); err != nil {
+			return ctrl.Result{}, err
+		}
+	}
```
Daniel-Fan, line 188: Hi @qpdpQ 

Could we update the namespace for cleaning cluster singleton service to `MasterNs` which is typically `ibm-common-services`?
The main purpose of this cleaning is to migrate cluster singleton service from `MasterNs` to `ControlNs`. In another word, if it is in `ControlNs` already, we do not have to delete them.

Daniel-Fan: Also I am wondering that does the sequence of removing operator and operand matter?

Would we consider removing operand first, so operator will have chance to reconcile the resources, and remove the finalizers in the CRs if it is necessary.

Daniel-Fan (approved):
/lgtm

PR #1670, opened 2023-12-01
Skip reconciliation if CR is not in tenant scope

Daniel-Fan:
issue: https://github.ibm.com/ibmprivatecloud/roadmap/issues/61446

CS operator v3 will [skip the reconciliation for the CS CR which is not in the current scope.](https://github.com/IBM/ibm-common-service-operator/blob/91e0f67dfccbed9f64438d9c48d260e8962e4dd1/controllers/commonservice_controller.go#L108), avoid any changes or update to that CR.


YCShen1010 (approved):
/lgtm

PR #1683, opened 2023-12-04
Fix swapped EDB image override configmap reference key/name

pgodowski:
The key/name values for EDB image override configmap entry in ibm-cpp-config are swapped. Fixed this.


Daniel-Fan (approved):
/lgtm

PR #1684, opened 2023-12-04
Isolate License Service Reporter during self-isolation

YCShen1010:
Issue: https://github.ibm.com/ibmprivatecloud/roadmap/issues/61347

Isolate License Service Reporter

- change persistentVolumeReclaimPolicy to Retain
- label LSR PV as LSR PV for further LSR upgrade

PR #1687, opened 2023-12-05
[323] Isolate License Service Reporter during self-isolation (#1684)

YCShen1010:
* isolate LSR

* add permission for LSR, PVC and PV

* add get pv function


Daniel-Fan (approved):
/lgtm

PR #1688, opened 2023-12-05
Trigger CR reconciliation with mapping ConfigMap

Daniel-Fan:
Address issue https://github.ibm.com/IBMPrivateCloud/roadmap/issues/61530

### issue
After modifying `common-service-maps` ConfigMap, to add test1 namespace into the scope of CS 3.x (along with ibm-common-services), nothing happened
- expected that test1 namespace would be added into CS 3.x scope automatically, since `common-service-maps` was edited

### To-do
The update on `common-service-maps` ConfigMap should trigger the reconciliation of CS CR.


YCShen1010 (approved):
/lgtm

PR #1689, opened 2023-12-05
Update excluded namespace list before removing nss CRs managed by ODLM

YCShen1010:
Updates on self-isolation process:

1. Updating NSS CRs `common-service` and `nss-odlm-scope` by adding updatedNsList and removing excludedNsList
2. Remove NSS CRs `nss-managedby-odlm` and `odlm-scope-managedby-odlm`


Daniel-Fan (approved):
/lgtm

PR #1694, opened 2023-12-07
Avoid overwrite fields in merged CR

Daniel-Fan:
Fix issue https://github.ibm.com/IBMPrivateCloud/roadmap/issues/61393

When we are merging all CS CRs in the tenant scope, we extract the profile from each individual CS CR into a list.
For example
- `common-service` CR with `starter` size
- `cp4i-cs` CR with `large` size
- `cp4ba-cs` CR with `medium` size

Then we merge different profiles together to get the LARGEST value(it respect to `large` size), and save it into a summarized profile `summaryCR`.

But I always initialized this summarized profile variable `summaryCR` **as long as either one of `resources` OR `spec` field is empty**.
```
if summaryCR == nil || summaryCR.(map[string]interface{})["spec"] == nil || summaryCR.(map[string]interface{} ["resources"] == nil {
	summaryCR = map[string]interface{}{
		"name":      operator.(map[string]interface{})["name"].(string),
		"spec":      map[string]interface{}{},
		"resources": []interface{}{},
	}
}
```

Many of our profiles only contain one field. So `summaryCR` for each operator is always be initialized
```
### Only contain spec
- name: ibm-im-operator
    spec:
      authentication:
        authService:
          resources:
            limits:
              cpu: 2000m

### Only contain resources
- name: keycloak-operator
  resources:
    - apiVersion: k8s.keycloak.org/v2alpha1
      data:
        spec:
          db:
            host: keycloak-edb-cluster-rw
      force: true
      kind: Keycloak
      name: cs-keycloak
```

To fix this, I change it to initialize the field individually, to avoid overwrite.

### DIFF
--- a/controllers/operandconfig.go
+++ b/controllers/operandconfig.go
@@ -101,12 +101,16 @@ func mergeCSCRs(csSummary, csCR, ruleSlice []interface{}, serviceControllerMappi
 	for _, operator := range csCR {
 		summaryCR := getItemByName(csSummary, operator.(map[string]interface{})["name"].(string))
 		rules := getItemByName(ruleSlice, operator.(map[string]interface{})["name"].(string))
-		if summaryCR == nil || summaryCR.(map[string]interface{})["spec"] == nil || summaryCR.(map[string]interface{})["resources"] == nil {
+		if summaryCR == nil {
 			summaryCR = map[string]interface{}{
 				"name":      operator.(map[string]interface{})["name"].(string),
 				"spec":      map[string]interface{}{},
 				"resources": []interface{}{},
 			}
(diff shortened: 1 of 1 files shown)


bluzarraga:
Using this image, I ran the scenarios @Daniel-Fan described successfully. I also tried a couple scenarios that I had run into while investigating root cause successfully so I think this pr is good to merge. The scenario that mimics what aiops originally saw was where the CS CR was present before the cs operator which I mimicked in the following way:
- scaled down cs operator
- deleted default CS CR
- deleted operandconfig 
- created dummy aiops CS CR using large size
- replaced cs operator image in csv
- scaled cs operator back up
- watched for reconcile to complete
- verified operandconfig was using large size after cs operator intialization

Great work Daniel!

PR #1695, opened 2023-12-07
fix: get extreme value on resources field when spec field does not exist

Daniel-Fan:
To fix issue https://github.ibm.com/IBMPrivateCloud/roadmap/issues/61204

Update the code to not skip loop when ONLY `spec` field is `nil`.
Otherwise, it will not able to reconcile `resources` field when `spec` field is `nil`

### Previously
```
for _, opService := range opconServices {
                // Skip loop if spec is nil
		if opService.(map[string]interface{})["spec"] == nil {
			continue
		}
                update_spec_field
                
                // update resources when resources is NOT nil
                if opService.(map[string]interface{})["resources"] != nil {
			update_resources_field
		}
}
```

### Now
```
for _, opService := range opconServices {
                // update spec when spec is NOT nil
		if opService.(map[string]interface{})["spec"] != nil {
			update_spec_field
		}
               
                // update resources when resources is NOT nil
                if opService.(map[string]interface{})["resources"] != nil {
			update_resources_field
		}
}
```

### DIFF
--- a/controllers/operandconfig.go
+++ b/controllers/operandconfig.go
@@ -530,24 +530,23 @@ func (r *CommonServiceReconciler) getExtremeizes(ctx context.Context, opconServi
 
 	for _, opService := range opconServices {
 		crSummary := getItemByName(configSummary, opService.(map[string]interface{})["name"].(string))
-		if opService.(map[string]interface{})["spec"] == nil {
-			continue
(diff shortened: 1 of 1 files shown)


YCShen1010 (approved):
/lgtm

PR #1696, opened 2023-12-07
update UID and resourceVersion when restoring LS CRs

YCShen1010:
Issue: https://github.ibm.com/IBMPrivateCloud/roadmap/issues/61543
- Add delete verb for `ConfigMaps` resources
- set/overwrite UID and resourceVersion when restoring LS CRs


Daniel-Fan (approved):
/lgtm

PR #1698, opened 2023-12-07
avoid overwriting CS CR

bluzarraga:
Same fix as https://github.com/IBM/ibm-common-service-operator/pull/1694. More details can be found there


Daniel-Fan (approved):
/lgtm

PR #1699, opened 2023-12-07
[efix] fix: get extreme value on resources field when spec field does not exist

Daniel-Fan:
Cherry-pick of #1695


YCShen1010 (approved):
/lgtm

PR #1702, opened 2023-12-08
Add managedByCsOperator label for ConfigMap

Daniel-Fan:
issue: https://github.ibm.com/IBMPrivateCloud/roadmap/issues/61558

When `common-service-maps` ConfigMap does not contain label `operator.ibm.com/managedByCsOperator: true`, any creation, deletion, update and other generic event on the ConfigMap will not trigger the reconciliation of v3 LTSR self-isolation.

We need to make sure v4 operator add this label into the ConfigMap when v4 operator updates or creates `common-service-maps` ConfigMap.

### DIFF
--- a/controllers/bootstrap/init.go
+++ b/controllers/bootstrap/init.go
@@ -703,6 +703,12 @@ func (b *Bootstrap) CreateCsMaps() error {
 		Data: data,
 	}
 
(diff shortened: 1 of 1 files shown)


YCShen1010 (approved):
/lgtm

PR #1710, opened 2023-12-13
added permissions explicitly for cloud-native-postgresql-image-list c…

bitscuit:
…onfigmap

### DIFF
--- a/bundle/manifests/ibm-common-service-operator.clusterserviceversion.yaml
+++ b/bundle/manifests/ibm-common-service-operator.clusterserviceversion.yaml
@@ -189,11 +189,6 @@ spec:
               resources:
                 - configmaps
               verbs:
-                - create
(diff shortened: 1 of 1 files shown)


bundle/manifests/ibm-common-service-operator.clusterserviceversion.yaml
```diff
@@ -182,6 +182,19 @@ spec:
               verbs:
                 - delete
                 - update
+            - apiGroups:
+                - ""
+              resourceNames:
+                - cloud-native-postgresql-image-list
```
pgodowski, line 188: is this the ONLY configmap which we need to update?

pgodowski (changes requested):

Daniel-Fan (approved):
/lgtm

PR #1727, opened 2024-01-04
Theme updates

Jamstah:


sgrube (commented):
@Jamstah The installer will be rebuilding and inserting the theme jar into the installer image at build time so the theme.jars are going to be deleted soon.
Can you remove cloud-pak-theme.jar from this PR and update your Makefile target to produce `cloudpak-theme.jar` to match the install (assuming this target is for development).

Jamstah:
You mean like this?

I don't think the makefile is just for development. I would suggest that the CICD pipeline for the installer image runs the makefile to generate the jar as well.

sgrube:
We didn't want to add the JVM to our current prow image just for the jar command, so the archive is zipped and added to the installer image via the installer CICD process.  We will revisit a specific build for the theme if it grows beyond simply creating a zip from this content.

Daniel-Fan (approved):

sgrube:
@Jamstah yup, jar is just an archive, but I did copy the manifest directory from an original jar to be safe (it was largely empty though).
I agree that having the build instructions there is good and was happy to leave the target in there.  I left it to the install squad to zip in any manner they see fit since they are delivering the archive to the cluster.  The makefile came after their support was added.

PR #1728, opened 2024-01-04
Create EDB Cluster CR after License Secret is injected for operator

Daniel-Fan:
Issue: https://github.ibm.com/IBMPrivateCloud/roadmap/issues/60937#

### Existing issue:
1. ODLM install EDB operator
2. ODLM create a Job for injecting EDB License Information into EDB operator
   1. Job create a Secret containing EDB License information
   2. :heavy_exclamation_mark: Job restart EDB operator to load secret value
3. ODLM create a EDB Cluster CR for Keycloak
4. EDB operator reconciles EDB Cluster CR.
   - :x: Sometimes when EDB operator is initializing the database/PVC for EDB Cluster CR, EDB operator is restarted by Job in Step 2.ii.
   - This restart will cause EDB Cluster CR reconciliation stuck at status `Setting up primary`.

### Solution:
1. ODLM install EDB operator
2. ODLM create a Job for injecting EDB License Information into EDB operator
   1. Job create a Secret containing EDB License information
   2. :heavy_exclamation_mark: Job restart EDB operator to load secret value
   3. :white_check_mark: Job add an annotation into the secret, indicating the restart of EDB operator is completed.
      - `license-config-secret-applied: true`
5. ODLM hold and do NOT create a EDB Cluster CR for Keycloak until above secret containing the annotation.
   - :white_check_mark:  When the EDB operator restarts is completed, and secret is annotated, ODLM create EDB Cluster CR.
6. EDB operator reconciles EDB Cluster CR.
   - :white_check_mark: EDB operator will NOT be restarted during the reconciliation of EDB Cluster CR.

### Test
- After the Job is completed, it's pod logs indicate that operator pod is deleted, and secret is annotated.
   ```bash
    > oc logs create-postgres-license-config-l2w5f -n test-edb

    Defaulted container "restart-edb-pod" out of: restart-edb-pod, edb-license (init)

    pod "postgresql-operator-controller-manager-1-18-7-6b66dcccf-dmhf6" deleted
    secret/postgresql-operator-controller-manager-config annotated
   ```
- EDB operator pod was re-created at time `2024-01-04T21:11:03Z`
   ```yaml
    kind: Pod
    apiVersion: v1
    metadata:
      name: postgresql-operator-controller-manager-1-18-7-6b66dcccf-zfzsc
      creationTimestamp: '2024-01-04T21:11:03Z'
   ```
- EDB Cluster CR is created at time `2024-01-04T21:11:19Z` which is later then EDB operator restarted time.
  ```yaml
  apiVersion: postgresql.k8s.enterprisedb.io/v1
  kind: Cluster
  metadata:
    creationTimestamp: '2024-01-04T21:11:19Z'
    name: keycloak-edb-cluster
  ```


controllers/constant/odlm.go
```diff
@@ -777,8 +777,10 @@ spec:
                 - command:
                   - bash
                   - '-c'
-                  - >-
+                  args:
+                  - |
                     kubectl delete pods -l app.kubernetes.io/name=cloud-native-postgresql
```
pgodowski, line 782: could this error and exit in case there is not yet running edb operator Pod?

controllers/constant/odlm.go
```diff
@@ -847,6 +849,15 @@ spec:
             apiGroup: rbac.authorization.k8s.io
       - apiVersion: postgresql.k8s.enterprisedb.io/v1
         data:
+          dummy-field:
```
pgodowski, line 852:
```suggestion
          license-key-sentinel:
```

pgodowski (changes requested):

pgodowski:
> Why we need to use a dummy-field in EDB Cluster CR?

Besides my suggestion to change the name of this field, I am a bit concern about this approach in general, i.e. to assume / delegate to k8s the field pruning. This might introduce some k8s api server auditing events, without any good reason, introducing the potential confusion.

I was reading the EDB operator [docs](https://www.enterprisedb.com/docs/postgres_for_kubernetes/latest/pg4k.v1/#postgresql-k8s-enterprisedb-io-v1-ClusterSpec) and found this sample too: https://www.enterprisedb.com/docs/postgres_for_kubernetes/latest/samples/cluster-example-full.yaml where there is `.spec.description` field which seems can have some free form text.

So, perhaps what we can do is to add `.spec.description` with some value coming up from the `postgresql-operator-controller-manager-config` `Secret`? Of course not just `true`, but something like "Database with applied license key" or something along those lines.

controllers/constant/odlm.go
Daniel-Fan: I think `deleting pod by label` will not return error in case edb operator pod is not found.
```
➜  ~ oc delete pod -l app.kubernetes.io/name=cloud-native-postgresql -n test-edb
pod "postgresql-operator-controller-manager-1-18-7-6b66dcccf-xrjwt" deleted
➜  ~ echo $?
0

➜  ~ oc delete pod -l app.kubernetes.io/name=xxxxxxxxxx -n test-edb
No resources found
➜  ~ echo $?
0
```

But indeed if we try to delete a specific pod, it returns error in case that pod does not exist.
```
➜  ~ oc delete pod xxxxxx
Error from server (NotFound): pods "xxxxxx" not found
➜  ~ echo $?
1
```

Daniel-Fan:
Thank you @pgodowski for the review!

It is really a good idea to add `.spec.description` with some value coming up from the `postgresql-operator-controller-manager-config` Secret.

Updating the Cluster CR `.spec.description` with `EDB Database with IBM License Key`, could you please help to review again?

pgodowski:
it looks good to me know - thanks a lot @Daniel-Fan !

YCShen1010 (approved):
/lgtm

PR #1730, opened 2024-01-05
revert changes to account for z and power binaries

bluzarraga:
In late October, working alongside the mongo, Power, and Z teams, the mongo image was updated to include mongodump and mongorestore binaries capable of connecting to mongo via ssl connection on Power and Z clusters (which was previously not supported). This represented a significant security improvement. However, this requires specific versions of the mongo image that may not be guaranteed to exist on a cluster. There are further complications when dealing with airgap clusters (SERT raised a SEV1 in December because the script stopped functioning when it could no longer pull the mongo image on an airgap cluster). We did not see this problem previously because we did not specify the mongo image used; we just used the existing image on the cluster.

Leaving the script as is (ie not merging in this PR) will prevent SERT tests from completing in airgap clusters. We either make this change or a much larger one to resolve the problem altogether. More explanation on the problem and possible solution found in https://github.ibm.com/IBMPrivateCloud/roadmap/issues/61674

### DIFF
--- a/preload_data.sh
+++ b/preload_data.sh
@@ -29,7 +29,7 @@ TO_NAMESPACE=""
 NUM=$#
 TEMPFILE="_TMP.yaml"
 DEBUG=0
-s390x_ENV="false"
+z_or_power_ENV="false"
 
 # ---------- Command variables ----------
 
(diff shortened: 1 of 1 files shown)


preload_data.sh
```diff
@@ -138,6 +138,20 @@ function prereq() {
         error "Namespace $TO_NAMESPACE does not exist (or oc command line is not logged in)"
         exit 1
     fi
+    mongo_node=$(${OC} get pods -n $FROM_NAMESPACE -o wide | grep icp-mongodb-0 | awk '{print $7}')
+    architecture=$(${OC} describe node $mongo_node | grep "Architecture:" | awk '{print $2}')
+    if [[ $architecture == "s390x" ]]; then
+      s390x_ENV="true"
+      info "Z or Power cluster detected, be prepared for multiple restarts of mongo pods. This is expected behavior."
```
Daniel-Fan, line 145: @bluzarraga 

May I ask that if Z and Power Cluster both require non-ssl connection, should we update the architecture variable to `s390x` or `ppc64le`?

bluzarraga: Yes, good catch

Daniel-Fan (approved):
/lgtm

PR #1734, opened 2024-01-08
Resolve issue with WaitForFirstConsumer volumeBindingMode

bluzarraga:
A couple changes needed to be made to resolve the problem found in SERT issue 60021

- The previous variable used in the check was not set prior to being used, changed to use the existing `stgclass` variable instead
- There was no check on the restore side, only the backup. Copy and pasted the check on the backup side to also be on the restore side
- noticed an unrelated problem with claim ref repopulating before deleting the old pvc. Changed the claimref clearance to after the old pvc was deleted

Tested the script beginning to end by creating a new storageclass `rook-cephfs-test` which was a copy pasted `rook-cephfs` but changed the `volumeBindingMode` to `WaitForFirstConsumer` which should be representative of the problem experienced in the issue.

Waiting to see if SERT can produce an AWS cluster to have a more direct test.

### DIFF
--- a/preload_data.sh
+++ b/preload_data.sh
@@ -517,14 +517,15 @@ function swapmongopvc() {
   fi
 
   ${OC} patch pv $VOL -p '{"spec": { "persistentVolumeReclaimPolicy" : "Retain" }}'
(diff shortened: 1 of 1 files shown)


Daniel-Fan (approved):
/lgtm

PR #1739, opened 2024-01-10
Fix bug in theme - switch from cloud-pak to cloudpak name

Jamstah:
Bugs from importing from my repo - I was using cloud-pak as the name and the right name is cloudpak.

This enables the theme to show up for admin/account consoles.


Jamstah:
FYI @sgrube

Daniel-Fan (approved):
/lgtm

PR #1740, opened 2024-01-10
Update preview dir path to support multi non-root user

Daniel-Fan:
Issue: https://github.ibm.com/IBMPrivateCloud/roadmap/issues/61784

Each execution of script will re-direct the preview file into different directory. So it will support different non-root users to run the same script under on same Linux based client.

After executing `setup_singleton.sh` multiple times, this is the files and directories under `/tmp` folder
```
> ls -r /tmp/*-preview

/tmp/setup-singleton-20240110105758-preview:
operatorgroup.yaml

/tmp/setup-singleton-20240110105549-preview:
operatorgroup.yaml issuer.yaml        certificate.yaml

/tmp/isolate-20240110100317-preview:
```


YCShen1010 (approved):
/lgtm

PR #1743, opened 2024-01-10
Reference EDB License Secret from OperatorNamespace

Daniel-Fan:
Issue: https://github.ibm.com/IBMPrivateCloud/roadmap/issues/61807

When CommonService is deployed in All Namespace mode or separation of control and data, `postgresql-operator-controller-manager-config` secret will be in operator namespace. And EDB CR is in a different namespace which is services namespace.


YCShen1010 (approved):
/lgtm

PR #1747, opened 2024-01-10
updated opcon so Keycloak uses OCP certs

bitscuit:
Updated default OperandConfig so that Keycloak uses OCP service serving cert instead of generating Certificate from cert-manager
- prefixed intermediary resources like the Service and service cert secret name with `cpfs-opcon-`, open for feedback to change name
- Route and Secret containing certs both still have the same name as what was GA-ed, so backwards compatibility maintained

Also updated role.yaml to properly generate CSV

How to test:

Pre-reqs:

1. pull secret for EDB and EDB catalogsource

Steps:

1. Install cs-operator from GA
2. Edit the image in CSV to `quay.io/henry_h_li/common-service-operator-amd64:dev` or build your own image from this PR
3. Delete the default OperandConfig and wait for it to be re-created
4. Create OperandRequest for keycloak

```
apiVersion: operator.ibm.com/v1alpha1
kind: OperandRequest
metadata:
  name: keycloak-test
  namespace: cp1
spec:
  requests:
    - operands:
        - name: keycloak-operator
      registry: common-service
      registryNamespace: ibm-common-services
```

### DIFF
--- a/bundle/manifests/ibm-common-service-operator.clusterserviceversion.yaml
+++ b/bundle/manifests/ibm-common-service-operator.clusterserviceversion.yaml
@@ -8,6 +8,11 @@ metadata:
           "apiVersion": "operator.ibm.com/v3",
           "kind": "CommonService",
           "metadata": {
+            "labels": {
+              "app.kubernetes.io/instance": "ibm-common-service-operator",
+              "app.kubernetes.io/managed-by": "ibm-common-service-operator",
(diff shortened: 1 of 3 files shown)


bitscuit:
Note that the changes in the CSV are auto-generated, and mostly have to do with spacing

Daniel-Fan (approved):
/lgtm

PR #1752, opened 2024-01-11
Overwrite the resource field when default rule is not specified

Daniel-Fan:
issue: https://github.ibm.com/IBMPrivateCloud/roadmap/issues/61204

There are two fields for resources' configuration in OperandConfig
- `.spec` section has certain rule to determine whether or not allow overwriting the data.
- `.resources` section does not have such rule, it should always allow overwriting the data.

This PR is to enable overwriting the data under `.resources` section when there are only default rules or no rules specified.


YCShen1010 (approved):
/lgtm

PR #1756, opened 2024-01-16
added cert label to keycloak cert secret

bitscuit:
in order to leverage pod restart logic in cs-operator, the label is necessary


Daniel-Fan (approved):
/lgtm

PR #1759, opened 2024-01-18
Add common to the list of theme types so it works in jar format

Jamstah:
Follow on from #1739 which didn't work if the theme was in a jar file.


Daniel-Fan (approved):
/lgtm

PR #1771, opened 2024-01-25
Set key token when creating the LSR secret

YCShen1010:
Need to set the key in the LSR secret for the Licensing service instance.

To prevent the error:
Screenshot 2024-01-21 at 6 44 25 PM


Daniel-Fan (approved):
/lgtm

PR #1788, opened 2024-02-06
Pin the  ODLM channel at v4.3 for CS 4.5+

YCShen1010:
fix issue: https://github.ibm.com/IBMPrivateCloud/roadmap/issues/62185

- The ODLM channel will be at `v4.2` for CPFS 4.2-4.4 release
- The ODLM channel will be pinned at version `v4.3` for each release starting from 4.5


Daniel-Fan (approved):
/lgtm

PR #1813, opened 2024-02-23
setup singleton ensures cm-maps created

bitscuit:
to ensure that a controlNamespace is always available for migration
to decouple from isolation needing to be run first

### DIFF
--- a/cp3pt0-deployment/common/migrate_singleton.sh
+++ b/cp3pt0-deployment/common/migrate_singleton.sh
@@ -68,6 +68,7 @@ function main() {
         fi
 
         if [[ $ENABLE_LICENSE_SERVICE_REPORTER -eq 1 ]]; then
+            isolate_license_service_reporter
             migrate_license_service_reporter
         fi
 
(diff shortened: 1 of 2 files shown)


Daniel-Fan:
@bitscuit 
I am wondering that if we should count the all namespace migration scenario that we did not and will not have this `common-service-maps` ConfigMap. Also we are going to migrate License service into `ibm-licensing` namespace.

cp3pt0-deployment/common/migrate_singleton.sh
```diff
@@ -87,26 +88,77 @@ function main() {
...
+        info "Expecting exactly one IBMLicenseServiceReporter in cluster.${count} found."
+        return 0
+    fi
+
+    local ns=$(echo "$lsr_cr" | cut -d ' ' -f1)
+
+    local return_value=$( ("${OC}" get crd ibmlicenseservicereporters.operator.ibm.com > /dev/null && echo exists) || echo fail)
+    if [[ $return_value == "exists" ]]; then
+
+        return_value=$("${OC}" get ibmlicenseservicereporters -A --no-headers | wc -l)
```
Daniel-Fan, line 112: Are we able to move the LSR number counting logic down to here, after we check LSR CRD?

```
return_value=$("${OC}" get ibmlicenseservicereporters -A --no-headers | wc -l)
if [[ $return_value -eq 0 ]]; then
   return 0
fi
if [[ count -ne 1 ]]; then
   info "Expecting exactly one IBMLicenseServiceReporter in cluster.${count} found."
   return 0
fi

# continue change PVC to Retain
```

Daniel-Fan (commented):
Hi @bitscuit 
Would you mind reviewing this? Thank you!

Daniel-Fan (approved):
/lgtm

PR #1815, opened 2024-02-28
Check scoped namespaces from nss ConfigMap instead of watch_namespace

YCShen1010:
Fix for issue: https://github.ibm.com/IBMPrivateCloud/roadmap/issues/62427

If the Namespace Scope operator does not immediately patch `WATCH_NAMESPACE` in the cs-operator CSV, we will then check the number of scoped namespaces from the `namespace-scope` ConfigMap before proceeding to clean up the NSS operator and CRs.

### DIFF
--- a/controllers/bootstrap/init.go
+++ b/controllers/bootstrap/init.go
@@ -1185,14 +1185,14 @@ func (b *Bootstrap) CleanNamespaceScopeResources() error {
 	if err != nil {
 		klog.Errorf("Failed to get %s configmap: %v", constant.NamespaceScopeConfigmapName, err)
 		return err
-	} else if nssCmNs == nil && err == nil {
-		klog.Infof("The %s configmap is not found in the %s namespace, skip cleaning up", constant.NamespaceScopeConfigmapName, b.CSData.OperatorNs)
+	} else if nssCmNs == nil {
+		klog.Infof("The %s configmap is not found in the %s namespace, skip cleaning the NamespaceScope resources", constant.NamespaceScopeConfigmapName, b.CSData.OperatorNs)
 		return nil
 	}
 
 	// If the topology is (NOT ALL NS Mode) and (NOT Simple) , return
 	if b.CSData.WatchNamespaces != "" && len(nssCmNs) > 1 {
-		klog.Infof("The topology is not All Namespaces Mode or Simple Topology, skip cleaning up")
+		klog.Infof("The topology is not All Namespaces Mode or Simple Topology, skip cleaning the NamespaceScope resources")
 		return nil
 	}
 
--- a/controllers/bootstrap/init.go
+++ b/controllers/bootstrap/init.go
@@ -1180,8 +1180,19 @@ func (b *Bootstrap) DeployCertManagerCR() error {
 // NamespaceScope resources include common-service, nss-managedby-odlm, nss-odlm-scope, and odlm-scope-managedby-odlm
 func (b *Bootstrap) CleanNamespaceScopeResources() error {
 
+	// get namespace-scope ConfigMap in operatorNamespace
+	nssCmNs, err := util.GetNssCmNs(b.Reader, b.CSData.OperatorNs)
+	if err != nil {
+		klog.Errorf("Failed to get %s configmap: %v", constant.NamespaceScopeConfigmapName, err)
+		return err
+	} else if nssCmNs == nil && err == nil {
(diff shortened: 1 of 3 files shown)


controllers/common/util.go
```diff
@@ -745,7 +748,7 @@ func GetCmOfNss(r client.Reader, operatorNs string) (*corev1.ConfigMap, error) {
 	nssConfigmap := &corev1.ConfigMap{}
 
 	for {
```
Daniel-Fan, line 750: Could we just remove this for loop? I feel that this is not needed anymore

controllers/common/util.go
```diff
@@ -757,7 +760,6 @@ func GetCmOfNss(r client.Reader, operatorNs string) (*corev1.ConfigMap, error) {
 		}); err == nil {
 			break
 		} else {
-			klog.Errorf("Failed to get configmap %v/%v: %v", operatorNs, constant.NamespaceScopeConfigmapName, err)
 			return nil, err
 		}
```
Daniel-Fan, line 764: If the for loop is removed, I believe we could simplify the code to
```
if err:= utilwait.PollImmediate(); err != nil {
    return nil, err
}
```

controllers/bootstrap/init.go
```diff
@@ -1180,8 +1180,19 @@ func (b *Bootstrap) DeployCertManagerCR() error {
...
+		return err
+	} else if nssCmNs == nil && err == nil {
+		klog.Infof("The %s configmap is not found in the %s namespace, skip cleaning up", constant.NamespaceScopeConfigmapName, b.CSData.OperatorNs)
+		return nil
+	}
+
 	// If the topology is (NOT ALL NS Mode) and (NOT Simple) , return
-	if b.CSData.WatchNamespaces != "" && len(strings.Split(b.CSData.WatchNamespaces, ",")) > 1 {
+	if b.CSData.WatchNamespaces != "" && len(nssCmNs) > 1 {
+		klog.Infof("The topology is not All Namespaces Mode or Simple Topology, skip cleaning up")
```
Daniel-Fan, line 1195:
```suggestion
		klog.Infof("The topology is not All Namespaces Mode or Simple Topology, skip cleaning the NamespaceScope resources")
```

controllers/bootstrap/init.go
```diff
@@ -1180,8 +1180,19 @@ func (b *Bootstrap) DeployCertManagerCR() error {
 // NamespaceScope resources include common-service, nss-managedby-odlm, nss-odlm-scope, and odlm-scope-managedby-odlm
 func (b *Bootstrap) CleanNamespaceScopeResources() error {
 
+	// get namespace-scope ConfigMap in operatorNamespace
+	nssCmNs, err := util.GetNssCmNs(b.Reader, b.CSData.OperatorNs)
+	if err != nil {
+		klog.Errorf("Failed to get %s configmap: %v", constant.NamespaceScopeConfigmapName, err)
+		return err
+	} else if nssCmNs == nil && err == nil {
+		klog.Infof("The %s configmap is not found in the %s namespace, skip cleaning up", constant.NamespaceScopeConfigmapName, b.CSData.OperatorNs)
```
Daniel-Fan, line 1189:
```suggestion
		klog.Infof("The %s configmap is not found in the %s namespace, skip cleaning the NamespaceScope resources", constant.NamespaceScopeConfigmapName, b.CSData.OperatorNs)
```

controllers/bootstrap/init.go
```diff
@@ -1180,8 +1180,19 @@ func (b *Bootstrap) DeployCertManagerCR() error {
 // NamespaceScope resources include common-service, nss-managedby-odlm, nss-odlm-scope, and odlm-scope-managedby-odlm
 func (b *Bootstrap) CleanNamespaceScopeResources() error {
 
+	// get namespace-scope ConfigMap in operatorNamespace
+	nssCmNs, err := util.GetNssCmNs(b.Reader, b.CSData.OperatorNs)
+	if err != nil {
+		klog.Errorf("Failed to get %s configmap: %v", constant.NamespaceScopeConfigmapName, err)
+		return err
+	} else if nssCmNs == nil && err == nil {
```
Daniel-Fan, line 1188: I believe we do not need to check `err == nil`, it has been checked by `err != nil` already.
```suggestion
	} else if nssCmNs == nil {
```

Daniel-Fan (approved):
/lgtm

PR #1818, opened 2024-02-29
Pin the ODLM channel at v4.3 for CS 4.6+

YCShen1010:
CS version `4.4.1` will become `4.5.0`, and `4.5.0` will then progress to `4.6.0`,so pin the channel of ODLM at version 4.3 for CS 4.6+

- The ODLM channel will be the same as the version of Common Service for 4.0 and 4.1 release
- The ODLM channel will be at `v4.2` for CPFS 4.2-4.5 release
- The ODLM channel will be pinned at version `v4.3` for each release starting from 4.6


Daniel-Fan (approved):
/lgtm

PR #1831, opened 2024-03-05
create/update NSS CR before updating NSS operator

YCShen1010:
fix for issue: https://github.ibm.com/IBMPrivateCloud/roadmap/issues/61531

The fix is used to recreate/update the `common-service` NamespaceScope CR promptly after cleaning NSS resources to prevent obtaining an incorrect number of namespace members caused by the failure of the NSS operator upgrade the next time the script runs.


Daniel-Fan (approved):
/lgtm

PR #1832, opened 2024-03-05
Continue isolation when the resource is not found

Daniel-Fan:
Self-isolation should proceed when certain resources are not found.

For example
- if ODLM subscription is not found, self isolation should proceed because it is able to assume that ODLM is scaled down(not installed).
- if License Service Reporter's PVC is not found, self isolation should skip it and proceed because there is no LSR PVC to update.


YCShen1010 (approved):
/lgtm

PR #1845, opened 2024-03-08
Remove problematic code to reduce error

YCShen1010:
As we introduce `error trapping in the $FUNCNAME`, we need to remove the problematic code to reduce errors that could lead to script termination


Daniel-Fan (approved):
/lgtm

PR #1855, opened 2024-03-11
update velero post hook annotation

qpdpQ:
remove database and secrets separately, otherwise it might cause permission issue because of `readonlyrootfilesystem`


Daniel-Fan (approved):
/lgtm

PR #1868, opened 2024-03-19
Preload IM client credentials for admin

rwhundley:
The `platform-oidc-credentials` and `oauth-client-secret` Secrets both contain client registrations that are expected to persist through an isolated upgrade and migration from MongoDB to EDB. Not having these secrets established before the IAM Operator starts leads to split-brain problems between the Operator and the platform-auth-service Operand.


Daniel-Fan (approved):
/lgtm

PR #1874, opened 2024-03-23
Delete NSS resources by patching finalizer

YCShen1010:
fix for issue: https://github.ibm.com/IBMPrivateCloud/roadmap/issues/62702

If the NSS operator is automatically deleted during migration, we will patch the finalizer for the remaining NSS resources that also need to be deleted


Daniel-Fan (approved):
/lgtm

PR #1881, opened 2024-03-28
Valero/OADP support for Keycloak's EDB

pgodowski:
Resolves https://github.ibm.com/IBMPrivateCloud/roadmap/issues/60711#issuecomment-74907167

cc @giacomoch

### DIFF
--- a/controllers/constant/odlm.go
+++ b/controllers/constant/odlm.go
@@ -809,6 +809,9 @@ spec:
       - apiVersion: postgresql.k8s.enterprisedb.io/v1
         data:
           spec:
+            inheritedMetadata:
+              annotations:
+                backup.velero.io/backup-volumes: pgdata,pg-wal
             description:
               templatingValueFrom:
                 objectRef:
@@ -855,6 +858,11 @@ spec:
             walStorage:
               size: 1Gi
         force: true
+        annotations:
+          k8s.enterprisedb.io/addons: ["velero"]
+          k8s.enterprisedb.io/snapshotAllowColdBackupOnPrimary: enabled
+        labels:
+          foundationservices.cloudpak.ibm.com: keycloak
         kind: Cluster
         name: keycloak-edb-cluster
 `


giacomoch (approved):

Daniel-Fan (approved):
/lgtm

controllers/constant/odlm.go
```diff
@@ -855,6 +858,11 @@ spec:
             walStorage:
               size: 1Gi
         force: true
+        annotations:
+          k8s.enterprisedb.io/addons: ["velero"]
```
bluzarraga, line 862: This actually needs to be '["velero"]' otherwise it is invalid syntax

bluzarraga (commented):
Do we also want to make the same changes for IM's edb deployment or do they share this config?

controllers/constant/odlm.go
```diff
@@ -855,6 +858,11 @@ spec:
...
+          k8s.enterprisedb.io/snapshotAllowColdBackupOnPrimary: enabled
+        labels:
+          foundationservices.cloudpak.ibm.com: keycloak
```
bluzarraga, line 865: less important but I think this should be keycloak-data to match what is already present on the scripts side

controllers/constant/odlm.go
bluzarraga: This was already changed here https://github.com/IBM/ibm-common-service-operator/commit/c6afe100ce0dc5aa9a24c2a8ad5e2e01df3ca978

controllers/constant/odlm.go
bluzarraga: that's fine, we can have two separate labels to differentiate between pgdump approach and velero addon approach since they shouldn't conflict. It is easier to change on the scripts side anyways so once we change to fixed EDB version we can make the change on the scripts side.

bluzarraga:
> Do we also want to make the same changes for IM's edb deployment or do they share this config?

It's my understanding that we are waiting on new EDB version with bug fix before including this change for IM so disregard this question for now.

Daniel-Fan:
> Do we also want to make the same changes for IM's edb deployment or do they share this config?

My understand is that
1. Yes, we will use those annotations into IM's EDB deployment for B&R process when we switch to velero add-on after a new EDB release containing the fix.
2. Can those annotations for IM's EDB deployment exist out of the box, like what we did for Keycloak's EDB deployment?
   - The questions arises because those annotations might be conflicted with what CPD team will add into IM's EDB deployment.
   - However, CS/ODLM allow annotations and label overwritten, meaning even those velero annotation exists on IM's EDB deployment initially, CPD team could overwrite it by updating CS CR.

bluzarraga:
> those annotations might be conflicted with what CPD team will add into IM's EDB deployment.

It would depend on how cpd adds these annotations. If they also include them in the operand config, in my (brief) testing, you could have (roughly, it's been a minute since I looked at it):`k8s.enterprisedb.io/addons: '["velero","external-backup-adapter"]'` and as long as you give the hooks specified under the external-backup-adapter configuration different names than the typical velero ones (ie anything other than `post.hook.backup.velero.io/command`, `pre.hook.backup.velero.io/command`) they should not conflict.

But if CPD is overwriting the annotation field `k8s.enterprisedb.io/addons` after the cluster object is created then they will likely remove Bedrock's velero addon.

giacomoch:
Note sure that it matters, but do note that the `Cluster` in the CR should not be changed for any of the IM and Zen work as we dont share the `Cluster`.

I imagine the above complication arises given both Zen and IM use the same `Cluster` CR. Just want to make sure the Keycloak one is not changed :)

bluzarraga:
correct, this would not affect keycloak

PR #1883, opened 2024-03-28
Remove default size profile and instance from OperandConfig template

YCShen1010:
Issue: https://github.ibm.com/ibmprivatecloud/roadmap/issues/61284

After [implementing](https://github.com/IBM/operand-deployment-lifecycle-manager/pull/1019) the ODLM's capability to deep merge Kubernetes resources logic, we could remove the default size profile and instance in OperandConfig template. Therefore, we should observe a consistent change in the number of instances during common-service reconciliation, rather than fluctuating between 3, 1, and 3.

Before removing the hardcoded size and instance:
1. When deploying KeyCloak for the first time, the instance created in KeyCloak CR would be `1`.
2. Then if the user configures a large size in CS CR, the instance will change to `3`.
3. If it is upgraded to a new version, we will observe the instance changing back to `1`, as CS applies the default value of 1 from OperandConfig during reconciliation.
4. Finally, it will ultimately change to `3`, as the size is configured as large in the CS CR

After removing:
1. The instance would always remain `3` without alteration, as long as the user configures it to a large size.

### DIFF
--- a/controllers/constant/odlm.go
+++ b/controllers/constant/odlm.go
@@ -641,7 +641,6 @@ spec:
               tlsSecret: cs-keycloak-tls-secret
             ingress:
               enabled: false
-            instances: 1
             unsupported:
               podTemplate:
                 spec:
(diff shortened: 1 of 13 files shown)


Daniel-Fan (approved):
/lgtm

PR #1896, opened 2024-04-02
Create OperatorConfig for EDB when specified

bitscuit:
Only allows for configuring replicas. The affinity and topologyspreadconstraints are hardcoded. 

Current implementation assumes that CommonService CR only has one entry for operatorConfig, which is enough for CP4I.

Current implementation will result in the most recently reconciled CommonService CR's operatorConfig to take effect, i.e. overwrites will happen if there are multiple CommonService CRs. Intentional for now since will take too long to implement max size rule.

Needed to create a special branch for ODLM to include the new OperatorConfig APIs based off of the v1.19.0 release so that cs-operator can use the type without updating any other modules https://github.com/IBM/operand-deployment-lifecycle-manager/tree/operatorconfig-v1.19.0


Daniel-Fan (approved):
/lgtm

PR #1901, opened 2024-04-03
add velero plugin to im edb cluster config by default

bluzarraga:
By default, we want to enable the velero backup addon for IM EDB cluster. This way, in all use cases except CP4D, IM EDB will follow the officially supported BR process for EDB. In the CP4D case where IM EDB will be sharing cluster config with zen, CP4D will overwrite the annotation with their own via the CommonService CR. This will remove the ability to backup IM via velero plugin so we will have a fork in the instructions to specify using the pgdump approach instead. I stuck with calling the velero label cs-db instead of cs-db-data as we already have in scripts to facilitate this use case based fork.

Discussion welcome.

### DIFF
--- a/controllers/constant/odlm.go
+++ b/controllers/constant/odlm.go
@@ -990,8 +990,16 @@ spec:
         kind: Cluster
         name: common-service-db
         force: true
+        annotations:
+          k8s.enterprisedb.io/addons: ["velero"]
+          k8s.enterprisedb.io/snapshotAllowColdBackupOnPrimary: enabled
+        labels:
+          foundationservices.cloudpak.ibm.com: cs-db
         data:
           spec:
+            inheritedMetadata:
+              annotations:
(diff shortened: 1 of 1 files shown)


bluzarraga:
Meant to add this comment to this PR but accidentally placed it on 1902: https://github.com/IBM/ibm-common-service-operator/pull/1902#issuecomment-2035536266
Copy/pasting for clarity here:
```
A couple reasons to put this on hold after discussion with Yanni:
- DCUT has already passed
- velero bug has not technically been solved on the EDB side
- not fully understood how the overwriting of the addon annotation will work out, currently just have a theoretical understanding
- the only adopters of 4.6 are CPD who are most affected by this change

essentially, more can go wrong by including this early than can be gained. Better to target in 4.6.1
```

bluzarraga:
defer to 4.6.2

Daniel-Fan (approved):
/lgtm

PR #1902, opened 2024-04-03
add velero plugin based labels, clarify labels

bluzarraga:
The velero plugin based BR is currently setup to use the labels `keycloak` and `cs-db` for velero to notice and process the keycloak and IM EDB clusters for backup. This is to differentiate between the velero plugin approach (facilitated and officially supported by the EDB operator) and the pgdump approach we implemented outside of the operator. The pgdump approach uses `keycloak-data` and `cs-db-data` for comparison. I have added comments to these four values as well as the similarly structured License Service Reporter and Zen values to provide clarity on their usage.

### DIFF
--- a/velero/schedule/schedule-common-services.yaml
+++ b/velero/schedule/schedule-common-services.yaml
@@ -38,7 +38,7 @@ spec:
         - crd
         - lsr #non-data license service reporter resources
         - lsr-data #license service reporter data
-        - cs-db-data #pgdump
-        - cs-db #velero plugin
-        - keycloak-data #pgdump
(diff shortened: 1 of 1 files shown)


bluzarraga:
A couple reasons to put this on hold after discussion with Yanni:
- DCUT has already passed
- velero bug has not technically been solved on the EDB side
- not fully understood how the overwriting of the addon annotation will work out, currently just have a theoretical understanding
- the only adopters of 4.6 are CPD who are most affected by this change

Daniel-Fan (approved):
/lgtm

PR #1906, opened 2024-04-03
Update CRD to be forward compatible

Daniel-Fan:
Add preserve unknown field for CS CR field .`spec.services[*]`

```yaml
kind: CommonService
metadata:
  name: common-service
spec:
  license:
    accept: false
  operatorNamespace: cs-condition
  servicesNamespace: cs-condition
  size: starterset
  services:
     - name: ibm-im-operator
       resources:
          - xxxx
       spec:
            xxxx
       ...  # Any arbitrary field
```


bitscuit (approved):
/lgtm

PR #1915, opened 2024-04-09
reference cpfs version in Keycloak pod template

Daniel-Fan:
issue: https://github.ibm.com/IBMPrivateCloud/roadmap/issues/62819

Keycloak CR's pod will have a reference on `annotations.version` from `cs-keycloak-theme` ConfigMap.
When the `cs-keycloak-theme` ConfigMa is updated in a new release(new theme update), Keycloak pod will be patched with a new annotation `cpfsVersion: 4.6.0`. As a result, the pod will be restarted to pick up the new theme styling.

### Test
- Keycloak CR will be updated with latest pod template
   Screenshot 2024-04-09 at 2 21 45 PM
- Keycloak pod will be updated with new annotation
  Screenshot 2024-04-09 at 2 22 23 PM

### DIFF
--- a/controllers/constant/odlm.go
+++ b/controllers/constant/odlm.go
@@ -645,6 +645,16 @@ spec:
               enabled: false
             unsupported:
               podTemplate:
+                metadata:
+                  annotations:
+                    cpfsVersion:
+                      templatingValueFrom:
+                        objectRef:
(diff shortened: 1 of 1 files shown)


bitscuit (approved):
/lgtm

PR #1916, opened 2024-04-09
Resolve replica number reversion issue

YCShen1010:
Fix for issue: https://github.ibm.com/IBMPrivateCloud/roadmap/issues/62729

#### Issue description
This intermittent issue occurs when the profile size is configured to anything other than `small/starterset`. After an isolate upgrade, the replica number may occasionally revert to 1, whereas it should actually match the configured profile size.

#### Reason for failure
If multiple CommonServices with varying profile sizes are created within the same tenant, the CommonService operator should prioritize selecting the larger size and higher limitations. This selection should be applied to the OperandConfig consistently, regardless of the order in which the CS CRs were created. This approach ensures that the values chosen are always larger and correct, rather than being chosen randomly.

#### How to test
Test image: quay.io/yuchen_shen/cs_operator:cs_condition
Apply multiple CS CR with different size, in the OperandConfig will always see the largest value of replica/limit.


Daniel-Fan (approved):
/lgtm

PR #1917, opened 2024-04-10
condition check for OperandRegistry version comparison

YCShen1010:
fix for issues:
https://github.ibm.com/ibmprivatecloud/roadmap/issues/61040
https://github.ibm.com/IBMPrivateCloud/roadmap/issues/61273

Ensure that the version of OperandRegistry is greater than or equal `4.0.0` before proceeding to clean up NSS resources.

### DIFF
--- a/controllers/bootstrap/init.go
+++ b/controllers/bootstrap/init.go
@@ -1246,6 +1246,17 @@ func (b *Bootstrap) CleanNamespaceScopeResources() error {
 			klog.Errorf("Failed to get common-service OperandRegistry: %v", err)
 			return err
 		}
+
+		// Check if there is v4 OperandRegistry exists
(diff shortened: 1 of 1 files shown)


controllers/bootstrap/init.go
```diff
@@ -1246,6 +1246,17 @@ func (b *Bootstrap) CleanNamespaceScopeResources() error {
 			klog.Errorf("Failed to get common-service OperandRegistry: %v", err)
 			return err
 		}
+
+		// Check if there is v4 OperandRegistry exists
+		if operandRegistry.Annotations != nil {
```
Daniel-Fan, line 1251: Could we handle the case where `opereandRegistry == nil` when `error == nil`

Daniel-Fan (approved):
/lgtm

PR #1919, opened 2024-04-11
annotate Keycloak pod with theme version instead of CPFS version

Daniel-Fan:
issue: https://github.ibm.com/IBMPrivateCloud/roadmap/issues/62819#issuecomment-76863148

We annotate the Keycloak pod with the theme version. As long as the theme does not change, keycloak pod does not need to be restarted. It would be independent from CPFS upgrade.

### Test
- CS operator CSV will contain the theme version annotation
   Screenshot 2024-04-10 at 10 51 10 PM

- `cs-keycloak-theme` ConfigMap will contain the theme version annotation.
  Screenshot 2024-04-10 at 10 51 32 PM

- Keycloak server pod will contain the theme version annotation.
  Screenshot 2024-04-10 at 10 51 58 PM


YCShen1010 (approved):
/lgtm

PR #1924, opened 2024-04-16
Match scripts branch to scripts-dev branch for BR

bluzarraga:
This PR looks bigger than it really is but essentially all the development work for BR has been merged into scripts-dev and not scripts but all of the documentation points to the scripts branch (by design since we already publicly exposed this branch in documentation).

This means we need to make sure all of the changes made for 4.6 are reflected in the scripts branch to match up with documentation precedent.

I have replaced the velero directory from the scripts branch with the one from the scripts-dev branch resulting in the changes present in this pr. I have also gone through and verified the changes made do not outright break anything. I also updated the default cpfs utils image to 4.6.


Daniel-Fan (approved):
/lgtm

PR #1925, opened 2024-04-16
[4.6.1] Check if third party cert-manger deployed before applying catalogsource

YCShen1010:
issue: https://github.ibm.com/ibmprivatecloud/roadmap/issues/61575

### Context 
We introduce a flag `--check-cert-manager` to run the cert-manager deployment check before user applying the CatalogSource and returning the indicator value:

- return 0: there is no cert-manger installed in the whole cluster
- return 1: there is IBM v3 or v4 cert-manger installed
- return2: third-party cert-manger is deployed

### DIFF
--- a/cp3pt0-deployment/setup_singleton.sh
+++ b/cp3pt0-deployment/setup_singleton.sh
@@ -322,7 +322,7 @@ function install_cert_manager() {
             return 0
         fi
     else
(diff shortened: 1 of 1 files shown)


Daniel-Fan (approved):
/lgtm

PR #1927, opened 2024-04-18
remove wait for job function leftover from keycloak

bluzarraga:
It is currently not being called but should still be removed as it is unecessary


Daniel-Fan (approved):
/lgtm

PR #1930, opened 2024-04-19
revert cpfs utils image to 4.5 to match repo

bluzarraga:
the cpfs utils repo is still using 4.5 for the new image so need to revert the image values in the scripts repo


Daniel-Fan (approved):
/lgtm

PR #1937, opened 2024-04-22
add keycloak label to schedule

bluzarraga:
We use a different label for the pgdump (keycloak-data) and cluster annotation (keycloak) BR approaches. We need to make sure both are covered in the schedule yaml file.


Daniel-Fan (approved):
/lgtm

PR #1939, opened 2024-04-22
Bump 4.7.0 for CPFS 4.7.0 release

Daniel-Fan:
Also remove image reference for `cpfs-utils` which is only shipped with 4.6.x stream


YCShen1010 (approved):
/lgtm

PR #1946, opened 2024-04-25
match dev to scripts, add oidc rerun command

bluzarraga:
Need to rerun the oidc client registration job after restoring cs db data. Better to include it as part of the script than to add another step to the documentation. I also matched the scripts-dev script to what is currently in the scripts branch as I did a little cleanup when merging in https://github.com/IBM/ibm-common-service-operator/pull/1924


Daniel-Fan (approved):
/lgtm

PR #1947, opened 2024-04-25
add rerun oidc job command

bluzarraga:
Mostly same changes as https://github.com/IBM/ibm-common-service-operator/pull/1946. Need to rerun oidc registration job after restore is completed and it is much better to add this step to the script than to add to documentation.


Daniel-Fan (approved):
/lgtm

PR #1951, opened 2024-04-26
Fix zen extension cleanup logic

bluzarraga:
For issue https://github.ibm.com/IBMPrivateCloud/roadmap/issues/62738.

Increase the zen timeout to reduce the likelihood of timing out during a zen data restore.

Change the extension delete api request to be in line with what zen core api expects by replacing `/` with `%2F` after consulting with zen team.


Daniel-Fan (approved):
/lgtm

PR #1952, opened 2024-04-29
[scripts-dev] Update network policy

Daniel-Fan:
- Remove networkpolicies which has empty `.spec.podSelector:{}`.
    - An empty `.spec.podSelector:{}` means that the effect of this network policies applies to all the pods in the same namespace.
    - Generally we should always have network policy explicitly state `.spec.podSelector:{}` as best practice.
- Fix argument shift when using `-u` or `-e` flag, it should NOT shift to next argument since there is namespace variable assignment


YCShen1010 (approved):
/lgtm

PR #1958, opened 2024-05-01
removed internal OperandRequest from edb-keycloak

bitscuit:
- for cloud-native-postgresql, which reduced performance

added back EDB licensing job for edb-keycloak because internal OperandRequest is removed


Daniel-Fan (approved):
/lgtm

PR #1968, opened 2024-05-02
set max connection value as a string

YCShen1010:
Fix issue: https://github.ibm.com/IBMPrivateCloud/roadmap/issues/63360

The default value of max connection in EDB cluster CR should be a string type


Daniel-Fan (approved):
/lgtm

PR #1971, opened 2024-05-03
Update ODLM CRs after ODLM has a minor version upgrade

Daniel-Fan:
issue: https://github.ibm.com/IBMPrivateCloud/roadmap/issues/61732

CS operator will wait for ODLM minor version upgrade before proceeding to next step. e.g., ODLM `v4.2.3` -> `v4.3.0` with a channel switch `v4.2` to `v4.3`.
Also when CS operator detects an ODLM minor version upgrade, meaning there are potential CRD changes, CS operator will force ODLM CRs(OperandConfig and OperandRegistry) update, to ensure that CRs are applied to the cluster with latest ODLM CRD changes.

The file `controllers/bootstrap/init_test.go` includes the unit test for two new added functions.
The end to end test was done as well on the cluster with image build.

### DIFF
--- a/controllers/bootstrap/init.go
+++ b/controllers/bootstrap/init.go
@@ -28,6 +28,7 @@ import (
 
 	utilyaml "github.com/ghodss/yaml"
 	olmv1alpha1 "github.com/operator-framework/api/pkg/operators/v1alpha1"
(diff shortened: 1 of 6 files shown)


YCShen1010 (approved):
/lgtm

PR #1976, opened 2024-05-06
update oidc rerun logic, correct bug

bluzarraga:
Previous implementation techincally worked but always output the following error:

> Error from server (Forbidden): jobs.batch "oidc-client-registration" is forbidden: cannot set blockOwnerDeletion if an ownerReference refers to a resource you can't set finalizers on: , <nil>

This updated implementation does not output any errors and now also includes actual wait logic to verify the job completes. Wait logic borrowed from keycloak script where a bug was found and updated.

### DIFF
--- a/velero/schedule/common-service-db/cs-db-br-script-cm.yaml
+++ b/velero/schedule/common-service-db/cs-db-br-script-cm.yaml
@@ -2,7 +2,7 @@ apiVersion: v1
 kind: ConfigMap
 metadata:
   name: cs-db-br-configmap
-  namespace: <cs-db namespace>
+  namespace: cs
   labels:
     foundationservices.cloudpak.ibm.com: cs-db-data
 data:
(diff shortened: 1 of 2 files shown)


velero/schedule/common-service-db/cs-db-br-script-cm.yaml
```diff
@@ -2,7 +2,7 @@ apiVersion: v1
 kind: ConfigMap
 metadata:
   name: cs-db-br-configmap
-  namespace: <cs-db namespace>
+  namespace: cs
```
Daniel-Fan, line 5: Hi @bluzarraga 
Do we need to update this back to `<cs-db namespace>`?

bluzarraga: yes, good catch!

Daniel-Fan (approved):
/lgtm

PR #2015, opened 2024-05-29
specify dump location with emptyDir volume

bluzarraga:
CP4BA ran into issue https://github.ibm.com/IBMPrivateCloud/roadmap/issues/63644 where they were seeing a failure with the mongodb-backup job created by the preload_data.sh script:
```
# oc logs mongodb-backup-fw6vv -n ibm-common-services
2024-05-28T19:05:09.097+0000	Failed: error dumping metadata: error creating directory for metadata file /dump/dump/platform-db: mkdir /dump/dump: permission denied
```
This failure was preventing successful upgrade. I went onto the cluster and added an emptyDir volume to each instance of a mongo restore or backup job and mounted it to the location where we try to write the mongodump out to "/dump/dump". After making this change on the cluster where the problem was seen, I saw the job and the preload_data.sh script complete successfully.

### DIFF
--- a/preload_data.sh
+++ b/preload_data.sh
@@ -353,8 +353,10 @@ function dumpmongo() {
   if [[ "$currentns" -ne "$FROM_NAMESPACE" ]]; then
     error "Cannot switch to $FROM_NAMESPACE"
   fi
-
+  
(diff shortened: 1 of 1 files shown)


preload_data.sh
```diff
@@ -370,6 +371,14 @@ spec:
       containers:
       - name: cs-mongodb-backup
         image: $ibm_mongodb_image
+        securityContext:
```
Daniel-Fan, line 374: Hi Ben, do we wanna remove this `securityContext`?

PR #2021, opened 2024-06-03
remove the generic pod definition

bluzarraga:
Security tool kept flagging the generic pod yaml for security issues. The yaml file is not being used anywhere so best to just remove it. It looks like the original function of the pod was to provide a single yaml for both mongo and zen BR but it was never picked up. We have talked about consolidating into one yaml but if we are to do that, it will need to be through the cpfs-utils image as that image already contains OC, YQ, and mongorestore/mongodump tools.


Daniel-Fan (approved):
/lgtm

PR #2023, opened 2024-06-04
Add info functions

YCShen1010:
Issue: https://github.ibm.com/IBMPrivateCloud/roadmap/issues/62014
Adding the info functions directly into the script instead of calling them from until.sh.


bitscuit (approved):
/lgtm

PR #2026, opened 2024-06-04
63639

bluzarraga:
Change the file location for the extensions.txt generated to keep track of all the zen extensions needing to be cleaned up on the restore side. When using the emptyDir for this (previous implementation) the directory is not included in the persistent storage so it is not included in the backup. This change ensures the extensions are copied over to persistent storage and therefore backed up


Daniel-Fan (approved):
/lgtm

PR #2035, opened 2024-06-13
Cpfs util job

bluzarraga:
This Pr is opened to address https://github.ibm.com/IBMPrivateCloud/roadmap/issues/63689. Essentially, these changes move the setup tenant script execution to a job instead of a deployment. This way, there is a clear start and end time allowing for cleanup of the permissions and related resources once the job is completed. The permissions for the job are pretty large in scope due to the nature of the setup_tenant script so it is imperative to reduce the amount of time they exist and can be accessed.

So instead of:
- creating a deployment on the backup cluster with full set of permissions
- backing up the deployment and relevant permissions, PVC, and configmap
- restoring deployment, PVC, configmap, and permissions on restore cluster
- exec'ing into deployed pod on restore cluster and triggering setup_tenant script

We now:
- deploy the cpfs util deployment with limited permissions on the backup cluster
- deploy the setup tenant job in a suspended state on backup cluster with all its permissions
- backup all the above
- restore the above onto restore cluster
- remove the `suspend` field from the job, triggering it to run
- can instruct the user to clean up these resources immediately

There are other somewhat related changes in this PR pertaining to the deploy-br-resources.sh script to facilitate the deployment and cleanup of the BR resources (not just for cpfs util and setup tenant). This script is not complete but is useful in testing at the moment so I have kept it in this PR. If it is more appropriate to remove it and add it to a separate PR I can do that too. This script is related to this issue https://github.ibm.com/IBMPrivateCloud/roadmap/issues/63828


bluzarraga:
The changes from this PR will require doc updates once merged into the _scripts_ branch (this pr is for scripts-dev)

PR #2036, opened 2024-06-17
Label the CommonService CRD so it can be selected for backup

Jamstah:

To backup common services with velero we ask the user to add a label to the backup the CommonService CRD. This user added label will be reset after every operator upgrade, so lets add a label to the CRD ourselves that the user can select.

See doc here: https://www.ibm.com/docs/en/cloud-paks/foundational-services/4.4?topic=fsbr-cloud-pak-foundational-services-backup-restore-clusters-single-instance-foundational-services

Fixes #

1. How the test is done?

### DIFF
--- a/bundle/manifests/operator.ibm.com_commonservices.yaml
+++ b/bundle/manifests/operator.ibm.com_commonservices.yaml
@@ -8,6 +8,7 @@ metadata:
     app.kubernetes.io/instance: ibm-common-service-operator
     app.kubernetes.io/managed-by: ibm-common-service-operator
     app.kubernetes.io/name: ibm-common-service-operator
+    foundationservices.cloudpak.ibm.com: crd
   name: commonservices.operator.ibm.com
 spec:
   group: operator.ibm.com
--- a/config/crd/bases/operator.ibm.com_commonservices.yaml
+++ b/config/crd/bases/operator.ibm.com_commonservices.yaml
@@ -4,6 +4,8 @@ kind: CustomResourceDefinition
 metadata:
   annotations:
     controller-gen.kubebuilder.io/version: v0.14.0
+  labels:
+    foundationservices.cloudpak.ibm.com: crd
   name: commonservices.operator.ibm.com
 spec:
   group: operator.ibm.com
--- a/api/v3/commonservice_types.go
+++ b/api/v3/commonservice_types.go
@@ -298,6 +298,7 @@ const (
 
 // +kubebuilder:object:root=true
 // +kubebuilder:subresource:status
+// +kubebuilder:metadata:labels="foundationservices.cloudpak.ibm.com=crd"
 // +operator-sdk:gen-csv:customresourcedefinitions.displayName="CommonService"
 
 // CommonService is the Schema for the commonservices API. This API is used to


bundle/manifests/operator.ibm.com_commonservices.yaml
```diff
@@ -8,6 +8,7 @@ metadata:
     app.kubernetes.io/instance: ibm-common-service-operator
     app.kubernetes.io/managed-by: ibm-common-service-operator
     app.kubernetes.io/name: ibm-common-service-operator
+    foundationservices.cloudpak.ibm.com: commonservice
```
giacomoch, line 11: we dont use this label, can we use the same keycloak label we already have for all the keycloak CRs?
```suggestion
    foundationservices.cloudpak.ibm.com: keycloak
```

giacomoch (changes requested):
Consistent labelling with what we already label all CS resources

bundle/manifests/operator.ibm.com_commonservices.yaml
pgodowski: We cannot say that CommonSvcs CD is labelled with Keycloak, since it's not Keycloak :)

pgodowski (approved):
@bluzarraga pls take a look- thx

bundle/manifests/operator.ibm.com_commonservices.yaml
Jamstah: We'll need to add the value to our backup doc, I agree with Piotr that its not just keycloak.

giacomoch: we're not backing up Common services though, we're only backing up keycloak, this might imply it's backing up more than just Keycloak. I'll drop it though, semantic in the end.

Jamstah: If we want to support that, then we'd need something like:
`foundationservices.cloudpak.ibm.com/keycloak: backup`

Otherwise if you wanted to backup another common services component you wouldn't be able to add a second value to the label.

It would have been good to have `foundationservices.cloudpak.ibm.com/component: commonservice` but we missed that for the keycloak label.

Jamstah: We should probably have done a design session on the labels sooner :)

giacomoch: This is just what CS did for their own taxonomy. They wanted us to follow their convention not ours. I think we should follow the taxonomy of the product but here we are :)

config/crd/bases/operator.ibm.com_commonservices.yaml
```diff
@@ -4,6 +4,8 @@ kind: CustomResourceDefinition
 metadata:
   annotations:
     controller-gen.kubebuilder.io/version: v0.14.0
+  labels:
+    foundationservices.cloudpak.ibm.com: commonservice
```
bluzarraga, line 8: we use `foundationservices.cloudpak.ibm.com: crd` as a label for the crd in existing docs https://www.ibm.com/docs/en/cloud-paks/foundational-services/4.7?topic=fsbr-cloud-pak-foundational-services-backup-restore-clusters-single-instance-foundational-services

bluzarraga: We could technically use this instead since our command in our docs specifies `--overwrite` and/or we could update our docs/restore resources but going strictly on what's already in place, we would prefer crd instead of commonservice

Jamstah: If you prefer that though, I don't mind. Restoring any of the CRDs shouldn't be a problem as they're all linked to the subscription version (in principle).

giacomoch: I'd vote the CR name as that's what we do for all the intregration CRs

bluzarraga: Let me test what we discussed in https://github.ibm.com/IBMPrivateCloud/roadmap/issues/63820. If we can include them in the same restore then it shouldn't matter that the labels are the same (and we may update our process to reflect that). Otherwise I would prefer `crd`

Jamstah: @bluzarraga ping on this one, did you test it?

bluzarraga: Yes, it does work but after discussing with the rest of the team we would prefer the label be `crd` instead of `commonservice`

Jamstah: PR updated to use CRD.

I'd be interested to know why you chose to go with crd instead of commonservice, but feel free to merge when ready (and please label for backport to SC2.)

bluzarraga:
Job failure seems to be consistent with cs operator and not indicative of a problematic change. Adding these labels shouldn't be breaking anything so will merge

PR #2056, opened 2024-06-24
[4.0-4.5] spectrum fusion example recipe 

qpdpQ:
example recipe for spectrum fusion with mongodb and zen


bluzarraga (changes requested):
Once I got rid of the typo I pointed out, I got it working. Lets clean that up and then we can merge it in

bluzarraga (approved):

PR #2062, opened 2024-06-28
fix for unbound variable when namespace-scope cm is missing

qpdpQ:
unbound variable issue, in rerun namespace-scope cm could missing
Fixes # https://github.ibm.com/IBMPrivateCloud/roadmap/issues/64024

### DIFF
--- a/cp3pt0-deployment/common/utils.sh
+++ b/cp3pt0-deployment/common/utils.sh
@@ -877,7 +877,7 @@ EOF
 function update_cscr() {
     local operator_ns=$1
     local service_ns=$2
-    local nss_list=$3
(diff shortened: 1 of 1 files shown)


cp3pt0-deployment/common/utils.sh
```diff
@@ -988,7 +988,7 @@ function cleanup_cp2() {
 # clean up webhook deployment and webhookconfiguration
 function cleanup_webhook() {
     local control_ns=$1
-    local nss_list=$2
+    local local nss_list=${2:-""}
```
bitscuit, line 991:
```suggestion
    local nss_list=${2:-""}
```

local was repeated

PR #2063, opened 2024-06-28
fix for unbound variable when namespace-scope cm is missing

qpdpQ:
fix for unbound variable when namespace-scope cm is missing
Fixes # https://github.ibm.com/IBMPrivateCloud/roadmap/issues/64024


bitscuit (approved):
/lgtm

PR #2067, opened 2024-06-28
Improve the error msg for private catalogsource validating

YCShen1010:
When running the `setup_singleton.sh` with `-- enable-private-catalog` flag set, the error message of CatalogSource validation has been updated 

Improvement：

- When the `--enable-private-catalog` flag is set, there is a warning message that prompts the namespaces of singleton operators should be in the same namespace as the singleton CatalogSource.
- If the specific CatalogSource in its namespace is not found in the cluster, then the script will try to get an available global CatalogSource and will error out if there is more than one CatalogSource available in the entire cluster.

Issue: https://github.ibm.com/IBMPrivateCloud/roadmap/issues/63998#issuecomment-83385625


Daniel-Fan (approved):
/lgtm

PR #2085, opened 2024-07-09
Update events permission in nss managed role

YCShen1010:
Copy the events-related permissions collected from the 4.4 release into the 4.6 NSS managed role YAML to fix the permission issue hit in the upgrade path.

Fixes https://github.ibm.com/IBMPrivateCloud/roadmap/issues/64035#issuecomment-84268980

`nss-maanged-bedrock-core-role.yaml` is the template role that needs to be applied after each upgrade done by the script.


Daniel-Fan (approved):
/lgtm

PR #2086, opened 2024-07-09
edit tmp files instead of actual so the script can be rerun multiple times

bluzarraga:
https://github.ibm.com/IBMPrivateCloud/roadmap/issues/63828

Ability to deploy the peripheral resources required for BR of cs db, mongodb, zen 4 & 5, and keycloak. Also deploys the utility resources used in SF.


bluzarraga:
ready for review now

velero/schedule/deploy-br-resources.sh
```diff
@@ -28,6 +28,7 @@ function main() {
   if [[ $CLEANUP == "true" ]]; then
     cleanup
   else
+    mkdir tmp
```
Daniel-Fan, line 31: Hi @bluzarraga 
Just wanna confirm that if we are running the script in a k8s pod container, the pod has the ability to create local directory `tmp` in current path.

If it does not have such permission, we could consider using existing `/tmp` directory in the container's local filesystem, and create sub folder `mkdir /tmp/common-service-db` to host the files.

bluzarraga: yeah originally that's what I was doing but the script also has to be able to run outside of a container. I am planning on specifically including an emptydir in the path in the container where this script will be running to be able to create this dir

Daniel-Fan (approved):
/lgtm

PR #2087, opened 2024-07-09
Overwrite nss managed role during upgrade

YCShen1010:
The script will loop through all the namespaces from `.spec.namespaceMembers` in nss `common-service`, and overwrite `nss-managed-role` YAML file in each ns if it already exists with passed-in role when the `--with-minimal-rbac` flag is on.

Fixes: https://github.ibm.com/IBMPrivateCloud/roadmap/issues/64035


Daniel-Fan (approved):
/lgtm

PR #2088, opened 2024-07-10
enhancement on cp2 resources cleanup

qpdpQ:
enhancement on cp2 control namespace cleanup
#### case1: all namespace cleanup
in this case, we cleanup podpresets CR and secretshare CR in all namespaces
#### case2: tenant namespace cleanup
in this case, we cleanup podpresets CR and secretshare CR in tenant namespaces
#### case3: edge case(in rerun we not found namespace-scope cm)
in this case, since namespace-scope cm is empty it means there is only one cs-operator in the cluster, so it is safe to cleanup  podpresets CR and secretshare CR in all namespace

Fixes # https://github.ibm.com/IBMPrivateCloud/roadmap/issues/64063

### DIFF
--- a/cp3pt0-deployment/common/cleanup_cp2_control.sh
+++ b/cp3pt0-deployment/common/cleanup_cp2_control.sh
@@ -33,7 +33,7 @@ function main() {
     cleanup_NamespaceScope $CONTROL_NS
 
     # cleanup webhookc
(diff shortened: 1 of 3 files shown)


Daniel-Fan (approved):
/lgtm

PR #2092, opened 2024-07-15
fix contorl_ns not found

qpdpQ:
return error when contorl namespace not found and add yq arguments
Fixes # https://github.ibm.com/IBMPrivateCloud/roadmap/issues/63694


Daniel-Fan (approved):
/lgtm

PR #2114, opened 2024-08-01
Add user management bindinfo

bluzarraga:
I have added the necessary bindinfo for user management operator in a way mirroring keycloak. I am still trying to get the operand config to populate the operandbindinfo into the operandconfig as well as automatically create the operandbindinfo when ibm-user-management-operator is requested.

Currently, the secret `mcsp-im-integration-details` is called `mcsp-im-integration-api-key` but it's my understanding that IM is changing the name of the secret to be more generic so this may need to be updated if they do not end up calling it `mcsp-im-integration-details`.


YCShen1010 (approved):
/lgtm

PR #2125, opened 2024-08-02
add rerun parameter

bluzarraga:
Add a parameter that allows for re-running of the preload data script without added manual cleanup


Daniel-Fan (approved):
/lgtm

PR #2154, opened 2024-08-20
update cs module version to v4

bluzarraga:
To facilitate changes required for https://github.ibm.com/IBMPrivateCloud/roadmap/issues/64551, we need to update the module version for the cs operator to v4 so we can specify v4 tags and releases when importing in go.mod. Should be similar to this ODLM pr that did the same https://github.com/IBM/operand-deployment-lifecycle-manager/pull/1039/files


Daniel-Fan (approved):
/lgtm

PR #2158, opened 2024-08-22
hide error message in retry

qpdpQ:
hide error message in retry patch subscription
Fixes # https://github.ibm.com/IBMPrivateCloud/roadmap/issues/64558


Daniel-Fan:
Hi @qpdpQ 
Would you help to check if using `2>/dev/null` will be more elegant?

This is the reference: https://unix.stackexchange.com/questions/192042/how-do-i-suppress-error-messages-from-cp#:~:text=To%20suppress%20error%20output%20in,specific%20construct%20may%20vary%20slightly.

Daniel-Fan (approved):
/lgtm

PR #2168, opened 2024-08-28
added opreg entry for dev test new keycloak

bitscuit:
unblock any manual testing of newer keycloak versions


Daniel-Fan (approved):
/lgtm

PR #2181, opened 2024-09-06
add fallback channels for EDB operator

Daniel-Fan:
Add fallback channels for `common-service-postgresql`, `edb-keycloak` and `cloud-native-postgresql-v1.22`, so they could tolerate both `stable` and `stable-v1.22` channels.

issue: https://github.ibm.com/IBMPrivateCloud/roadmap/issues/64744


bitscuit (approved):
/lgtm

PR #2186, opened 2024-09-10
added new flag to indicate whether service is user managed

bitscuit:
How to test

1. Deploy common-service-operator from catalogs with 4.6.6+
2. Build this PR `make build-dev-image`
3. Change image in CSV on cluster to use built image
4. Update CommonService CR with following example
5. Verify OperandRegistry contains `userManaged: true` for all postgresql entries

CommonService CR example

```yaml
kind: CommonService
metadata:
  name: common-service
  labels:
    foundationservices.cloudpak.ibm.com: commonservice
spec:
  license:
    accept: false
  operatorConfigs:
    - name: common-service-postgresql
      userManaged: true
  operatorNamespace: byo
  servicesNamespace: byo
  size: starterset
```


Daniel-Fan (approved):
/lgtm

PR #2187, opened 2024-09-12
obfuscate metrics secret values

bluzarraga:
Metrics pod user and secret values are exposed in the script. We can take the values from the existing mongo in the original namespace instead and obfuscate the value in the script itself.
Fixes #

1. How the test is done?
Successful run of preload data script


bluzarraga:
Tested locally and this change has no impact on preload data script working

Daniel-Fan (approved):
/lgtm

PR #2197, opened 2024-09-16
[CD] add BTS with version in operandconfig

qpdpQ:
add new entry for BTS in operandconfig


Daniel-Fan (approved):
/lgtm

PR #2198, opened 2024-09-16
[SC2] Add new entry for BTS with version in operandconfig

qpdpQ:
add new entry for bts in opconfig


Daniel-Fan (approved):
/lgtm

PR #2199, opened 2024-09-16
Use Unstructured obj to copy CR to other ns

Daniel-Fan:
issue: https://github.ibm.com/PrivateCloud-analytics/CPD-Quality/issues/28858

### Context
When there are old CS 3.x installed together with CS 4.x in the same cluster, it is usual that old CS 3.x installed later on the cluster, so entire OCP cluster is using the old CRD from CS 3.x. In this case, the field `.spec.services[*].spec` is a required field on old CRD from CS 3.x.

And we have documented that user has to explicitly set field `.spec.services[*].spec: {}` even it is empty
```
kind: CommonService
metadata:
  name: common-service
  namespace: operator-ns
spec:
  license:
    accept: true
  operatorNamespace: latest-validated
  services:
    - name: common-service-postgresql
      resources:
         ...
      spec: {}
```

### Issue
When CS 4.x propagates CS CR from operator namespace into other namespaces into the tenant, CS operator renders CS CR into structure `apiv3.CommonService`. However, this API structure will automatically **ignore** the field when it is empty.
Therefore, field `.spec.services[*].spec: {}` is pruned after the CR is copied into other namespace. And it is causing CRD compatibility issues
```
kind: CommonService
metadata:
  name: common-service
  namespace: operator-ns
spec:
  license:
    accept: true
  operatorNamespace: latest-validated
  services:
    - name: common-service-postgresql
      resources:
         ...
      # spec: {}       <-------------- this line will be removed when converting template into structure `apiv3.CommonService`
```

### Solution
In order to always preserve the field `.spec.services[*].spec: {}` even it is empty, we would use Unstructured obj to store the CS CR object, which will preserve every field in the template.

### Test
After applying the dev image `docker-na-public.artifactory.swg-devops.com/hyc-cloud-private-scratch-docker-local/ibmcom/common-service-operator-amd64:dev`, I could see that the CS CR in tenant namespace contains all empty fields and they are NOT pruned.
Screenshot 2024-09-16 at 2 02 50 PM

### DIFF
--- a/controllers/bootstrap/init.go
+++ b/controllers/bootstrap/init.go
@@ -1581,43 +1581,47 @@ func (b *Bootstrap) ConfigCertManagerOperandManagedByOperator(ctx context.Contex
 func (b *Bootstrap) PropagateDefaultCR(instance *apiv3.CommonService) error {
 	// Copy Master CR into namespace in WATCH_NAMESPACE list
 	watchNamespaceList := strings.Split(b.CSData.WatchNamespaces, ",")
-	csLabel := make(map[string]string)
-	// Copy from the original labels to the target labels
-	for k, v := range instance.Labels {
-		csLabel[k] = v
-	}
-	csAnnotation := make(map[string]string)
-	// Copy from the original Annotations to the target Annotations
-	for k, v := range instance.Annotations {
-		csAnnotation[k] = v
-	}
-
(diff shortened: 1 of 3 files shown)


bitscuit (approved):
/lgtm

PR #2202, opened 2024-09-16
updated BYO EDB operator handling

bitscuit:
- enables configuration via ibm-cpp-config
- with EDB_USER_MANAGED_OPERATOR_ENABLED: "true"
- which updates default CS CR with operatorConfig for internal-use-only-edb and sets userManaged
- internal-use-only-edb is new entry that references package for EDB, but has no-op so ODLM does not install it even if user creates OperandRequest for it

How to test

1. deploy common-services from CD catalog
2. update OperandRegistry with entry for `internal-use-only-edb`
3. ` make build-dev-image` and update cs-operator image on your cluster
4. create following configmap and restart cs-operator
5. verify that OperandRegistry is updated with `userManaged: true` for EDB entries
6. remove ibm-cpp-config and restart cs-operator
7. verify that OperandRegistry is updated with `userManaged` field removed

```yaml
  - channel: stable
    name: internal-use-only-edb
    namespace: "{{ .CPFSNs }}"
    packageName: cloud-native-postgresql
    scope: public
    installPlanApproval: Automatic
    installMode: no-op
```

```yaml
kind: ConfigMap
apiVersion: v1
metadata:
  name: ibm-cpp-config
  labels:
    operator.ibm.com/managedByCsOperator: 'true'
data:
  EDB_USER_MANAGED_OPERATOR_ENABLED: "true"
```

### DIFF
--- a/controllers/bootstrap/init.go
+++ b/controllers/bootstrap/init.go
@@ -1885,3 +1885,64 @@ func (b *Bootstrap) UpdateResourceWithLabel(resources *unstructured.Unstructured
 	}
 	return nil
 }
+
+func (b *Bootstrap) UpdateEDBUserManaged() error {
+	operatorNamespace, err := util.GetOperatorNamespace()
+	if err != nil {
+		return err
+	}
+	defaultCsCR := &apiv3.CommonService{}
+	csName := "common-service"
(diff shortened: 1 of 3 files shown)


Daniel-Fan (approved):
/lgtm

PR #2206, opened 2024-09-18
using`ibm-licensing-service-instance` deployment to ensure the exact namespace

qpdpQ:
Handle the edge case which v3 and v4 licensing operator co-existence in the same cluster.

- If multiple Licensing operators is found, we use `ibm-licensing-service-instance` deployment to ensure the exact namespace where licensing v4 is installed.
- If multiple licensing-service-instance found, error out

Fixes # https://github.ibm.com/IBMPrivateCloud/roadmap/issues/64782


Daniel-Fan (approved):
/lgtm

PR #2209, opened 2024-09-19
reduce the number of certs and secrets labeled to just cs-ca and zen-…

bluzarraga:
…ca plus already specified

Instead of labeling every secret and certificate watched by cert manager, we limit it to the "primary" resources that all other secrets are generated based on. These are the cs-ca-certificate and its accompanying secret and the zen-ca-cert-secret. In addition, there are other secrets previously specified in the script that are still labeled like the IM admin credentials since we do not want those to be regenerated with a different value.

Previously, it did not matter if we labeled them all but now that we need to support restoring into a different namespace, it is very difficult to create the logic that would individually update namespace related values (like dns lists) in secrets like the common-service-db-tls-cert. The secret values should be the same as on the original cluster because they are inherited from the cs-ca-certificate-secret so as long as that secret is carried over, the regenerated secret should contain the same values but with updated namespace values.

Fixes #

1. How the test is done?
Test BR as normal but using this script to label certs and secrets
If resources fail to come up due to certificate issues after restore then we are missing certs/secrets.

I have tested this previously but it was a few weeks ago with a simple BR setup (one without zen or im) so it's possible this is not sufficient when including zen and im

### DIFF
--- a/velero/backup/cert-manager/label-cert-manager.sh
+++ b/velero/backup/cert-manager/label-cert-manager.sh
@@ -35,8 +35,8 @@ do
     oc label issuer $NAME -n $NAMESPACE foundationservices.cloudpak.ibm.com=cert-manager --overwrite=true
 done
 
-# Get all certificates in all namespaces and add foundationservices.cloudpak.ibm.com=cert-manager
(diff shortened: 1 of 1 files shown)


Daniel-Fan (approved):
/lgtm

PR #2237, opened 2024-10-03
label length limited to 64 characters

qpdpQ:
 k8s label name length limit to 64 characters, so olm will cut off any label longer than 64 characters
 We rely on label to find csv and subscription, so we also need to cut of the label if it is longer than 64 characters.
script enhancement for # https://github.ibm.com/IBMPrivateCloud/roadmap/issues/64378
(will open pr for odlm and nss operator)


YCShen1010:
Test passed, when given a string with more than 64 characters, it will be truncated to 63 characters.

YCShen1010 (approved):
/lgtm

PR #2242, opened 2024-10-07
waiting for licensing instance deployment

qpdpQ:
waiting for licensing instance deployment before migrating licensing


YCShen1010 (approved):
/lgtm

PR #2252, opened 2024-10-15
conditionally wait for ibm-licensing-instance

qpdpQ:
We should only wait for ibm-licensing-instance when ibm-licensing operator is found, it could help to reducing the waiting time.

### DIFF
--- a/cp3pt0-deployment/setup_singleton.sh
+++ b/cp3pt0-deployment/setup_singleton.sh
@@ -231,7 +231,15 @@ function is_migrate_licensing() {
     fi
 
     title "Check migrating LTSR ibm-licensing-operator"
-    wait_for_licensing_instance_deployment
+    # wait for ibm-licensing-operator instance 
+    local licensing_operator_exist=$("$OC" get clusterserviceversion.operators.coreos.com -A | (grep ibm-licensing-operator || echo "fail"))
+    if [[ $licensing_operator_exist != "fail" ]]; then
+        wait_for_licensing_instance_deployment
+    else
+        info "No ibm-licensing-operator found, skipping"
+        return 0
+    fi
+
     local ns=$("$OC" get deployments -A | grep ibm-licensing-service-instance | cut -d ' ' -f1)
     if [ -z "$ns" ]; then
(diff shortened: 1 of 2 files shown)


cp3pt0-deployment/setup_singleton.sh
```diff
@@ -231,7 +231,15 @@ function is_migrate_licensing() {
     fi
 
     title "Check migrating LTSR ibm-licensing-operator"
-    wait_for_licensing_instance_deployment
+    # wait for ibm-licensing-operator instance 
+    local licensing_operator_exist=$("$OC" get clusterserviceversion.operators.coreos.com -A | (grep ibm-licensing-operator || echo "fail"))
```
YCShen1010, line 235: I am thinking if checking the deployment here is better than CSV? In case we hit the known OLM issue with unbounded CSV and sub?

```
local licensing_operator_exist=$("$OC" get deployment -A | (grep ibm-licensing-operator || echo "fail"))
```

cp3pt0-deployment/setup_singleton.sh
```diff
@@ -231,7 +231,15 @@ function is_migrate_licensing() {
...
+    # wait for ibm-licensing-operator instance 
+    local licensing_operator_exist=$("$OC" get deployment -A | (grep ibm-licensing-operator || echo "fail"))  
+    if [[ $licensing_operator_exist != "fail" ]]; then
+        wait_for_licensing_instance_deployment
+    else
+        info "No ibm-licensing-operator found, skipping"
+        return 0
+    fi
+
     local ns=$("$OC" get deployments -A | grep ibm-licensing-service-instance | cut -d ' ' -f1)
```
YCShen1010, line 243: Sorry, one more comment, I noticed this line is also included in `wait_for_licensing_instance_deployment`, is it possible move this check under the above `if` condition? Thanks

Daniel-Fan:
@qpdpQ 
Can you provide some test result for the above changes? Thank you!

Daniel-Fan:
@qpdpQ
Can you provide some test results for the above changes? Thank you!

Daniel-Fan (approved):
/lgtm

PR #2257, opened 2024-10-16
Use 2 replicas for Keycloak EDB on starterset (#2256)

Daniel-Fan:
Using one replica prevents cluster upgrade due to the pod disruption budget, so set the default to use two.
cherry pick of #2256


YCShen1010 (approved):
/lgtm

PR #2273, opened 2024-10-27
exclude internal generated operandrequest from Backup/Restore

qpdpQ:
stop labelling operandrequest which is generated by odlm or other services, those operandrequests should be regenerated in restore. 
We should only label primary OperandRequests
Fixes # https://github.ibm.com/PrivateCloud-analytics/CPD-Quality/issues/31736


YCShen1010 (approved):
/lgtm

PR #2287, opened 2024-11-04
updated turbo profile controller to handle data.resource

bitscuit:
Currently limited to setting `.spec.services[].resources.data.resources.limits.cpu` to `{}` value instead of entire `limits` field because some internal functionality is removing the `{}` value for entire `limits` field.

The code as-is will work regardless because when one of the k8s resources fields gets set to `{}`, then ODLM will ignore sizing.

How to test

1. build image from this PR
2. install cs-operator on your cluster
3. update image on cluster with image from step 1
4. add the following to CommonService CR `profileController: turbo`
5. verify that OperandConfig has `cpu: {}` for the EDB and Keycloak CRs


Daniel-Fan (approved):
/lgtm

PR #2300, opened 2024-11-13
removed existing operator ORMs since cannot hardcode CSV name

bitscuit:
because we do not officially support them yet, so avoid giving any chance of customers accidentally installing these


Daniel-Fan (approved):
/lgtm

PR #2301, opened 2024-11-13
script to apply ORM files

bitscuit:
adapted from network policy script


turbonomic/install_orm.sh
```diff
@@ -0,0 +1,177 @@
...
+    if [[ -z "${CS_NAMESPACE}" ]]; then
+        CS_NAMESPACE=$(oc project -q)
+    fi
+
+    # checking for ibm-common-service-operator in CS_NAMESPACE
+    if [[ -z "$(oc -n ${CS_NAMESPACE} get csv --ignore-not-found | grep 'ibm-common-service-operator')" ]]; then
+        info "IBM Cloud Pak foundational services are not installed in namespace ${CS_NAMESPACE}"
+    else
+        success "IBM Cloud Pak foundational services found in namespace ${CS_NAMESPACE}"
+    fi
```
Daniel-Fan, line 148: Do we need to support separation of duty that `CS_NAMESPACE` will represent operand namespace?

Or we could support separation of duty in next release when we have ORM resources for both operators and operands?

turbonomic/operands/cert-manager.yaml
```diff
@@ -1,8 +1,10 @@
 apiVersion: devops.turbonomic.io/v1alpha1
 kind: OperatorResourceMapping
 metadata:
-  name: cert-manager-orm
-  namespace: $namespace
+  name: cert-manager-config-orm
+  namespace: {{ placeholder_namespace }}
+  labels:
+    component: cpfs
```
Daniel-Fan, line 7: Although cert manager CR is cluster scope resource, I am wondering if this is the best practice to apply cert manager ORM in CPFS namespace? It should still work because cert manager CR is visible in CPFS namespace. Would it be better to have cert manager ORM deployed in cert manager namespace?

bitscuit: Good point, will adjust so that it accepts cert-manager namespace and use ibm-cert-manager as default

turbonomic/install_orm.sh
bitscuit: I was thinking these should always be created in services namespace. Completely forgot that this check negates that. Will update.

Daniel-Fan (approved):
/lgtm

PR #2327, opened 2024-12-05
suspend go routines until operator cache is ready

qpdpQ:
Go routines in ibm-common-service operator relies on operator [controller's cache](https://github.com/IBM/ibm-common-service-operator/blob/b947cd93e81eccefd769c64628de3c730fa1797a/main.go#L124) to be ready. It will be ready when [Manager is set up](https://github.com/IBM/ibm-common-service-operator/blob/b947cd93e81eccefd769c64628de3c730fa1797a/main.go#L253). So we need to suspend go routines until manager is setup
Fixes # https://github.ibm.com/IBMPrivateCloud/roadmap/issues/58246

### DIFF
--- a/main.go
+++ b/main.go
@@ -164,6 +164,7 @@ func main() {
 		// Delete Keycloak Cert
 		go goroutines.CleanupResources(ch)
 
+		klog.Infof("Setup commonservice manager")
 		if err = (&controllers.CommonServiceReconciler{
 			Bootstrap: bs,
 			Scheme:    mgr.GetScheme(),
@@ -174,6 +175,7 @@ func main() {
 		}
 
 		// start go routines
+		klog.Infof("Start go routines")
 		ch <- bs
 
 		// check if cert-manager CRD does not exist, then skip cert-manager related controllers initialization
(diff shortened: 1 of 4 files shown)


main.go
```diff
@@ -178,6 +181,8 @@ func main() {
 		}
 		if !exist && err == nil {
 			klog.Infof("cert-manager CRD does not exist, skip cert-manager related controllers initialization")
+			// start go routines
+			ch <- bs
```
Daniel-Fan, line 185: Hi @qpdpQ 
I am wondering that are we able to send the single right after we setup manager for `CommonServiceReconciler{}` at line 175?
```
                if err = (&controllers.CommonServiceReconciler{
			Bootstrap: bs,
			Scheme:    mgr.GetScheme(),
			Recorder:  mgr.GetEventRecorderFor("commonservice-controller"),
		}).SetupWithManager(mgr); err != nil {
			klog.Errorf("Unable to create controller CommonService: %v", err)
			os.Exit(1)
		}
                
                ch <- bs
```

Daniel-Fan (approved):
/lgtm

PR #2338, opened 2025-01-13
Enhance Keycloak 26 truststore file and sa certs import

YCShen1010:
**What this PR does**: In Keycloak 24 and 26, it supports the truststore field directly defined in the spec field, allowing to specify the secret in it. Additionally, when running on a Kubernetes or OpenShift environment, well-known locations of trusted certificates, such as `/var/run/secrets/kubernetes.io/serviceaccount/ca.crt`, are included automatically. 
Therefore, we can remove the import of kube service account certificates from the startup script.

Doc reference: https://docs.redhat.com/en/documentation/red_hat_build_of_keycloak/26.0/html/operator_guide/advanced-configuration-#advanced-configuration-truststores

**Issue**: 
https://github.ibm.com/IBMPrivateCloud/roadmap/issues/65326#issuecomment-101278352
https://github.ibm.com/IBMPrivateCloud/roadmap/issues/65678

**How to test**:
Test Image: quay.io/yuchen_shen/cs_operator:truststore
1. delete the OperandConfig and apply the test image
2. install keycloak, the console could be accessed successfully.

### DIFF
--- a/bundle/manifests/ibm-common-service-operator.clusterserviceversion.yaml
+++ b/bundle/manifests/ibm-common-service-operator.clusterserviceversion.yaml
@@ -23,7 +23,7 @@ metadata:
     capabilities: Seamless Upgrades
     cloudPakThemesVersion: styles4100.css
     containerImage: icr.io/cpopen/common-service-operator:4.12.0
-    createdAt: "2025-03-18T20:20:54Z"
+    createdAt: "2025-03-18T21:27:31Z"
     description: The IBM Cloud Pak foundational services operator is used to deploy IBM foundational services.
     features.operators.openshift.io/disconnected: "true"
     features.operators.openshift.io/fips-compliant: "true"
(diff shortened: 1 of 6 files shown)


internal/controller/constant/odlm.go
```diff
@@ -1041,6 +1029,94 @@ spec:
         force: true
         kind: ConfigMap
         name: cs-keycloak-user-profile
+      - apiVersion: v1
+        kind: ServiceAccount
+        name: convert-secret-sa
+      - apiVersion: rbac.authorization.k8s.io/v1
+        kind: Role
+        name: convert-secret-role
```
Daniel-Fan, line 1037: Can we use a name which is more specific to keycloak resources?

bundle/manifests/ibm-common-service-operator.clusterserviceversion.yaml
```diff
@@ -383,6 +383,8 @@ spec:
                             fieldPath: metadata.annotations['olm.targetNamespaces']
                       - name: OPERATOR_NAME
                         value: ibm-common-service-operator
+                      - name: UTILS_IMAGE
```
Daniel-Fan, line 386: Can we use naming convention `CPFS_UTILS_IMAGE`?

Daniel-Fan (approved):
/lgtm

PR #2345, opened 2025-01-20
Enable granting permission for token exchange

YCShen1010:
If the clients that want to exchange tokens for a different client need to be authorized in the Admin Console, so a token-exchange fine grain permission `- admin-fine-grained-authz` was enabled by default in Keycloak CR

Fixes # https://github.ibm.com/IBMPrivateCloud/roadmap/issues/65708

1. After `- admin-fine-grained-authz` is enabled, the permission section in the client is shown up

_**Before adding into CR:**_

Screenshot 2025-01-20 at 3 56 17 PM

_**After**:_

Screenshot 2025-01-20 at 4 05 47 PM

**Doc reference:** 
https://www.keycloak.org/securing-apps/token-exchange#_internal-token-to-internal-token-exchange

### DIFF
--- a/controllers/constant/odlm.go
+++ b/controllers/constant/odlm.go
@@ -947,6 +947,7 @@ spec:
             features:
               enabled:
                 - token-exchange
+                - admin-fine-grained-authz
             db:
               host: keycloak-edb-cluster-rw
               passwordSecret:


Daniel-Fan (approved):
/lgtm

PR #2352, opened 2025-01-27
Mongo certficate DNS name patch

qpdpQ:
enhancement for DNS name patch, previous PR only update the DNSnames in certificate `icp-mongodb-client-cert` is not enough, this certificaet isn't used directly by the MongoDB runtime. Mongodb init container will use a leaf certificate and that leaf certificate is signed by `icp-mongodb-client-cert`, we need to update that leaf certificate instead.

While preload_data.sh is executed, the script would:
1. Scale down ibm-mongodb-operator
2. Update the script in icp-mongodb-init CM's .data["on-start.sh"] [field](https://github.com/IBM/ibm-common-service-operator/blob/scripts-dev/preload_data.sh#L2037) and add the additional DNS names
3. Trigger a rolling upgrade on the icp-mongodb StatefulSet, adding `migrating: "true"` to `.spec.template.metadata.labels` in mongodb StatefulSet
4. Scale up ibm-mongodb-operator

Fixes #  https://github.ibm.com/IBMPrivateCloud/roadmap/issues/65642

### DIFF
--- a/preload_data.sh
+++ b/preload_data.sh
@@ -331,6 +331,7 @@ function patch_cm() {
       # scale down icp-mongodb statefulset
       ${OC} scale statefulSet -n ${FROM_NAMESPACE} icp-mongodb --replicas=0
     fi
+    
     # update icp-mongodb-init configmap
     cat << EOF | ${OC} apply -n $FROM_NAMESPACE -f -
 kind: ConfigMap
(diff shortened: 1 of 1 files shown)


bitscuit (approved):
/lgtm

PR #2358, opened 2025-01-31
remove crs from helm chart

qpdpQ:
remove certificate crs and EDB image list from helm chart, those can be created automatically now


bitscuit (approved):
/lgtm

PR #2363, opened 2025-02-03
added CommonService CR to helm chart

bitscuit:
fixed incorrect operator image name


bitscuit:
it's timing out, maybe too much load or something wrong with the build machine

PR #2364, opened 2025-02-04
refactor cs-operator to enable common-service controller

qpdpQ:
Refactor cs-operator to de-couple it from OLM API
Enable common-service controller, so most of the key feature in cs-operators working in no olm environment, except:
1. webhooks (need to create webhook service and inject webhook-secret)
2. odlm and nss cleanup if it is upgrade from LTSR (need to cleanup sub and csv)
3. ~~operator status check (need to check deployment instead of sub and csv)~~

Fixes #  https://github.ibm.com/IBMPrivateCloud/roadmap/issues/65720

### DIFF
--- a/controllers/constant/constant.go
+++ b/controllers/constant/constant.go
@@ -114,6 +114,11 @@ const (
 	CSMappingConfig = "ibm-cs-ns-mapping-webhook-configuration"
 )
 
+// Bedrock Deployment names for No OLM
+var (
+	DeploymentsName = []string{"ibm-common-service-operator", "operand-deployment-lifecycle-manager", "ibm-namespace-scope-operator", "ibm-commonui-operator", "ibm-iam-operator"}
+)
+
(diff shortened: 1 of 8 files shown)


controllers/constant/constant.go
```diff
@@ -114,6 +114,11 @@ const (
 	CSMappingConfig = "ibm-cs-ns-mapping-webhook-configuration"
 )
 
+// Bedrock Deployment names for No OLM
+var (
+	DeploymentsName = []string{"ibm-common-service-operator", "operand-deployment-lifecycle-manager", "ibm-namespace-scope-operator", "ibm-commonui-operator", "ibm-iam-operator"}
```
bitscuit, line 119: I don't see why we need to put status info on the operator deployments. OLM cs-operator did not give status on operators, so why do we need to do this for non-OLM mode?

bitscuit (approved):
/lgtm

PR #2370, opened 2025-02-07
[WIP] OLM -> No OLM migration job helm chart

bluzarraga:
Create migration job for upgrade from OLM install to no OLM install.

Job is based on CPD migration template found here https://github.ibm.com/Mengdie-Chu/non-olm-installer/tree/main/charts/cpd-platform-migration and essentially runs a script where the sub, csv, and rbac for installer and bedrock services are cleaned up so they can be replaced by a helm install. CRs like NSS and CS are included to "adopt" the existing ones without outright replacing them


no-olm/migration/migration.sh
```diff
@@ -0,0 +1,80 @@
+#!/usr/bin/env bash
```
bitscuit, line 1: Was this standalone script meant to be checked in? I see a comment that it is a PoC

bluzarraga: It is poc but its easier to test changes to the script in script form. I'll remove it before merging in

bluzarraga:
closing in favor of https://github.com/IBM/ibm-common-service-operator/pull/2379

PR #2379, opened 2025-02-12
create olm -> no olm migration chart in master branch

bluzarraga:

Create migration job chart for migration from OLM install to no OLM install for bedrock components.

Job is based on CPD migration template found here https://github.ibm.com/Mengdie-Chu/non-olm-installer/tree/main/charts/cpd-platform-migration and essentially runs a script where the sub, csv, and rbac for installer and bedrock services are cleaned up so they can be replaced by a helm install. CRs like NSS and CS are included to "adopt" the existing ones without outright replacing them.

Cleans up services and tethered namespaces as well.

Fixes # https://github.ibm.com/IBMPrivateCloud/roadmap/issues/65830

1. How the test is done?

- Install CS, ODLM, NSS, IM, EDB, and UI via OLM (so as normal). Make sure to include cert manager in initial install.
- Once CPFS is ready, edit migration chart values.yaml to include correct values
- install the migration chart, include --take-ownership flag in helm install command
    - If using toplevel values file instead, make sure to still set the imEnabled value to true
- verify that subscriptions, csvs, roles, rolebindings, and service accounts are deleted from the install. Should be very few pods if any running in operator namespace. 
- Verify services namespace still has required operands running as expected
- install cluster and namespace scope helm charts, may need --take-ownership flag if using helm install command
- verify operators come ready as expected, verify can login to cp-console still

Alternate test could be to install cluster scoped charts before running migration to verify if there is a difference in the final install if the order changes

### DIFF
--- a/bedrock-migration/values.yaml
+++ b/bedrock-migration/values.yaml
@@ -4,6 +4,6 @@ imEnabled: "true" #should this be a global value? Is there a better way to deter
 global:
   operatorNamespace: "cs-op"
   instanceNamespace: "cs-serv"
-  tetheredNamespaces: ["test"]
+  tetheredNamespaces: [""]
   imagePullPrefix: icr.io
   imagePullSecret: ibm-entitlement-key
--- a/bedrock-migration/values.yaml
+++ b/bedrock-migration/values.yaml
@@ -4,6 +4,6 @@ imEnabled: "true" #should this be a global value? Is there a better way to deter
 global:
   operatorNamespace: "cs-op"
   instanceNamespace: "cs-serv"
-  tetheredNamespaces: [""]
+  tetheredNamespace: [""]
   imagePullPrefix: icr.io
   imagePullSecret: ibm-entitlement-key
--- a/bedrock-migration/values.yaml
+++ b/bedrock-migration/values.yaml
@@ -1,0 +1,9 @@
+imEnabled: "true" #should this be a global value? Is there a better way to determine this?
+
+#stand-in global values for testing
+global:
+  operatorNamespace: "cs-op"
+  instanceNamespace: "cs-serv"
+  tetheredNamespaces: ["test"]
+  imagePullPrefix: icr.io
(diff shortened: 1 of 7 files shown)


bluzarraga:
Something to keep an eye on with testing is whether the CS Cr is overwritten by what is included in the cs operator chart post migration.

So by that I mean, after running the migration, when installing the cs operator chart, if there were configurations in the CS CR not included in the cs operator chart, do they get overwritten

bluzarraga:
Big and confusing merge conflicts when switching the branch from mast to release-future so keeping it master for now

bedrock-migration/values.yaml
```diff
@@ -0,0 +1,9 @@
+imEnabled: "true" #should this be a global value? Is there a better way to determine this?
+
+#stand-in global values for testing
+global:
+  operatorNamespace: "cs-op"
+  instanceNamespace: "cs-serv"
+  tetheredNamespaces: ["test"]
```
bitscuit, line 7: I think this should be left as empty array 
```suggestion
  tetheredNamespaces: []
```

bedrock-migration/values.yaml
```diff
@@ -0,0 +1,9 @@
+imEnabled: "true" #should this be a global value? Is there a better way to determine this?
+
+#stand-in global values for testing
+global:
+  operatorNamespace: "cs-op"
+  instanceNamespace: "cs-serv"
+  tetheredNamespaces: [""]
```
YCShen1010, line 7: Hi Ben, CPD changed the tethered namespace to `tetheredNamespace` without `s`: https://github.ibm.com/Mengdie-Chu/non-olm-installer/blob/main/top-override.yaml#L7

```suggestion
  tetheredNamespace: [""]
```

Jeremy-Cheng-stack (approved):
Tested and works. For reference, need to have ibm-entitlement-key created before running the chart.

bluzarraga:
merging based on Jeremy's approval

PR #2381, opened 2025-02-12
Sync release-future branch with master

bluzarraga:
We are using the future branches to make changes to the helm charts so we can continue to make changes to helm charts while code freeze is in place for 4.11 GA


bitscuit (approved):
/lgtm

PR #2386, opened 2025-02-14
Keycloak 26 enable internal connection

rickysway:
As of Keycloak 25 onwards, with the move to Hostname V2, for internal cluster Pod to Service communication you need to enable `backchannelDynamic` (effectively adding env var `KC_HOSTNAME_BACKCHANNEL_DYNAMIC: 'true'` to Keycloak), we opt for setting the `env` as a customer could potentially installed Keycloak 24 in another namespace after this installation, which will break their Keycloak 26 installation if we relied on `backchannelDynamic` due to the change of CRD
https://www.keycloak.org/server/hostname#_utilizing_an_internal_url_for_communication_among_clients

According to Keycloak Hostname V2 docs when `spec.hostname.backchannelDynamic: true` the hostname needs to be a URL and not just hostname, so adding `https://` to the start of the hostname value
https://www.keycloak.org/server/hostname#_validations

Fixes https://github.ibm.com/IBMPrivateCloud/roadmap/issues/65922

Unsure if the below section "backport" is needed


internal/controller/constant/odlm.go
```diff
@@ -964,7 +964,7 @@ spec:
                     apiVersion: route.openshift.io/v1
                     kind: Route
                     name: keycloak
-                    path: .spec.host
+                    path: https://+.spec.host
```
rickysway, line 967: This does not work on Keycloak 22/24. Noting the url becomes `https://https//........`
This works on Keycloak 26 however

https://github.com/user-attachments/assets/22f69b75-16ab-46bc-87d1-9af92d15c946

YCShen1010:
Hello @rickysway, I think we could directly define `.spec.additionalOptions` in the Keycloak CR to enable `KC_HOSTNAME_BACKCHANNEL_DYNAMIC` equals to true instead, and then it will be present in the env in pod yaml.  like
```
data:
    spec:
    ...
    additionalOptions:                          <----- newly added
      - name: hostname-backchannel-dynamic
        value: "true"
    ...
    ...
    proxy:
      headers: xforwarded
    features:
      enabled:
        - token-exchange
        - admin-fine-grained-authz
```

When it is enabled as `true`, it requires `https://` as a prefix for the hostname. We are going to enhance the ODLM templating logic to handle this, and I could open a new PR with this update later. 
Thank you for pointing out these changes to us.

rickysway:
Closing as no longer needed

PR #2391, opened 2025-02-17
Polishmigration

bluzarraga:
**What this PR does / why we need it**: More robust error handling and logging. It is also idempotent now. As long as the namespace scope configmap is present, the job can be run multiple times without issue

Fixes #

1. How the test is done?

### DIFF
--- a/bedrock-migration/templates/bedrock-migration-job.yaml
+++ b/bedrock-migration/templates/bedrock-migration-job.yaml
@@ -113,7 +113,7 @@ spec:
                 {{- end }}
 
                 #get nss roles?
-                echo "${roles}"
+                echo "Roles to delete: ${roles}"
 
                 if [[ ! -z $roles ]]; then
                   oc delete role $roles -n $ns --ignore-not-found
--- a/bedrock-migration/templates/bedrock-migration-job.yaml
+++ b/bedrock-migration/templates/bedrock-migration-job.yaml
@@ -12,9 +12,7 @@ spec:
     spec:
       containers:
       - name: bedrock-migration-job
-        #should we be using our own image or the cpd image? I vote our own
-        # image: '{{ .Values.global.imagePullPrefix }}/{{ .Values.global.migrationJobImgName }}@{{ .Values.global.migrationJobImgDigest }}'
-        image: icr.io/cpopen/cpfs/cpfs-utils:4.6.7
+        image: {{ .Values.global.imagePullPrefix }}/{{ .Values.cpfs.imageRegistryNamespaceOperator }}/cpfs/cpfs-utils:4.6.7
         command:
           - bash
           - -c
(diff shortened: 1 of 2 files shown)


bedrock-migration/templates/bedrock-migration-job.yaml
```diff
@@ -24,34 +22,65 @@ spec:
...
 
+            
             nssSub=$(oc get subscription.operators.coreos.com -n "$operatorNamespace" -o jsonpath="{.items[?(@.spec.name=='ibm-namespace-scope-operator')].metadata.name}")
-            nssCSV=$(oc get --ignore-not-found sub $nssSub -n $operatorNamespace -o jsonpath='{.status.currentCSV}')
-            nssVersion=$(oc get --ignore-not-found csv $nssCSV -n $operatorNamespace -o jsonpath='{.spec.version}')
-
+            if [[ -z $nssSub ]]; then
+              echo "NamespaceScope Subscription not present in operator namespace $operatorNamespace, skipping deleting Subscription and CSV."
+            else
+              nssCSV=$(oc get --ignore-not-found sub $nssSub -n $operatorNamespace -o jsonpath='{.status.currentCSV}')
```
pgodowski, line 30: please use fully qualified Subscription name - the `sub` and `subscription` is amgibuous

bluzarraga:
One thing I'm seeing with the migration is that the `--take-ownership` flag doesn't prevent the empty cs cr included in the migration from overwriting the existing cs cr on the cluster. I'm thinking about maybe making a copy of the existing cs cr during the migration process to make sure any existing configurations are not lost during the migration process. It seems like it would be overwritten no matter what with how it is now.

I am not as worried about nss cr becuase if nss is installed than the cr should be configured the same way via helm as it was via OLM.

bluzarraga:
I ran another test and I was mistaken, the empty cr does not replace the existing one. But my point about installing the cs operator chart and the included cs cr in there still stands as it would replace the existing cs cr. Also, if we helm uninstall the migration chart after it is done, it will cleanup the nss cr and the cs cr since they are "owned" by the migration chart

bluzarraga:
Separate from the comments above, one thing I can't seem to get around is this error message in the migration logs:
```
E0218 21:34:38.371814      66 reflector.go:150] k8s.io/client-go/tools/watch/informerwatcher.go:146: Failed to watch *unstructured.Unstructured: unknown
E0218 21:34:39.644088      66 reflector.go:150] k8s.io/client-go/tools/watch/informerwatcher.go:146: Failed to watch *unstructured.Unstructured: unknown
```
whenever I delete the csv. As far as I can tell, it doesn't really mean anything but its always outputted into the logs and could easily be misleading

bitscuit (approved):
/lgtm

PR #2393, opened 2025-02-20
Remove non-OLM env from OLM bundle

Daniel-Fan:
When we run `make generate-all` to update the bundle for RH certification, the non-OLM env variables are accidentally added into the `bundle` folder from `config` folder.

Makefile is updated, so that `make generate-all` will not add non-OLM env variables into bundle folder.
Those non-OLM env variables will only be introduced when it is rendering the helm chart via `make helm`.

Fixes https://github.ibm.com/IBMPrivateCloud/roadmap/issues/65966


bitscuit (approved):
/lgtm

PR #2396, opened 2025-02-24
no-olm controller

qpdpQ:
What this PR does / why we need it:
sepereate common-service controller into two controller one for olm and another for non-olm

everything should work as normal except webhook
this image can be used for test `quay.io/yuchen_li1/common-service-operator-amd64:dev`


bitscuit (approved):
tested on a cluster and worked, waiting for another review before merge

bitscuit (approved):
/lgtm

PR #2400, opened 2025-02-25
updated version to 4.12.0

bitscuit:


YCShen1010 (approved):
/lgtm

bitscuit:
I might force the merge because it's just at the linter timing out. Could be an issue with the build machine, in which case, safe to ignore the error, since no other tests are failing.

PR #2410, opened 2025-02-27
Remove truststore startup script for keycloak 24 and 26

YCShen1010:
### What this PR does / why we need it:
In Keycloak 24 and 26, it supports the truststore field directly defined in the spec field, allowing to specify the secret in it. Additionally, when running on a Kubernetes or OpenShift environment, well-known locations of trusted certificates, such as `/var/run/secrets/kubernetes.io/serviceaccount/ca.crt`, are included automatically.
Therefore, we can remove startup script which includes truststore setup and the import of kube service account certificates.

### Issue:
https://github.ibm.com/IBMPrivateCloud/roadmap/issues/65326#issuecomment-101278352
https://github.ibm.com/IBMPrivateCloud/roadmap/issues/65678

### How to test:
1. Test image: `quay.io/yuchen_shen/cs_operator:truststore`
2. Check the new created `common-service` OperandConfig with the new optionalFields, this mean, if it is new CRD with `.spec.truststore`, then we do not need to run the startup script
    ```
    - path: .spec.unsupported.podTemplate.spec.containers[0].command
            operation: remove
            matchExpressions:
              - objectRef:
                  name: keycloaks.k8s.keycloak.org
                  apiVersion: apiextensions.k8s.io/v1
                  kind: CustomResourceDefinition
                key: .spec.versions[0].schema.openAPIV3Schema.properties.spec.properties.truststores
                operator: Exists
      ```
3. Install 24+ `keycloak-operator` and check keycloak cr there is no following command in containers
    ```
    - command:
        - /bin/sh
        - /mnt/startup/cs-keycloak-entrypoint.sh
    ```
4. Check the first few lines in `cs-keycloak-0`, it should start with 
    ```
    Changes detected in configuration. Updating the server image.
    Updating the configuration and installing your custom providers, if any. Please wait.
    ```
    no more: 
    ```
    Building the truststore file ...
    Importing default service account certificates ...
    ```
    Also, in terminal, `cd mnt/truststore` there is no more `keycloak-truststore.jks` existing.
5. Last check, if the keycloak route could open and login.
6. Compatibility with KeyCloak 22, delete 24 CRD and install 22, check above step, the result should be changed as above mentioned.

**Doc reference**: https://docs.redhat.com/en/documentation/red_hat_build_of_keycloak/26.0/html/operator_guide/advanced-configuration-#advanced-configuration-truststores

### DIFF
--- a/api/v3/commonservice_types.go
+++ b/api/v3/commonservice_types.go
@@ -47,6 +47,7 @@ type CSData struct {
 	ExcludedCatalog         string
 	StatusMonitoredServices string
 	ServiceNames            map[string][]string
+	UtilsImage              string
 }
 
 // +kubebuilder:pruning:PreserveUnknownFields
--- a/bundle/manifests/ibm-common-service-operator.clusterserviceversion.yaml
+++ b/bundle/manifests/ibm-common-service-operator.clusterserviceversion.yaml
@@ -23,7 +23,7 @@ metadata:
     capabilities: Seamless Upgrades
     cloudPakThemesVersion: styles467.css
     containerImage: icr.io/cpopen/common-service-operator:latest
-    createdAt: "2025-03-06T19:45:25Z"
+    createdAt: "2025-03-05T20:28:45Z"
     description: The IBM Cloud Pak foundational services operator is used to deploy IBM foundational services.
     nss.operator.ibm.com/managed-operators: ibm-common-service-operator
     nss.operator.ibm.com/managed-webhooks: ""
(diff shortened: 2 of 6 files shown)


YCShen1010:
1. Add `spec.truststores` stanza with the trusted certs using CRD matchexpression approach
    ```
    truststores:
              my-truststore:
                secret:
                  name: cs-keycloak-ca-certs
    ```
2. How to check, inside the container, check whether the custom certificates are present under the path `/opt/keycloak/conf/truststores/secret-cs-keycloak-ca-certs`.  Try Adding and deleting certificates will take effect immediately.

YCShen1010:
New enhancement added:
1. In keycloak 24+, the kube certs are automatically included in `/var/run/secrets/kubernetes.io/serviceaccount/ca.crt,/var/run/secrets/kubernetes.io/serviceaccount/service-ca.crt'` when truststore is defined
2. Create a job `convert-secret-job` to convert custom certificates from the `cs-keycloak-ca-certs` ConfigMap to a Secret, which can be defined directly in `spec.truststore`. The conversion script is included in the `convert-secrets CM, which is mounted in the job.
3. Introduce the `cpfs-utils` image into the CS CSV env values and use the csData structure to retrieve and render it into the job.

Daniel-Fan (approved):
/lgtm

PR #2425, opened 2025-03-07
Add Postgres services in non-olm bootstrap

Daniel-Fan:
It is used to fix the issue that `cloud-native-postgresql`, `cloud-native-postgresql-v1.22` and `cloud-native-postgresql-v1.25` entry missing in OperandConfig under non-olm deployment

### Test
After patching the dev image ` quay.io/daniel_fan/common-service-operator-amd64:dev` on the cluster with non-olm deployment, the OperandConfig contains those PostgreSQL entries
Screenshot 2025-03-07 at 2 48 40 PM


bitscuit (approved):
/lgtm

PR #2426, opened 2025-03-07
handle error returning

Daniel-Fan:
There is a corner case that Cs operator is hitting [an error from line 289](https://github.com/IBM/ibm-common-service-operator/blob/1a36d1997101cdbc5414933994cf5c510fa95090/internal/controller/commonservice_controller.go#L289).

However, there is a discrepancy between [what err is returned from `err := r.Bootstrap.UpdateEDBUserManaged()`](https://github.com/IBM/ibm-common-service-operator/blob/1a36d1997101cdbc5414933994cf5c510fa95090/internal/controller/commonservice_controller.go#L285C5-L285C46) and [what error is returned for reconciliation](https://github.com/IBM/ibm-common-service-operator/blob/1a36d1997101cdbc5414933994cf5c510fa95090/internal/controller/commonservice_controller.go#L290). Thus, the error was not actually returned to trigger the re-queue on reconciliation.

Fixes https://github.ibm.com/IBMPrivateCloud/roadmap/issues/66102

### Test
The change ensures that CS operator would always re-queue the reconciliation request until its CommonService CR reaching `Succeeded` phase.


YCShen1010 (approved):
/lgtm

PR #2432, opened 2025-03-13
added cpfs specific imagePullPrefix, use global if empty

bitscuit:
To handle different dev images being in different dev registries. Has no impact for GA


YCShen1010 (approved):
/lgtm

PR #2435, opened 2025-03-13
added new labels variable to values

bitscuit:
to allow for generic configuration of helm templates' `.metadata.labels`


Daniel-Fan (approved):
/lgtm

PR #2446, opened 2025-03-18
Enhancement on EnsureLabels function

qpdpQ:
The issue happened on both olm and no-olm environment
- cs-operator is looking for `common-services-maps`and `namespace-scope` in [serviceNamespace](https://github.com/IBM/ibm-common-service-operator/blob/master/internal/controller/bootstrap/init.go#L1966),but they are not in service namespace in 4.x version, so it will return an empty value
- we [forgot to check](https://github.com/IBM/ibm-common-service-operator/blob/master/internal/controller/bootstrap/init.go#L2054) if input values is nil or not before we update the label
- cs-operator don't have update permission on certificates and issuers, so it can't update the label in those resources

Fixes # https://github.ibm.com/IBMPrivateCloud/roadmap/issues/66196

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -74,7 +74,7 @@ else
 ARTIFACTORYA_REGISTRY ?= "docker-na-public.artifactory.swg-devops.com/hyc-cloud-private-scratch-docker-local/ibmcom"
 endif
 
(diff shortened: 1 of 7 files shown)


Daniel-Fan (approved):
/lgtm

PR #2448, opened 2025-03-19
Remove image list cm and edb config secret during migration

bluzarraga:
**What this PR does / why we need it**: Resolves issue with certain resources being created by the edb chart. These resources are ok to be regenerated and do not need to be carried over so we can remove them in the migration process and allow the EDB chart to recreate them as normal.

Fixes #

1. How the test is done?


bitscuit (approved):
/lgtm

PR #2454, opened 2025-03-24
retain the nss managed from roles to prevent lack of permissions

bluzarraga:
NSS roles occupy a fickle spot, NSS operator specifically looks for the `nss-managed-role-from-<ns>` role to be able to project permissions but it cannot create this role itself and the helm chart also does not create this role. This role is only created through the authorize-namespace.sh script run at fresh install time. If we delete this role during the migration then we would need to rerun the authorize namespace script in order to re-create these permissions. It was a mistake to remove these permissions in the first place but I forgot these roles were created via the authorize-namespace.sh script and not by the operator itself.

So instead of rerunning authorize namespace, we will opt not to delete the roles during migration. NSS operator will function as expected with this change.


bitscuit (approved):
/lgtm

PR #2487, opened 2025-04-07
remove last , in commonservice label template

qpdpQ:
if Values.cpfs.labels is empty in helm chart, the last ',' in labels might cause issue in creating commonservice cr, so we need to move it into loop


Daniel-Fan (approved):
/lgtm

PR #2491, opened 2025-04-14
[CD] add handling for no-op mode in operator status checks

Daniel-Fan:
When CS operator detects a no-op operator requested by users from OperandRegistry status, it should skip checking no-op operator status given no-op operator will not deployed by ODLM

Fixes https://github.ibm.com/IBMPrivateCloud/roadmap/issues/66456


YCShen1010 (approved):
/lgtm

PR #2501, opened 2025-04-29
Update backchannel-dynamic as the first-class citizen of the Keycloak CR

YCShen1010:
Based on the CRD schema, the `backchannel-dynamic` setting should be configured as a boolean property under the `hostname` section, not as an option in additionalOptions

**Which issue(s) this PR fixes** 
Fixes # https://github.ibm.com/IBMPrivateCloud/roadmap/issues/66572

**How to test**:
1. Test images
    CS: quay.io/yuchen_shen/cs_operator:kc_warning
    ODLM: quay.io/yuchen_shen/odlm:kc_warning
2. Install CS 4.6.14 and replaced by above images
3. Request a Keycloak 26 and check the `backchannelDynamic` is under `hostname` field and no warning msg

### DIFF
--- a/internal/controller/constant/odlm.go
+++ b/internal/controller/constant/odlm.go
@@ -1114,31 +1114,59 @@ spec:
                 templatingValueFrom:
                   conditional:
                     expression:
-                      greaterThan:
-                        left:
-                          objectRef:
-                            apiVersion: apps/v1
-                            kind: Deployment
-                            name: rhbk-operator
-                            path: .metadata.labels.olm\.owner
-                        right:
-                          literal: rhbk-operator.v26.0.0
+                      and:
+                        - greaterThan:
+                            left:
+                              objectRef:
+                                apiVersion: apps/v1
+                                kind: Deployment
+                                name: rhbk-operator
+                                path: .metadata.labels.olm\.owner
+                            right:
+                              literal: rhbk-operator.v26.0.0
+                        - equal:
+                            left:
+                              objectRef:
+                                apiVersion: apiextensions.k8s.io/v1
+                                kind: CustomResourceDefinition
+                                name: keycloaks.k8s.keycloak.org
+                                path: .spec.versions[0].schema.openAPIV3Schema.properties.spec.properties.hostname.properties.backchannelDynamic.type
+                            right:
+                              literal: "boolean"
                     then:
                       boolean: true
             additionalOptions:
               templatingValueFrom:
                 conditional:
                   expression:
-                    greaterThan:
-                      left:
-                        objectRef:
-                          apiVersion: apps/v1
-                          kind: Deployment
-                          name: rhbk-operator
-                          path: .metadata.labels.olm\.owner
-                      right:
-                        literal: rhbk-operator.v24.0.0
+                    and:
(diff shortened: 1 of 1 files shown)


internal/controller/constant/odlm.go
```diff
@@ -1110,32 +1110,39 @@ spec:
                         name: keycloak
                         path: .spec.host
                       required: true
+              backchannelDynamic:
+                templatingValueFrom:
+                  conditional:
+                    expression:
+                      greaterThan:
```
bitscuit, line 1117: Is there an operator for `>=`? I ask because I see that this conditional is using `greaterThan`, so if Keycloak is greater than v26 https://github.com/IBM/ibm-common-service-operator/pull/2501/files#diff-59cd2fe3c427d508f05b096424b09148a5bfaa84b29e54918147e49935ac1c3aR1125.

But in the previous code, the conditional was `lessThan` v26 https://github.com/IBM/ibm-common-service-operator/pull/2501/files#diff-59cd2fe3c427d508f05b096424b09148a5bfaa84b29e54918147e49935ac1c3aR1125.

What is the expected behaviour if Keycloak is v26?

YCShen1010: Hi @bitscuit 

1. We do not support `>=` or `<=` in conditional statements at the moment.
2. The changes in this PR only affects the `hostname.backchannel` field, which must be set to true starting from Keycloak 26. 
3. For the following setting, it actually needs to be applied for Keycloak 24 or later (setting it for version 22 also works).
```
name: spi-user-profile-declarative-user-profile-config-file
value: /mnt/user-profile/cs-keycloak-user-profile.json       
```
4. The use of `lessThan` in the previous code was due to a limitation where we couldn’t define different conditions within a single array structure. For example, we couldn’t express logic like: “if KC > 26, set backchannel; if KC > 24, set the user profile config...” inside one `additionalOptions` array.
5. With the recent changes, I’ve moved the `backchannel` configuration to `.spec.hostname`, which allows to directly use `greaterThan` logic for setting the user profile config.

YCShen1010:
Hi @bitscuit, I have updated the condition to handle the co-existence scenario where multiple tenants deploy different Keycloak versions in the same cluster.

1. In `tenant-a` namespace, KC 26 is deployed first, and the CRD is at version 26. At this point, the `hostname.backchannel` field should be defined under the `.spec` field and set to `true`.
2. Later, `tenant-b` deploys KC 24 in its own namespace, causing the CRD to downgrade to version 24. Since KC 26 requires the back channel setting, we need to move `backchannelDynamic` back under `additionalOptions` as shown below:
    ```
    additionalOptions:
     - name: hostname-backchannel-dynamic
        value: 'true'
    ```
3. For any version lower than 26 (e.g., 24 and 22), this value does not need to be set.
4. If it is KC 26 operator and 26 CRD, we define it under `additionalOptions`, not under `.spec`, it will generate a warning message, that we need to avoid
    ```
    warning: You need to specify these fields as the first-class citizen of the CR: hostname-backchannel-dynamic
    ```

That are all the cases.

Daniel-Fan (approved):
/lgtm

PR #2504, opened 2025-05-01
bump up hook weight

bluzarraga:

Fixes # https://github.ibm.com/PrivateCloud-analytics/Zen/issues/41456

1. How the test is done?

### DIFF
--- a/bedrock-migration/templates/bedrock-migration-job.yaml
+++ b/bedrock-migration/templates/bedrock-migration-job.yaml
@@ -4,8 +4,8 @@ metadata:
   name: bedrock-migration-job
   namespace: {{ .Values.global.operatorNamespace }}
   annotations:
-    "helm.sh/hook-delete-policy": before-hook-creation
-    "helm.sh/hook": pre-install
+    "helm.sh/hook-delete-policy": before-hook-creation,hook-succeeded,hook-failed
+    "helm.sh/hook": post-install,post-upgrade
     "helm.sh/hook-weight": "2"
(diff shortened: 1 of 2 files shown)


mandymchu:
@bluzarraga based on our validation experience so far, you may want to delete your migration sa as the last action in your clean up job 
for example, https://github.ibm.com/PrivateCloud-analytics/cpd-platform-operator/blob/5.2.x/charts/cpd-platform-migration/templates/00-clean-job.yaml#L154

bluzarraga:
@mandymchu why delete it instead of giving it a hook like the roles and rolebindings?

mandymchu:
> @mandymchu why delete it instead of giving it a hook like the roles and rolebindings?

in cpd-platform's migration chart, we don't set helm hook for service account. This is because we found that it does not work properly if each objects defined in one template has a helm hook. As such, we only add helm hook for role, rolebinding and job, but not for service account. Meanwhile, deleting the service account by the clean up job

bluzarraga:
I have the chart working as a cluster admin at the moment, but have not been able to test as tenant admin yet

mandymchu:
@bluzarraga for test, you only need to install cpfs by the existing OLM way, and then run `helm install` against your migration chart via a tenant admin. No need to install zen or cpd platform

bluzarraga:
Successfully tested as tenant user, this pr is ready for review

Daniel-Fan (approved):
/lgtm

PR #2520, opened 2025-05-09
Update the image digest in the Cluster CR as referenced from the ConfigMap in the PostgreSQL operator bundle

YCShen1010:
By using EDB CASE 5.16+, request EDB 1.22/1.25. 
1. ODLM will read `cloud-native-postgresql-operand-images-config` ConfigMap, find the image digest and then put the value into Cluster CR `.spec.imageName`
2. If such ConfigMap does not exist, ODLM will fall back to read from `cloud-native-postgresql-image-list` ConfigMap to update imageName as before

Fixes # https://github.ibm.com/IBMPrivateCloud/roadmap/issues/66594

**How to test**:
1. Test Image: quay.io/yuchen_shen/cs_operator:edb_cm
2. Change the test image for cs operator and request `ibm-im-operator`
3. Check the value under `.spec.imageName` in the Cluster CR is the same as the value in `cloud-native-postgresql-operand-images-config` ConfigMap

### DIFF
--- a/internal/controller/constant/odlm.go
+++ b/internal/controller/constant/odlm.go
@@ -811,6 +811,10 @@ spec:
                           name: cloud-native-postgresql-image-list
                           key: edb-postgres-license-provider-image
                           namespace: {{ $.OperatorNs }}
+                      configMapKeyRef:
+                        name: cloud-native-postgresql-operand-images-config
+                        key: edb-postgres-license-provider-image
+                        namespace: {{ $.OperatorNs }}
                   name: edb-license
                   resources:
                     limits:
@@ -840,6 +844,10 @@ spec:
                           name: cloud-native-postgresql-image-list
                           key: edb-postgres-license-provider-image
                           namespace: {{ $.OperatorNs }}
+                      configMapKeyRef:
+                        name: cloud-native-postgresql-operand-images-config
+                        key: edb-postgres-license-provider-image
+                        namespace: {{ $.OperatorNs }}
                   name: restart-edb-pod
                   resources:
                     limits:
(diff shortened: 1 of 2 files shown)


internal/controller/constant/odlm.go
```diff
@@ -1768,11 +1768,12 @@ spec:
...
                     name: cloud-native-postgresql-image-list
-                    key: ibm-postgresql-14-operand-image
+                    key: ibm-postgresql-16-operand-image
                     namespace: {{ .OperatorNs }}
                 configMapKeyRef:
-                    name: ibm-cpp-config
-                    key: edb-keycloak-operand-image
+                  name: cloud-native-postgresql-operand-images-config
+                  key: ibm-postgresql-16-operand-image
+                  namespace: {{ .OperatorNs }}
```
Daniel-Fan, line 1776: Can we keep using Postgres 14 image for Keycloak EDB Cluster CR?

Direct upgrade between Postgres major versions is not supported.

Daniel-Fan (changes requested):

Daniel-Fan (changes requested):
Can we use update the image reference for EDB license job? For example, this is [one](https://github.com/IBM/ibm-common-service-operator/blob/fe1b8ea152d0acd8c36b786b59059c681b6ed94e/internal/controller/constant/odlm.go#L1666) of the image reference

```yaml
                  image:
                    templatingValueFrom:
                      default:
                        required: true
                        configMapKeyRef:
                          name: cloud-native-postgresql-image-list
                          key: edb-postgres-license-provider-image
                          namespace: {{ .OperatorNs }}
```

Daniel-Fan:
If the check test is failed to pass, we could increase the timeout to 480s in `common/scripts/lint_go.sh` file like [what we did for ODLM](https://github.com/IBM/ibm-common-service-operator/blob/master/common/scripts/lint_go.sh).

Daniel-Fan (approved):
/lgtm

PR #2507, opened 2025-05-02
added configuration to enable/disable HPA

bitscuit:
controller handling of t-shirt sizing replicas still to be implemented, but this unblocks certain testing

PR #2528, opened 2025-05-13
updated bundle CRD with hpa toggle

bitscuit:
Follow up PR because the previous one did not update the bundle manifests https://github.com/IBM/ibm-common-service-operator/pull/2507


Daniel-Fan (approved):
/lgtm

PR #2529, opened 2025-05-13
[SC2] Update the image digest in the Cluster CR as referenced from the Conf…

YCShen1010:
By using EDB CASE 5.16+, request EDB 1.22/1.25. 
1. ODLM will read `cloud-native-postgresql-operand-images-config` ConfigMap, find the image digest and then put the value into Cluster CR `.spec.imageName`
2. If such ConfigMap does not exist, ODLM will fall back to read from `cloud-native-postgresql-image-list` ConfigMap to update imageName as before

Cherry-pick: https://github.com/IBM/ibm-common-service-operator/pull/2520

Fixes # https://github.ibm.com/IBMPrivateCloud/roadmap/issues/66594

**How to test**:
1. Test Image: quay.io/yuchen_shen/cs_operator:edb_cm
2. Change the test image for cs operator and request `ibm-im-operator`
4. Check the value under `.spec.imageName` in the Cluster CR is the same as the value in `cloud-native-postgresql-operand-images-config` ConfigMap

### DIFF
--- a/common/scripts/lint_go.sh
+++ b/common/scripts/lint_go.sh
@@ -15,4 +15,4 @@
 # limitations under the License.
 #
 
-GOGC=25 golangci-lint run -c ./common/config/.golangci.yml --timeout=360s
+GOGC=25 golangci-lint run -c ./common/config/.golangci.yml --timeout=480s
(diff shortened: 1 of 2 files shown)


Daniel-Fan (approved):
/lgtm

PR #2543, opened 2025-06-05
Enable user-defined catalog source and namespace in CS CR

YCShen1010:
1.  User-Configurable Catalog Source for Services:
    *   Adopters can now specify their preferred catalog source (`spec.catalogName` and `spec.catalogNamespace`) directly in the `CommonService` CR.
    *   These values, when provided, will be propagated to `status.configStatus` in CS CR and subsequently used into `OperandRegistry`
    *   Operators and services deployed will then be sourced from this user-defined catalog
    *   If they are not provided in the CS CR, ODLM will select appropriate catalog for the operators
    *   The user-defined catalog are supported to be added, removed and changed to another one in CS CR
    *   Webhook will deny the request if only catalogsource name or catalogsource namespace is set in the CR, they should be set together

2.  Dedicated ODLM Catalog Source Aligned with CS Operator:
    *   The ODLM will now consistently use a dedicated catalog source as CS operator has

Fixes # https://github.ibm.com/IBMPrivateCloud/roadmap/issues/65905

**How to test**:
1. Test image: quay.io/yuchen_shen/cs_operator:catalog
2. Install a CSFS, and config catalog and its namespace under `spec.catalogName` and `spec.catalogNamespace` in CS CR, delete ODLM, restart CS pod
3. After ODLM coming up again, its catalog should be the same as CS operator and, the given catalog will be shown in OperandRegistry each entry

### DIFF
--- a/api/v3/commonservice_types.go
+++ b/api/v3/commonservice_types.go
@@ -355,17 +355,10 @@ func (r *CommonService) UpdateConfigStatus(CSData *CSData, operatorDeployed, ser
 		r.Status.ConfigStatus.ServicesNamespace = ServicesNamespace(CSData.ServicesNs)
 	}
 
-	if r.Spec.CatalogName != "" {
-		r.Status.ConfigStatus.CatalogName = r.Spec.CatalogName
-	} else {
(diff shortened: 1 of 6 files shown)


Daniel-Fan (approved):
/lgtm

PR #2394, opened 2025-02-21
Script updates for BR of no olm install

bluzarraga:
**What this PR does / why we need it**: This pr updates the labeling scripts to handle no olm scenarios as well as olm scenarios for BR. New restore files are also included in this pr to facilitate the restore process in a no olm scenario.

**Which issue(s) this PR fixes**:
Fixes # https://github.ibm.com/IBMPrivateCloud/roadmap/issues/65729

**Special notes for your reviewer**:

1. How the test is done?

**How to backport this PR to other branch**:
1. Add label to this PR with the target branch name `backport <branch-name>`
2. The PR will be automatically created in the target branch after merging this PR
3. If this PR is already merged, you can still add the label with the target branch name `backport <branch-name>` and leave a comment `/backport` to trigger the backport action

PR #2619, opened 2025-09-08
No olm core recipe

qpdpQ:
fusion recipe for no-olm cpfs services, including cs-operator, odlm, im, ui, edb and zen
Fixes # https://github.ibm.com/IBMPrivateCloud/roadmap/issues/67407

### DIFF
--- a/velero/backup/common-service/label-common-service.sh
+++ b/velero/backup/common-service/label-common-service.sh
@@ -523,7 +523,7 @@ function label_helm_cluster_scope(){
     ${OC} label secret sh.helm.release.v1.$odlm_release_name.v1 -n $odlm_release_namespace foundationservices.cloudpak.ibm.com=odlm-cluster  --overwrite=true 2>/dev/null
 
     #cs operator cluster resources (crds, clusterrole, clusterrolebinding), crd covered elsewhere in script
-    ${OC} label customresourcedefinition commonservices.operator.ibm.com foundationservices.cloudpak.ibm.com=cs-cluster --overwrite=true 2>/dev/null
+    ${OC} label customresourcedefinition commonservices.operator.ibm.com foundationservices.cloudpak.ibm.com=crd --overwrite=true 2>/dev/null
     ${OC} label clusterrole ibm-common-service-operator-$OPERATOR_NS foundationservices.cloudpak.ibm.com=cs-cluster  --overwrite=true 2>/dev/null
     ${OC} label clusterrolebinding ibm-common-service-operator-$OPERATOR_NS foundationservices.cloudpak.ibm.com=cs-cluster  --overwrite=true 2>/dev/null
     cs_release_name=$(${OC} get crd commonservices.operator.ibm.com -o jsonpath='{.metadata.annotations.meta\.helm\.sh/release-name}' --ignore-not-found)
@@ -580,7 +580,7 @@ function label_helm_namespace_scope(){
     ${OC} label rolebinding operand-deployment-lifecycle-manager foundationservices.cloudpak.ibm.com=odlm-chart -n $SERVICES_NS --overwrite=true 2>/dev/null
     
     #cs operator
-    ${OC} label commonservices common-service foundationservices.cloudpak.ibm.com=cs-chart -n $OPERATOR_NS --overwrite=true 2>/dev/null
+    ${OC} label commonservices common-service foundationservices.cloudpak.ibm.com=commonservice -n $OPERATOR_NS --overwrite=true 2>/dev/null
     ${OC} label deployment ibm-common-service-operator foundationservices.cloudpak.ibm.com=cs-chart -n $OPERATOR_NS --overwrite=true 2>/dev/null
     ${OC} label serviceaccount ibm-common-service-operator foundationservices.cloudpak.ibm.com=cs-chart -n $OPERATOR_NS --overwrite=true 2>/dev/null
     ${OC} label role ibm-common-service-operator foundationservices.cloudpak.ibm.com=cs-chart -n $OPERATOR_NS --overwrite=true 2>/dev/null
--- a/velero/spectrum-fusion/recipes/no-olm/core/child-cs-odlm-chart-recipe.yaml
+++ b/velero/spectrum-fusion/recipes/no-olm/core/child-cs-odlm-chart-recipe.yaml
@@ -19,6 +19,12 @@ spec:
       type: resource
     - includeClusterResources: true
       includedResourceTypes:
+        - customresourcedefinitions.apiextensions.k8s.io
+      labelSelector: foundationservices.cloudpak.ibm.com=crd
+      name: cs-crd
+      type: resource
+    - includeClusterResources: true
+      includedResourceTypes:
         # - clusterrole
         # - clusterrolebinding
         - customresourcedefinitions.apiextensions.k8s.io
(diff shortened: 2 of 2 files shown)


velero/spectrum-fusion/recipes/no-olm/core/child-zen-recipe-2.10.0.yaml
```diff
@@ -0,0 +1,112 @@
...
+    - labelSelector: foundationservices.cloudpak.ibm.com=zen5-data
+      name: zen-volume
+      includedNamespaces:
+        - <child recipe namespace>
+      type: volume
+    - includeClusterResources: true
+      includedResourceTypes:
+        - zenservices.zen.cpd.ibm.com
+        - customresourcedefinitions.apiextensions.k8s.io
+      labelSelector: foundationservices.cloudpak.ibm.com=zen
```
bluzarraga, line 21: The label-common-service.sh script labels the zenservice as zen-chart, we should probably either update the label-common-service.sh script to change what it labels in no olm mode or update this value to match what it is already doing

velero/spectrum-fusion/recipes/no-olm/core/child-cs-odlm-chart-recipe.yaml
```diff
@@ -0,0 +1,124 @@
...
+      labelSelector: foundationservices.cloudpak.ibm.com=cs-chart
+      name: cs-resources
+      type: resource
+    - includedResourceTypes:
+        - role
+        - rolebinding
+        - serviceaccount
+        - deployments
+        - secrets
+      labelSelector: foundationservices.cloudpak.ibm.com=odlm-chart
```
bluzarraga, line 44: Since odlm and cs operator now share a chart, they also share a helm secret associated with that chart. So we don't need to include the secret in the odlm resources since it will already be included in the cs-resources. I also get a partially failed message for that resource:
> [ERROR] 2025-09-16 21:01:12.754 [TM_0] workflow:195 - status="{'completionTimestamp': '2025-09-16T21:01:09Z', 'expiration': '2124-08-23T21:01:07Z', 'formatVersion': '1.1.0', 'hookStatus': {}, 'phase': 'Completed', 'progress': {'itemsBackedUp': 6, 'totalItems': 6}, 'startTimestamp': '2025-09-16T21:01:07Z', 'version': 1}"
[INFO] 2025-09-16 21:01:12.754 [TM_0] workflow:200 - {'name': 'cs-odlm-child-recipe.odlm-resources', 'isSuccess': False, 'message': 'There is 1 resource type missing from the backed up resources for this group. The following resources were found and backed up: (apps/v1/Deployment, rbac.authorization.k8s.io/v1/Role, rbac.authorization.k8s.io/v1/RoleBinding, v1/ServiceAccount)', 'severity': 'warning'}

you can see secret is missing from the resources backed up. This likely applies to the common ui/IM chart as well

velero/spectrum-fusion/recipes/no-olm/core/child-cs-odlm-chart-recipe.yaml
```diff
@@ -0,0 +1,124 @@
...
+      name: odlm-cluster-resources
+      type: resource
+    - includedResourceTypes:
+        - role
+        - rolebinding
+        - serviceaccount
+        - deployments
+        - secrets
+        - commonservices.operator.ibm.com
+      labelSelector: foundationservices.cloudpak.ibm.com=cs-chart
```
bluzarraga, line 35: common service cr is still labeled as commonservice and not included in cs-chart based on the label-common-service.sh script so the cs cr would not be covered in this group.

velero/spectrum-fusion/recipes/no-olm/core/child-cs-odlm-chart-recipe.yaml
```diff
@@ -0,0 +1,124 @@
...
+  name: cs-odlm-child-recipe
+  namespace: <child recipe namespace>
+spec:
+  appType: common-service
+  groups:
+    - includeClusterResources: true
+      includedResourceTypes:
+        - clusterrole
+        - clusterrolebinding
+        - customresourcedefinitions.apiextensions.k8s.io
```
bluzarraga, line 16: cs cr crd labeled as crd so would not be covered here if using the label-common-service.sh script, resulting in a partially failed

velero/spectrum-fusion/recipes/no-olm/core/child-im-ui-chart-recipe.yaml
```diff
@@ -0,0 +1,126 @@
...
+      labelSelector: foundationservices.cloudpak.ibm.com=iam-chart
+      name: im-resources
+      type: resource
+    - includeClusterResources: true
+      includedResourceTypes:
+        - role
+        - rolebinding
+        - serviceaccount
+        - deployments
+        - secrets
```
bluzarraga, line 45: same as for odlm here, iam owns the secret (at least in name)

velero/spectrum-fusion/recipes/no-olm/core/child-zen-recipe-2.10.0.yaml
bluzarraga: the zenservice crd is also labeled as zen-cluster so it is also not covered here. If neither are backed up we get a full failure

bluzarraga (changes requested):
This mostly comes down to labeling semantics. The label-common-service.sh script is not on the same page as some of the recipes here creating some gaps. Let's make sure to get on the same page as the script since other items are already based on how the script labels them

velero/backup/common-service/label-common-service.sh
```diff
@@ -267,7 +267,9 @@ function label_ibm_catalogsources() {
 function label_ns_and_related() {
 
     title "Start to label the namespaces, operatorgroups and secrets... "
-    namespaces=$(${OC} get configmap namespace-scope -n $OPERATOR_NS -oyaml | awk '/^data:/ {flag=1; next} /^  namespaces:/ {print $2; next} flag && /^  [^ ]+: / {flag=0}')
+    # namespaces=$(${OC} get configmap namespace-scope -n $OPERATOR_NS -oyaml | awk '/^data:/ {flag=1; next} /^  namespaces:/ {print $2; next} flag && /^  [^ ]+: / {flag=0}')
+    namespaces=$(${OC} get configmap common-service-maps -n kube-public -o jsonpath='{.data.common-service-maps\.yaml}' | awk -v ns="$OPERATOR_NS" 'BEGIN {found=0}/requested-from-namespace:/ {found=0} $0 ~ ns {found=1} found && /^ *-/ {print $2}' | tr '\n' ',' | sed 's/,$/\n/')
```
bluzarraga, line 271: The only problem with this is would alter the permissions for the label-common-service.sh job that CPD uses since the configmap is in kube public and we wouldn't be able to create permissions there when being used as a job. We ask the user to enter in all applicable namespaces as parameters for the script to run. I think we should just make the list of namespaces based on:
- op ns
- serv ns
- tethered namespace(s)
- cert manager ns (if specified)
- licensing ns (if specified)
- ls reporter ns (if specified)

bluzarraga (changes requested):
Changes to the recipe look good. We don't want to increase the permissions of the label-common-service.sh script though because of the attached job so lets rethink how we grab the namespaces

velero/spectrum-fusion/recipes/no-olm/core/child-cs-odlm-chart-recipe.yaml
bluzarraga: Ok I must not have used your version of the labeling script then, my bad. Only thing is if we change the labeling script we will need to update the no olm br instructions since it looks like the use the existing restore crd and restore cs cr steps https://github.com/IBM/ibm-common-service-operator/pull/2394/#issuecomment-2711539094 (from this issue https://github.ibm.com/IBMPrivateCloud/roadmap/issues/66140). It will probably be easier to leave the labels as is and add additional steps into the restore recipe. This way we also guarantee that the cs cr is there before the operator

PR #2625, opened 2025-09-18
Create Child recipe for UMS supported on Non-OLM to fusion BR process

YCShen1010:
- Add backup and restore support for IBM Usage Metering Service resources on Non-OLM in fusion by creating a child recipe and intergrating it with the parent CPFS recipe
- Including UMS CRDs amd CRs labeling functions into labeling script 

Fixes # https://github.ibm.com/IBMPrivateCloud/roadmap/issues/67328


bluzarraga (changes requested):
The recipe is functioning how I expect it to and that's good but I think we should make the directory changes in a separate PR. The fusion automation is currently setup to use the directory structure already in place so these changes would break the automation.

I think the folder structures make more sense but shouldn't be incorporated into this change and should also be made alongside changes to the automation script so it does not break

bluzarraga (approved):

PR #2638, opened 2025-09-30
Make namespace deletion optional for uninstall tenant script

bluzarraga:
**What this PR does / why we need it**: CPD has asked for an uninstall script that does not delete the namespaces and only deletes CPFS resources. The difficulty with this is that we use the namespace deletion to determine which resources to delete (when a namespace is terminating, it highlights the remaining resources with finalizers to remove to facilitate namespace deletion). No deleting the namespace means we have to be more prescriptive about what we are deleting.

In this PR, I have created a fork that would skip deleting the namespaces of the tenant in favor of deleting specific resources associated with CPFS after operand requests and operators are uninstalled. This is a relatively short list of resources though so its possible this would need to expand. For example, right now this will uninstall the zen operator but it will not delete the zenservice so all zen resources are still present minus the operator. Is that CPFS or Zen uninstall script responsibility?

Fixes # https://github.ibm.com/IBMPrivateCloud/roadmap/issues/67725

### DIFF
--- a/cp3pt0-deployment/uninstall_tenant.sh
+++ b/cp3pt0-deployment/uninstall_tenant.sh
@@ -75,11 +75,9 @@ function parse_arguments() {
             OPERATOR_NS=$1
             ;;
         --retain-ns)
-            shift
             RETAIN="true"
             ;;
         -f)
-            shift
             FORCE_DELETE=1
             ;;
         -v | --debug)
--- a/cp3pt0-deployment/uninstall_tenant.sh
+++ b/cp3pt0-deployment/uninstall_tenant.sh
@@ -19,6 +19,7 @@ OPERATOR_NS_LIST=""
 CONTROL_NS=""
 FORCE_DELETE=0
 DEBUG=0
+RETAIN="false"
 
 # ---------- Command variables ----------
 
@@ -46,7 +47,9 @@ function main() {
     delete_rbac_resource
     delete_webhook
     delete_unavailable_apiservice
-    delete_tenant_ns
+    if [[ $RETAIN == "false" ]]; then
+        delete_tenant_ns
+    fi
 }
 
 function parse_arguments() {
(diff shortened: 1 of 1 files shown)


cp3pt0-deployment/uninstall_tenant.sh
```diff
@@ -69,6 +74,10 @@ function parse_arguments() {
             shift
             OPERATOR_NS=$1
             ;;
+        --retain-ns)
+            shift
+            RETAIN="true"
+            ;;
         -f)
             shift
             FORCE_DELETE=1
```
Daniel-Fan, line 83:
```suggestion
        --retain-ns)
            RETAIN="true"
            ;;
        -f)
            FORCE_DELETE=1
```

If I understand it right, there is no additional argument following either the flag `--retain-ns` or `-f`, we might not need `shift` in this case.

Daniel-Fan (commented):
Also I am wondering if the `-f` flag will play any side effect in this case?

For example, if we enable `-f` and `--retain-ns` flag in the same time, it will [leave operators and operands untouched in the namespace](https://github.com/IBM/ibm-common-service-operator/pull/2638/files#diff-2fb1ab5b33fd59204fab21de962096557edf421245ecc39eb559053afc69178aR42).

cp3pt0-deployment/uninstall_tenant.sh
bluzarraga: So two things about the -f flag:
1. I want to change the function of the script as little as possible including possible parameters. Its possible this script is being used with set parameters already somewhere and I don't want to disrupt that. I want --retain-ns to be a separate "opt in" flag 
2. techincally, its role is to enable force delete of resources if the namespace gets stuck in terminating. Essentially, without this flag enabled, we do not run the remove_all_finalizers function which deletes all the resources the namespace is waiting on

PR #2641, opened 2025-10-06
Reduce cluster-level RBAC scope

YCShen1010:
Tighten the Common Service Operator’s RBAC footprint by limiting cluster-scoped permissions

Fixes # https://github.ibm.com/IBMPrivateCloud/roadmap/issues/67915

**What Changes**

- Pruned unused verbs/resources from role.yaml
- Shifted EDB ConfigMap write access into the namespaced Role.
- Propagated the RBAC updates through Helm templates, test fixtures, and the bundle CSV.

**How to test**
1. Test image:  `quay.io/yuchen_shen/cs_operator:reduce_permission`
2. Install the whole CPFS and see if everything working properly.


Daniel-Fan (commented):
Do we need update [this namespace scope RBAC yaml file](https://github.com/IBM/ibm-common-service-operator/blob/master/helm/templates/rbac.yaml) for helm chart based on the changes in CSV?

PR #2760, opened 2026-03-03
Edb license job removed from odlm.go

Jeremy-Cheng-stack:
Remove edb license job as it's no longer required and is causing edb to restart without a purpose which causes issues for other deployments. Remove all resources related to edb license.

Fixes #https://github.ibm.com/IBMPrivateCloud/roadmap/issues/68424

1. How the test is done?
Normal installation flow. Check with install of edb and im. Make sure these changes don't mark the operator's readiness prematurely,


Daniel-Fan (approved):
Thank you for the changes @Jeremy-Cheng-stack 

They look good to me!

PR #2776, opened 2026-03-24
Remove orphaned field from operandconfig

qpdpQ:
Remove orphaned field from operandconfig, after user remove it from CommonService CR.

Fixes # 
https://github.ibm.com/IBMPrivateCloud/roadmap/issues/68616
https://github.ibm.com/IBMPrivateCloud/roadmap/issues/69023


internal/controller/operandconfig.go
```diff
@@ -583,6 +583,13 @@ func (r *CommonServiceReconciler) updateOperandConfig(ctx context.Context, newCo
 
 	}
 
+	// Remove orphaned fields from OperandConfig
+	klog.V(2).Info("Removing orphaned fields from OperandConfig")
```
Daniel-Fan: Hi @qpdpQ 
May I ask that is there a reason that we are implementing a mechanism to "removeOrphanedFields"?

From my understanding, we would like to update OperandConfig when its hash value in the annotation does not match what we calculate based on configurations in CommonService CR. That being said, it would be something like
```golang
// Pseudocode for simpler approach:
func (r *CommonServiceReconciler) updateOperandConfig(ctx context.Context) error {
    // 1. Build complete desired state from ALL CommonService CRs
    desiredOpconServices := r.buildCompleteDesiredState(ctx)
    
    // 2. Calculate hash of desired state
    desiredHash := calculateHash(desiredOpconServices)
    
    // 3. Get existing OperandConfig
    existingOpcon := getOperandConfig(ctx)
    existingHash := calculateHash(existingOpcon.services)
    
    // 4. If hashes differ, replace entire services array
    if desiredHash != existingHash {
        existingOpcon.services = desiredOpconServices
        r.Update(ctx, existingOpcon)
    }
    
    return nil
}
```

Please let me know if I missed anything, thank you!

qpdpQ: I was thinking about adding logs to show which resource is removed, but this seems bring extra Complexity and maintenance difficulty will update it

Daniel-Fan:
Hi @qpdpQ 
If it is possible, would you please provide a new test result?
Thank you!

qpdpQ:
Test image `quay.io/yuchen_li1/common-service-operator-amd64:update_config`
adding following field about `zenFrontDoor` to cs cr
```
  services:
    - name: ibm-im-operator
      spec:
        authentication:
          config:
            zenFrontDoor: true
```
Screenshot 2026-04-27 at 5 11 29 PM
OperandConifg:
Screenshot 2026-04-27 at 5 11 40 PM

and then remove it from CS CR
Screenshot 2026-04-27 at 5 27 15 PM

OperandConfig:

Screenshot 2026-04-27 at 5 12 58 PM

Daniel-Fan (approved):
LGTM

PR #2794, opened 2026-04-21
Update migration job version to 0.0.4190

YCShen1010:
This implies that the migration chart needs to be refreshed with new image for every release, and the image digest will be injected into job template while CICD is building the chart. Subsequently, migration chart version need to be bumped every release like `0.0.x`

Fixes # https://github.ibm.com/IBMPrivateCloud/roadmap/issues/69228


Daniel-Fan (approved):
LGTM

PR #2811, opened 2026-05-05
Aggregate OperatorConfigs from all CommonService CRs

YCShen1010:
Users can now specify `operatorConfigs` with replica settings in any CommonService CRs, and the OperatorConfig will be properly created and updated with the aggregated configuration across all CRs in the cluster

Fixes # https://github.ibm.com/IBMPrivateCloud/roadmap/issues/69165

**How to test**:
Test image: quay.io/yuchen_shen/cs_operator:operatorconfig
1. Create a non-master CS CR with the custom replica like
```
spec:
  operatorConfigs:
    - name: cloud-native-postgresql-v1.25
      replicas: 3
```
2. Observe there would be three EDB pods shown up


Daniel-Fan (approved):
LGTM

PR #2813, opened 2026-05-06
fix: ibm-pg release name unbound variable in bash

qpdpQ:
Fixed all bash syntax errors in label-common-service.sh `-` unable to read by bash script use `_` instead


Daniel-Fan (approved):
LGTM

PR #2819, opened 2026-05-07
Support CNCF with kubectl commands in CPFS scripts

Jeremy-Cheng-stack:
Currently, the scripts are for OCP cluster using oc commands. We need support for CNCF which uses kubectl commands instead of oc. The PR creates a common util file for commands that are not the same between oc and kubectl. Where there are unique oc commands, it has been replaced so kubectl will work as well. 

Fixes #https://github.ibm.com/IBMPrivateCloud/roadmap/issues/67515
https://github.ibm.com/IBMPrivateCloud/roadmap/issues/68541

1. How the test is done?

Tested the 4 scripts in cp3pt0-deployment with a successful run-through for each one. 

Tested on cluster with flag ```--oc /usr/bin/kubectl``` when running the scripts.

### DIFF
--- a/cp3pt0-deployment/common/authorize-namespace.sh
+++ b/cp3pt0-deployment/common/authorize-namespace.sh
@@ -19,6 +19,10 @@
 # Project roles and role bindings to another namespace
 #
 
+# Get script directory for sourcing cli_compat.sh
+SCRIPT_DIR=$(cd $(dirname "$0") && pwd -P)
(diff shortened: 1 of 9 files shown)


Daniel-Fan (approved):
LGTM

Thank you @Jeremy-Cheng-stack !

After the PR is merged, could you please collaborate with SERT team to integrate those scripts into their CNCF cluster deployment for a comprehensive test? Thank you!

PR #2827, opened 2026-05-12
feat: pass user managed overrides to operand registry installation for non-OLM install

Daniel-Fan:
1. Updates the non-OLM CommonService controller to pass user-managed configuration overrides when installing or updating the OperandRegistry.
2. It also removes the temporary log message about skipping EDB Image ConfigMap creation.

It is to support BYO EDB operator at non-OLM installation AKA Helm Chart deployment.
When `userManaged` is configured for OperandRegistry, ODLM will stop looking for EDB operator deployment in operator namespace in non-OLM mode.


YCShen1010 (approved):
/lgtm

PR #2831, opened 2026-05-14
Update only annotations and spec for cs-ca-certificate

qpdpQ:
Update cs-ca-certificate when there is a change in annotation or spec. Also ensures that when the version annotation is updated in the cert, only the annotations and spec are modified during the update operation, preventing unintended changes to other fields

Fixes # https://github.ibm.com/IBMPrivateCloud/roadmap/issues/69136

### DIFF
--- a/internal/controller/bootstrap/init.go
+++ b/internal/controller/bootstrap/init.go
@@ -660,8 +660,23 @@ func (b *Bootstrap) CreateOrUpdateFromYaml(yamlContent []byte, alwaysUpdate ...b
 
 			resourceVersion := objInCluster.GetResourceVersion()
 			obj.SetResourceVersion(resourceVersion)
-			if err := b.UpdateObject(obj); err != nil {
-				errMsg = err
+
+			// Special handling for cs-ca-certificate: only update annotations and spec
+			if gvk.Kind == "Certificate" && obj.GetName() == "cs-ca-certificate" {
+				// Preserve existing object and only update annotations and spec
+				if newAnnotations := obj.GetAnnotations(); newAnnotations != nil {
+					objInCluster.SetAnnotations(newAnnotations)
+				}
+				if newSpec, found := obj.Object["spec"]; found {
+					objInCluster.Object["spec"] = newSpec
+				}
+				if err := b.UpdateObject(objInCluster); err != nil {
+					errMsg = err
+				}
+			} else {
+				if err := b.UpdateObject(obj); err != nil {
+					errMsg = err
+				}
 			}
 		}
 	}
--- a/internal/controller/bootstrap/init.go
+++ b/internal/controller/bootstrap/init.go
@@ -662,7 +662,7 @@ func (b *Bootstrap) CreateOrUpdateFromYaml(yamlContent []byte, alwaysUpdate ...b
 			obj.SetResourceVersion(resourceVersion)
 
 			// Special handling for cs-ca-certificate: only update annotations and spec
-			if gvk.Kind == "Certificate" && obj.GetName() == "cs-ca-certificate" {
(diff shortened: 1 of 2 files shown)


qpdpQ:
only certificate updated:
```
apiVersion: cert-manager.io/v1
kind: Certificate
metadata:
  annotations:
    version: 0.0.1
  resourceVersion: '36538775'
  name: cs-ca-certificate
  uid: d25c365b-21b3-4228-a7b9-1cea42d6fc1d
  creationTimestamp: '2026-05-13T15:40:16Z'
  generation: 8
  namespace: cs3
  labels:
    app.kubernetes.io/instance: cs-ca-certificate
    app.kubernetes.io/managed-by: cert-manager-controller
    app.kubernetes.io/name: Certificate
    ibm-cert-manager-operator/refresh-ca-chain: 'true'
    manage-cert-rotation: 'true'
    operator.ibm.com/managedByCsOperator: 'true'
spec:
  commonName: cs-ca-certificate
  duration: 17520h0m0s
  isCA: true
  issuerRef:
    kind: Issuer
    name: cs-ss-issuer
  privateKey:
    rotationPolicy: Always
  renewBefore: 5840h0m0s
  secretName: cs-ca-certificate-secret
  secretTemplate:
    labels:
      ibm-cert-manager-operator/refresh-ca-chain: 'true'
status:
  conditions:
    - lastTransitionTime: '2026-05-13T15:40:16Z'
      message: Certificate is up to date and has not expired
      observedGeneration: 8
      reason: Ready
      status: 'True'
      type: Ready
  notAfter: '2028-05-12T15:40:16Z'
  notBefore: '2026-05-13T15:40:16Z'
  renewalTime: '2027-09-12T07:40:16Z'
  revision: 1
``` 

No change in the secret and secret data
```
% oc get secret -n cs3 | grep cs-ca 
cs-ca-certificate-secret                               kubernetes.io/tls         3      9h
```

internal/controller/bootstrap/init.go
```diff
@@ -660,8 +660,23 @@ func (b *Bootstrap) CreateOrUpdateFromYaml(yamlContent []byte, alwaysUpdate ...b
 
 			resourceVersion := objInCluster.GetResourceVersion()
 			obj.SetResourceVersion(resourceVersion)
-			if err := b.UpdateObject(obj); err != nil {
-				errMsg = err
+
+			// Special handling for cs-ca-certificate: only update annotations and spec
+			if gvk.Kind == "Certificate" && obj.GetName() == "cs-ca-certificate" {
```
Daniel-Fan: Hi @qpdpQ 

May I ask that are there any specific fields changed unintentionally, leading us to this special handling for certificate?
From my understanding, CS operator controls Certificate's `.labels`, `.annotations` and `.spec`, and an object update request will update above three fields, leaving other fields including `.status` unchanged.

Please let me know if there is anything that I am missing, thank you!

qpdpQ: I think we allow user to modify the label of certificate (e.g. `ibm-cert-manager-operator/refresh-ca-chain: 'true'/ 'false'`), and I think we should not revert those labels during the upgrade.
So I only want to update the `annotation` and `spec` here

Daniel-Fan: Ah, it makes sense to me!

Daniel-Fan: Just a small recommendation to use [constant](https://github.com/IBM/ibm-common-service-operator/blob/73fb906a5847924492011233688d9a89df4114b9/internal/controller/constant/constant.go#L98) for Certificate name `cs-ca-certificate`

All other changes are good, thank you @qpdpQ

qpdpQ: good point! just updated

Daniel-Fan (approved):
LGTM

PR #2832, opened 2026-05-14
Find maximum replica count across all configs instead of first non-nil

YCShen1010:
Changed the loop to iterate through all configs and track the maximum replica value

Fixes # https://github.ibm.com/IBMPrivateCloud/roadmap/issues/69374

Test image: quay.io/yuchen_shen/cs_operator:edb
1. create two different CS CR with
    ```
    spec:
      operatorConfigs:
        - name: common-service-cnpg
          replicas: 2
    ```
    ```
    spec:
      operatorConfigs:
        - name: ibm-pg-operator-v28
          replicas: 4
   ```
2. As they are use same packagemanifest name, so the replica will go with the max of them which is 4 instead of only getting from the first config.

### DIFF
--- a/internal/controller/operatorconfig.go
+++ b/internal/controller/operatorconfig.go
@@ -99,18 +99,19 @@ func (r *CommonServiceReconciler) processOperatorConfigs(ctx context.Context, co
 	}
 
 	// Use the aggregated replica value (maximum across all CRs)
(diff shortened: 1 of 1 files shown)


Daniel-Fan (approved):
LGTM

PR #2845, opened 2026-06-08
skip if changeMap is longer than default map

qpdpQ:
skip the extra key when changedMap long than defaultMap to avoid panic issue.
This kind of issue could happened when user manually patch the operandconfig.
Fixes # https://github.ibm.com/IBMPrivateCloud/roadmap/issues/69248


Daniel-Fan (approved):
LGTM

PR #2849, opened 2026-06-10
Added owner references to the CS resources

YCShen1010:
**What this PR does / why we need it**:  Add owner reference for multiple Common Services resources, relying on Kubernetes garbage collection automatically cleaning them up when the CommonService CR is removed.

Added owner references to the following resources:

- cs-keycloak-theme ConfigMap
- cs-ca-certificate Certificate
- cs-ca-certificate-secret Secret

Fixes # https://github.ibm.com/IBMPrivateCloud/roadmap/issues/68246

**How to test**:

1. Verify owner references are set on the listed resources after CommonService CR creation
2. Test that resources are automatically deleted when CommonService CR is removed


internal/controller/bootstrap/init.go
```diff
@@ -83,6 +83,7 @@ type Bootstrap struct {
 	CSOperators          []CSOperator
 	CSData               apiv3.CSData
 	configMerger         ConfigMergerFunc
+	CSInstance           *apiv3.CommonService // CommonService CR instance for owner references
```
Daniel-Fan: **[Blocker — Race condition]** `CSInstance` stored as shared mutable state on the singleton `Bootstrap` struct. When multiple `CommonService` CRs are present, concurrent reconcile goroutines will overwrite this field without synchronisation, causing one reconcile to attach the wrong owner to a resource. Remove this field and pass `instance` as a parameter down through `addOwnerReference` and `shouldAddOwnerReference` instead.

internal/controller/bootstrap/init.go
```diff
@@ -241,6 +242,9 @@ func NewBootstrap(mgr manager.Manager) (bs *Bootstrap, err error) {
 
 // InitResources initialize resources at the bootstrap of operator
 func (b *Bootstrap) InitResources(instance *apiv3.CommonService, forceUpdateODLMCRs bool) error {
+	// Set the CommonService instance for owner references
+	b.CSInstance = instance
```
Daniel-Fan: This write to `b.CSInstance` is the root of the race condition — see the struct-level comment above. The `instance` pointer should be threaded through as a function argument rather than stored on the shared struct.

internal/controller/bootstrap/init.go
```diff
@@ -595,6 +599,87 @@ func (b *Bootstrap) CreateOrUpdateFromJson(objectTemplate string, alwaysUpdate .
...
+		return false
+	}
+
+	gvk := obj.GetObjectKind().GroupVersionKind()
+	name := obj.GetName()
+
+	// ConfigMaps that need owner references
+	if gvk.Kind == "ConfigMap" && gvk.Version == "v1" && gvk.Group == "" {
+		configMapsNeedingOwner := []string{
+			"cloud-native-postgresql-image-list",
```
Daniel-Fan: **[Blocker — Wrong resource scope]** `cloud-native-postgresql-image-list` does not appear anywhere else in this codebase — no constant, no template, no create/update call. We have removed this ConfigMap in 4.19.0 release, there is no need to set owner reference for it.

internal/controller/bootstrap/init.go
```diff
@@ -595,6 +599,87 @@ func (b *Bootstrap) CreateOrUpdateFromJson(objectTemplate string, alwaysUpdate .
...
+			"cs-keycloak-theme",
+			constant.IBMCPPCONFIG, // "ibm-cpp-config"
```
Daniel-Fan: **[Blocker — Wrong resource scope]** `ibm-cpp-config` is not created by this operator — it is a pre-existing platform-wide configuration ConfigMap. Line 1864 in this same file logs it as a *required* prerequisite: `"Configmap %s/%s is required"`. The operator only reads and propagates copies of it via `PropagateCPPConfig()`. Attaching a `CommonService` owner reference here would garbage-collect this critical cluster-wide config when the CR is deleted.

internal/controller/bootstrap/init.go
```diff
@@ -595,6 +599,87 @@ func (b *Bootstrap) CreateOrUpdateFromJson(objectTemplate string, alwaysUpdate .
...
+			constant.NamespaceScopeConfigmapName, // "namespace-scope"
```
Daniel-Fan: **[Blocker — Conflict with existing code]** `CleanNamespaceScopeResources()` (around line 2110 in this file) explicitly patches the `namespace-scope` ConfigMap to *remove* any owner references from it as part of the v3→v4 migration. Adding an owner reference here creates a fight with that cleanup code on every reconcile cycle. Additionally, this ConfigMap belongs to the `ibm-namespace-scope-operator`, not to `CommonService`.

internal/controller/bootstrap/init.go
```diff
@@ -595,6 +599,87 @@ func (b *Bootstrap) CreateOrUpdateFromJson(objectTemplate string, alwaysUpdate .
...
+	}
+
+	if b.CSInstance == nil {
+		klog.V(2).Infof("Skipping owner reference for %s/%s: CSInstance is nil", obj.GetNamespace(), obj.GetName())
+		return nil
+	}
+
+	// Create owner reference
+	ownerRef := metav1.OwnerReference{
+		APIVersion:         b.CSInstance.APIVersion,
```
Daniel-Fan: **[Blocker — Empty APIVersion at runtime]** The Kubernetes Go client strips `TypeMeta` from objects it decodes, so `b.CSInstance.APIVersion` will be `""` when `instance` is populated via `r.Reader.Get()`. This produces an invalid owner reference. Use `constant.APIVersion` (= `"operator.ibm.com/v3"`) which is already defined for this purpose and used at lines 2356, 2377, 2388.

internal/controller/bootstrap/init.go
```diff
@@ -595,6 +599,87 @@ func (b *Bootstrap) CreateOrUpdateFromJson(objectTemplate string, alwaysUpdate .
...
+		Kind:               b.CSInstance.Kind,
```
Daniel-Fan: **[Blocker — Empty Kind at runtime]** Same issue as `APIVersion` above — `b.CSInstance.Kind` will be `""` at runtime due to `TypeMeta` stripping. Use `constant.KindCR` (= `"CommonService"`) instead.

internal/controller/bootstrap/init.go
```diff
@@ -595,6 +599,87 @@ func (b *Bootstrap) CreateOrUpdateFromJson(objectTemplate string, alwaysUpdate .
...
+	return false
+}
+
+// addOwnerReference adds an owner reference to the object if appropriate
+func (b *Bootstrap) addOwnerReference(obj *unstructured.Unstructured) error {
+	if !b.shouldAddOwnerReference(obj) {
+		return nil
+	}
+
+	if b.CSInstance == nil {
```
Daniel-Fan: **[Nit — Unreachable code]** This `nil` check is dead code. `shouldAddOwnerReference()` already returns `false` when `b.CSInstance == nil`, so `addOwnerReference` can only reach this line when `CSInstance` is non-nil. Remove it.

Daniel-Fan (changes requested):
## Review

Thank you for working on owner reference management — the helper decomposition (`shouldAddOwnerReference` / `addOwnerReference`) is clean and the deduplication check before appending is a nice touch. However there are several issues that need to be addressed before merging, ranging from blockers to scope problems.

---

### 🔴 Blockers

**1. `APIVersion` and `Kind` will be empty strings**

The Kubernetes Go client strips `TypeMeta` from objects it decodes, so `b.CSInstance.APIVersion` and `b.CSInstance.Kind` will both be `""` at runtime when `instance` is populated via `r.Reader.Get()`. This will produce an invalid owner reference that Kubernetes either rejects or silently ignores — defeating the entire purpose of this PR.

Use the constants that are already defined for exactly this purpose and used elsewhere in the file (lines 2356, 2377, 2388):

```go
ownerRef := metav1.OwnerReference{
    APIVersion: constant.APIVersion, // "operator.ibm.com/v3"
    Kind:       constant.KindCR,     // "CommonService"
    Name:       b.CSInstance.Name,
    UID:        b.CSInstance.UID,
    ...
}
```

**2. Cross-namespace owner references will be rejected by the API server**

Kubernetes forbids owner references across namespaces. Several resources in the allowlist may live in `ServicesNs` while the `CommonService` CR lives in `OperatorNs`. When these differ, the API server returns a 422 Unprocessable Entity. A namespace equality guard is required in `addOwnerReference` before appending.
We can choose the CommonService CR named `common-service` in services namespace after this CommonService CR is created.

**3. `CSInstance` is shared mutable state — data race under concurrent reconciliation**

`Bootstrap` is a single struct shared across all reconcile loops. When multiple `CommonService` CRs exist and reconciliations run concurrently, each goroutine writes `b.CSInstance = instance` without any synchronisation. A subsequent read in `addOwnerReference` may see a different CR's pointer, attaching the wrong owner to a resource. The field should not be stored on the struct — pass `instance` as a parameter down the call chain instead.

---

### ⚠️ Over-broad allowlist — three resources should be removed

**`namespace-scope`**: This ConfigMap belongs to the `ibm-namespace-scope-operator`, not to `CommonService`. More critically, `CleanNamespaceScopeResources()` in this same file (around line 2110) explicitly patches this ConfigMap to *remove* any owner references from it as part of the v3→v4 migration. Adding an owner reference here creates a fight with that existing cleanup code on every reconcile cycle.

**`ibm-cpp-config`**: This is a pre-existing platform-wide configuration ConfigMap, not created by this operator. The code at line 1864 treats it as a required prerequisite (`"Configmap %s/%s is required"`). The operator only reads and propagates copies of it. Attaching a `CommonService` owner to it would cause this critical cluster-wide config to be garbage-collected when the CR is deleted.

**`cloud-native-postgresql-image-list`**: This name does not appear anywhere else in the codebase — no constant, no template, no create/update call. There is no evidence this operator creates or manages this ConfigMap. We already dropped this ConfigMap in 4.19.0 release.

The three that are appropriate are: `cs-ca-certificate` (Certificate), `cs-ca-certificate-secret` (Secret), and `cs-keycloak-theme` (ConfigMap) — all directly created by this operator's bootstrap.

---

### Minor nits

- The `nil` check on `b.CSInstance` at line 649 is unreachable — `shouldAddOwnerReference` already returns `false` if `CSInstance == nil`, so `addOwnerReference` can only reach line 649 when it is non-nil.
- The boolean pointer pattern `func() *bool { t := true; return &t }()` is verbose; consider a small helper or `ptr.To(true)` if `k8s.io/utils/ptr` is available.
- `"cs-ca-certificate-secret"` should be a named constant in `constant/constant.go` for consistency.
- No unit tests for the new helpers — given the namespace/TypeMeta edge cases, table-driven tests would be valuable.

internal/controller/bootstrap/init.go
```diff
@@ -595,7 +595,80 @@ func (b *Bootstrap) CreateOrUpdateFromJson(objectTemplate string, alwaysUpdate .
...
+	if gvk.Kind == "Certificate" && gvk.Group == "cert-manager.io" {
+		if name == constant.CSCACertificate { // "cs-ca-certificate"
+			return true
+		}
+	}
+
+	// Secret resources that need owner references
+	if gvk.Kind == "Secret" && gvk.Version == "v1" && gvk.Group == "" {
+		if name == "cs-ca-certificate-secret" {
+			return true
```
Daniel-Fan, line 622: **[Nit]** `"cs-ca-certificate-secret"` is a bare string literal. Define a named constant in `constant/constant.go` (e.g. `CSCACertificateSecret`) for consistency with `constant.CSCACertificate` used on line 621.

internal/controller/bootstrap/init.go
```diff
@@ -595,7 +595,80 @@ func (b *Bootstrap) CreateOrUpdateFromJson(objectTemplate string, alwaysUpdate .
...
+	}
+
+	// Create owner reference
+	ownerRef := metav1.OwnerReference{
+		APIVersion:         constant.APIVersion,
+		Kind:               constant.KindCR,
+		Name:               instance.Name,
+		UID:                instance.UID,
+		Controller:         func() *bool { t := true; return &t }(),
+		BlockOwnerDeletion: func() *bool { t := true; return &t }(),
```
Daniel-Fan: **[Nit]** The `func() *bool { t := true; return &t }()` pattern appears twice here. A small helper `func boolPtr(b bool) *bool { return &b }` or `ptr.To(true)` from `k8s.io/utils/ptr` is idiomatic in the Kubernetes ecosystem.

internal/controller/bootstrap/init.go
```diff
@@ -595,7 +595,80 @@ func (b *Bootstrap) CreateOrUpdateFromJson(objectTemplate string, alwaysUpdate .
...
+	for _, ref := range existingOwnerRefs {
+		if ref.UID == ownerRef.UID {
+			klog.V(2).Infof("Owner reference already exists for %s/%s", obj.GetNamespace(), obj.GetName())
+			return nil
+		}
+	}
+
+	// Add the new owner reference
+	existingOwnerRefs = append(existingOwnerRefs, ownerRef)
+	obj.SetOwnerReferences(existingOwnerRefs)
```
Daniel-Fan, line 665: **[Blocker — Cross-namespace owner reference]** `cs-ca-certificate`, `cs-ca-certificate-secret`, and `cs-keycloak-theme` all live in `ServicesNs`, while the owning `CommonService` CR lives in `instance.Namespace` (`OperatorNs`). These are separate namespaces in the standard topology. Kubernetes rejects owner references that cross namespace boundaries with a 422 error.

Add a namespace guard before this append:
```go
// Kubernetes does not allow cross-namespace owner references
if obj.GetNamespace() != "" && obj.GetNamespace() != instance.Namespace {
    klog.V(2).Infof("Skipping cross-namespace owner ref for %s/%s (owner is in %s)",
        obj.GetNamespace(), obj.GetName(), instance.Namespace)
    return nil
}
```

If the goal is GC of `ServicesNs` resources when the CR is deleted, a deletion finalizer on the `CommonService` CR that explicitly deletes them is the correct Kubernetes pattern — cross-namespace GC via owner references is not supported.

Daniel-Fan (changes requested):
## Review (updated)

Good progress on this revision — the three blockers from the first round (race condition, empty TypeMeta, over-broad allowlist) have all been addressed. The `CSInstance` field is gone, `constant.APIVersion`/`constant.KindCR` are now used correctly, and the allowlist has been trimmed to the three resources that are actually created by this operator.

One new compile error was introduced in the process, and the cross-namespace safety issue I flagged in the first review still hasn't been addressed. Details below.

---

### 🔴 Blocker 1: Compile error — `DeployResource` at line 1808 not updated

`DeployResource()` still calls `b.CreateOrUpdateFromYaml([]byte(...))` with only one argument, but the signature was changed to require a second `*apiv3.CommonService` parameter. This is a build-breaking error confirmed locally:

```
internal/controller/bootstrap/init.go:1808:34: not enough arguments in call to b.CreateOrUpdateFromYaml
    have ([]byte)
    want ([]byte, *v3.CommonService, ...bool)
```

`DeployResource` has no `instance` in scope, so pass `nil` — the resources it deploys are not in the owner-reference allowlist:

```go
err = b.CreateOrUpdateFromYaml([]byte(util.Namespacelize(cr, placeholder, b.CSData.ServicesNs)), nil)
```

---

### 🔴 Blocker 2: Cross-namespace owner reference — still unaddressed

`cs-ca-certificate`, `cs-ca-certificate-secret`, and `cs-keycloak-theme` all live in `ServicesNs` (set from `instance.Status.ConfigStatus.ServicesNamespace`). The owning `CommonService` CR lives in `instance.Namespace`, which maps to `OperatorNs`. These are two distinct namespaces in the standard topology — the codebase explicitly tracks the separation:

```go
if b.CSData.CPFSNs != b.CSData.ServicesNs { // line 284
```

Kubernetes rejects cross-namespace owner references with a 422. The `addOwnerReference` function must guard against this before appending:

```go
// Kubernetes does not allow cross-namespace owner references
if obj.GetNamespace() != "" && obj.GetNamespace() != instance.Namespace {
    klog.V(2).Infof("Skipping cross-namespace owner ref for %s/%s (owner is in %s)",
        obj.GetNamespace(), obj.GetName(), instance.Namespace)
    return nil
}
```

If the intent is for these resources to be garbage-collected with the CR, a deletion finalizer on the `CommonService` CR that explicitly cleans them up is the correct Kubernetes pattern — cross-namespace GC via owner references is not supported.

---

### Minor nits (carried from first review, still present)

- **Boolean pointer style**: `func() *bool { t := true; return &t }()` (lines 648–649) is verbose. A small helper or `ptr.To(true)` from `k8s.io/utils/ptr` is idiomatic.
- **Bare string constant**: `"cs-ca-certificate-secret"` at line 628 should be a named constant in `constant/constant.go` for consistency with `constant.CSCACertificate` on line 621.
- **No unit tests**: the new helpers still have no test coverage.

internal/controller/bootstrap/init.go
Daniel-Fan, line 665: **[Blocker — Cross-namespace owner reference]** This `append` runs without any check that `obj` and `instance` are in the same namespace. All three resources in the allowlist live in `ServicesNs` while the `CommonService` CR lives in `instance.Namespace` (`OperatorNs`). Kubernetes rejects cross-namespace owner references with HTTP 422.

Add a guard here, before the append:
```go
// Kubernetes does not allow cross-namespace owner references
if obj.GetNamespace() != "" && obj.GetNamespace() != instance.Namespace {
    klog.V(2).Infof("Skipping cross-namespace owner ref for %s/%s (owner is in %s)",
        obj.GetNamespace(), obj.GetName(), instance.Namespace)
    return nil
}
```
If the intent is GC of `ServicesNs` resources on CR deletion, a deletion finalizer is the correct pattern — cross-namespace garbage collection via owner references is not supported by Kubernetes.

internal/controller/bootstrap/init.go
Daniel-Fan, line 622: **[Nit]** `"cs-ca-certificate-secret"` is a bare string literal. Define a named constant in `constant/constant.go` (e.g. `CSCACertificateSecret`) for consistency with `constant.CSCACertificate` on line 621.

internal/controller/bootstrap/init.go
Daniel-Fan: **[Nit]** The `func() *bool { t := true; return &t }()` IIFE appears twice. Prefer `ptr.To(true)` from `k8s.io/utils/ptr`, or a small package-level helper.

Daniel-Fan (changes requested):
## Review (round 3)

The compile error from round 2 (`DeployResource` missing the `instance` argument) has been fixed. The build now passes cleanly.

However, the **cross-namespace owner reference blocker is still unaddressed** — no new code was added to handle it. All other previous findings remain as before.

---

### 🔴 Remaining Blocker: Cross-namespace owner reference

`addOwnerReference` still has no namespace guard. All three resources in the allowlist live in `ServicesNs`, while the owning `CommonService` CR lives in `instance.Namespace` (= `OperatorNs`). These are tracked as separate fields in `CSData` and are two distinct namespaces in the standard topology:

| Resource | Namespace (from template) |
|---|---|
| `cs-ca-certificate` | `b.CSData.ServicesNs` (via `Namespacelize`, line 2005) |
| `cs-ca-certificate-secret` | `b.CSData.ServicesNs` (cert-manager generates it in the same ns) |
| `cs-keycloak-theme` | `b.CSData.ServicesNs` (template: `namespace: "{{ .ServicesNs }}"`)

The `CommonService` CR itself lives in `instance.Namespace`, which equals `CSData.OperatorNs` (confirmed at `api/v3/commonservice_types.go:473`).

Kubernetes rejects cross-namespace owner references with HTTP 422. The fix is to add a guard in `addOwnerReference` before the `append`:

```go
// Kubernetes does not allow cross-namespace owner references
if obj.GetNamespace() != "" && obj.GetNamespace() != instance.Namespace {
    klog.V(2).Infof("Skipping cross-namespace owner ref for %s/%s (owner is in %s)",
        obj.GetNamespace(), obj.GetName(), instance.Namespace)
    return nil
}
```

Note: if the real goal is to garbage-collect `ServicesNs` resources when the CR is deleted, owner references cannot achieve this across namespaces. A deletion finalizer on the `CommonService` CR that explicitly deletes those resources is the correct Kubernetes pattern.

---

### Minor nits (still present)

- **Bare string constant**: `"cs-ca-certificate-secret"` should be a named constant in `constant/constant.go`.
- **Boolean pointer style**: `func() *bool { t := true; return &t }()` — prefer `ptr.To(true)` or a small helper.
- **No unit tests** for `shouldAddOwnerReference` / `addOwnerReference`.

Daniel-Fan:
## Current status after round 3 review

Good progress across the revisions — most of the original blockers have been resolved. Here is the consolidated state of what remains open as of commit `6580572`.

---

### ✅ Resolved

| Finding | Fixed in |
|---|---|
| Race condition — `CSInstance` shared mutable state on `Bootstrap` struct | commit `3a309373` |
| Empty `APIVersion`/`Kind` on owner reference | commit `3a309373` |
| Over-broad allowlist (`namespace-scope`, `ibm-cpp-config`, `cloud-native-postgresql-image-list`) | commit `987f58f5` |
| Compile error — `DeployResource` missing `instance` argument | commit `65805722` |

---

### 🔴 Still blocking: Cross-namespace owner reference

**Location:** `addOwnerReference()` at line 664 — the `append` has no namespace guard.

All three resources in the allowlist are deployed into `ServicesNs`:

| Resource | Namespace |
|---|---|
| `cs-ca-certificate` (Certificate) | `b.CSData.ServicesNs` — via `Namespacelize()` at line 2005 |
| `cs-ca-certificate-secret` (Secret) | `b.CSData.ServicesNs` — cert-manager creates it there |
| `cs-keycloak-theme` (ConfigMap) | `b.CSData.ServicesNs` — template: `namespace: "{{ .ServicesNs }}"` |

The owning `CommonService` CR lives in `instance.Namespace`, which maps to `OperatorNs` (confirmed at `api/v3/commonservice_types.go:473`). In the standard topology these are different namespaces.

**Kubernetes rejects cross-namespace owner references with HTTP 422.** The owner reference will never be accepted by the API server when `OperatorNs != ServicesNs`.

There are three ways to address this:

**Option A — Add a namespace guard (silently skips when namespaces differ):**
```go
// addOwnerReference, before the append at line 664:
if obj.GetNamespace() != "" && obj.GetNamespace() != instance.Namespace {
    klog.V(2).Infof("Skipping cross-namespace owner ref for %s/%s (owner is in %s)",
        obj.GetNamespace(), obj.GetName(), instance.Namespace)
    return nil
}
```
This makes the function safe, but the feature becomes a no-op in the typical multi-namespace topology.

**Option B — Use a deletion finalizer (correct approach for cross-namespace cleanup):**
Add a finalizer to the `CommonService` CR and, in the finalizer handler, explicitly delete the target resources in `ServicesNs`. This is the standard Kubernetes pattern for cross-namespace garbage collection since owner-reference-based GC is namespace-scoped.

**Option C — Scope this PR to same-namespace deployments only:**
If `OperatorNs == ServicesNs` (e.g. single-namespace install), owner references do work. Option A's guard can be combined with a log message to make the same-namespace-only behaviour explicit.

---

### 🔵 Minor nits (non-blocking)

- `"cs-ca-certificate-secret"` at line 628 — define a named constant (e.g. `constant.CSCACertificateSecret`) for consistency with `constant.CSCACertificate`
- `func() *bool { t := true; return &t }()` at lines 648–649 — prefer `ptr.To(true)` from `k8s.io/utils/ptr`
- No unit tests for `shouldAddOwnerReference` / `addOwnerReference` — table-driven tests would catch the namespace edge case and the allowlist logic

internal/controller/bootstrap/init.go
```diff
@@ -49,6 +49,7 @@ import (
 	"k8s.io/client-go/rest"
 	"k8s.io/client-go/tools/record"
 	"k8s.io/klog"
+	"k8s.io/utils/ptr"
 	"sigs.k8s.io/controller-runtime/pkg/client"
```
Daniel-Fan, line 54: **[Blocker — Wrong package path]** `k8s.io/utils/ptr` does not exist in the version pinned by this module (`v0.0.0-20230406110748-d93618cff8a2`). The `ptr` sub-package was added in a later release. Build fails locally:
```
no required module provides package k8s.io/utils/ptr
```
The `pointer` sub-package **does** exist in the pinned version and provides the same functionality:
```go
import "k8s.io/utils/pointer"

Controller:         pointer.Bool(true),
BlockOwnerDeletion: pointer.Bool(true),
```
Use `k8s.io/utils/pointer` instead to avoid a `go.mod` bump.

Daniel-Fan (changes requested):
## Review (round 4)

Great progress — the cross-namespace guard is now in place, the constant `CSCACertificateSecret` has been added, and `ptr.To(true)` is used for the bool pointers. Three of the four previously open findings are now resolved.

One new build-breaking issue was introduced: `k8s.io/utils/ptr` does not exist in the version of `k8s.io/utils` pinned by this module.

---

### 🔴 Blocker: `k8s.io/utils/ptr` does not exist in the pinned module version

This module pins `k8s.io/utils v0.0.0-20230406110748-d93618cff8a2` in `go.mod`. The `ptr` sub-package was only introduced in `k8s.io/utils` around `v0.0.0-20230726`. The pinned version does **not** contain a `ptr` directory — confirmed locally:

```
internal/controller/bootstrap/init.go:52:2: no required module provides package k8s.io/utils/ptr
```

The equivalent package that **does** exist in the pinned version is `k8s.io/utils/pointer`, which exposes `pointer.Bool(b bool) *bool`.

The fix is either:

**Option A** — Use the existing `k8s.io/utils/pointer` package (no `go.mod` change needed):
```go
import "k8s.io/utils/pointer"

// then:
Controller:         pointer.Bool(true),
BlockOwnerDeletion: pointer.Bool(true),
```

**Option B** — Bump `k8s.io/utils` to a version that contains `ptr` and run `go mod tidy` + vendor update. This is a larger change and may have transitive effects.

Option A is the minimal fix that requires no dependency changes.

---

### ✅ Resolved this round

| Finding | Commit |
|---|---|
| Cross-namespace owner reference — missing namespace guard | `7c28615` |
| `"cs-ca-certificate-secret"` bare string literal | `0049fd65` |
| `func() *bool { t := true; return &t }()` verbose bool pointer | `7c28615` (intent correct, wrong package) |

---

### 🔵 Remaining nit

- **No unit tests** for `shouldAddOwnerReference` / `addOwnerReference`. Given the namespace guard and the allowlist logic, table-driven tests would be a good safety net.

Daniel-Fan (approved):
## Review (round 5) — Approved ✅

All blockers and nits from the previous rounds have been resolved. The PR is now correct and ready to merge.

---

### ✅ Full resolution history

| Finding | Resolved |
|---|---|
| Race condition — `CSInstance` shared mutable state on `Bootstrap` struct | `3a309373` |
| Empty `APIVersion`/`Kind` on owner reference | `3a309373` |
| Over-broad allowlist (`namespace-scope`, `ibm-cpp-config`, `cloud-native-postgresql-image-list`) | `987f58f5` |
| Compile error — `DeployResource` missing `instance` argument | `65805722` |
| Cross-namespace owner reference — no namespace guard | `7c28615f` |
| `"cs-ca-certificate-secret"` bare string literal | `0049fd65` |
| `func() *bool { t := true; return &t }()` verbose bool pointer / wrong package | `fdd8939e` |

---

### What the final state looks like

- **`shouldAddOwnerReference`** — clean allowlist of three operator-owned resources (`cs-keycloak-theme`, `cs-ca-certificate`, `cs-ca-certificate-secret`), guarded by `instance == nil`.
- **`addOwnerReference`** — uses `constant.APIVersion` / `constant.KindCR` for a correct owner reference, has a proper cross-namespace guard (`obj.GetNamespace() != instance.Namespace`), uses `pointer.Bool(true)` from the already-pinned `k8s.io/utils/pointer` package, and deduplicates before appending.
- **`instance` threading** — no shared mutable state; `instance` is correctly passed through `renderTemplate` → `CreateOrUpdateFromYaml` → `addOwnerReference` for all relevant call sites, and `nil` is passed everywhere owner references are not intended.
- **Constants** — `CSCACertificateSecret` added to `constant/constant.go` alongside `CSCACertificate`.
- **Build** — `go build` and `go vet` pass cleanly on the bootstrap package.

---

### 🔵 Optional follow-up (non-blocking)

Unit tests for `shouldAddOwnerReference` and `addOwnerReference` would be a good addition in a follow-up PR — in particular for the namespace guard edge cases and the allowlist logic. The existing test suite has a pre-existing unrelated panic in `TestSetConfigMerger_NilByDefault` that should be addressed separately.

Daniel-Fan (approved):
LGTM

PR #2870, opened 2026-07-08
Set default value for autoScaleConfig

qpdpQ:
Set a default `autoScaleConfig: false` if it is not set, 
So when user remove this field from commonservice, cs-operator will change the value to false and ODLM will update the CR base on this
Fixes # https://github.ibm.com/IBMPrivateCloud/roadmap/issues/68616

### DIFF
--- a/internal/controller/config_extractor.go
+++ b/internal/controller/config_extractor.go
@@ -114,25 +114,27 @@ func extractFeatureConfigs(cs *apiv3.CommonService) ([]interface{}, error) {
 	}
 
 	// Extract AutoScaleConfig configuration
-	// Check if autoScaleConfig field was explicitly set in the spec
+	// If autoScaleConfig is not set, treat it as false
+	autoScaleConfigValue := false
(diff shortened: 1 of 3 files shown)


qpdpQ:
CS operator able to set this value without any config
```
 ~ % oc get commonservice common-service -oyaml -n cs2 | grep autoScaleConfig
 ~ % oc get operandconfig common-service -oyaml -n cs2 | grep autoScaleConfig
        autoScaleConfig: false
        autoScaleConfig: false
```

Daniel-Fan (changes requested):
Hi @qpdpQ 
Can you check if [test case `TestExtractCommonServiceConfigs_EmptyCS`](https://github.com/IBM/ibm-common-service-operator/blob/69061ada32437877896c6c3c09cc9a1edeee4461/internal/controller/config_extractor_test.go#L49) needs to be updated?

Daniel-Fan (approved):
LGTM

PR #2873, opened 2026-07-10
Fix OperandConfig Race Condition issue

qpdpQ:
## Summary
Ensure aggregated CommonService configuration is applied to OperandConfig before ODLM can create common-service-db, preventing a race where the IBM PG cluster CR is created without a storage class.

## Problem
When a storage class is set on a configurable CommonService CR instead of the master `common-service` CR, cs-operator previously created or updated OperandConfig in two phases:

- bootstrap created an initial OperandConfig
- later reconciliation aggregated all CommonService CRs and patched the final storage settings

If ODLM reconciled faster than cs-operator’s second phase, it could create common-service-db without `storageClass`, leaving the DB pending.

## Fix
- Build the aggregated desired state from all CommonService CRs **before** calling Bootstrap.InitResources
- Pass aggregated configs into InstallOrUpdateOpcon so bootstrap can create the first OperandConfig in its final form
- Add aggregated merge support in bootstrap/config_bridge.go
- Wire the aggregated merger in SetupWithManager
- Keep the existing post-bootstrap updateOperandConfig reconciliation as a convergence and safety pass
- Update the no-OLM path to use the new InstallOrUpdateOpcon signature

Fixes #  
https://github.ibm.com/IBMPrivateCloud/roadmap/issues/69720
https://github.ibm.com/PrivateCloud-analytics/CPD-Quality/issues/119054

### DIFF
--- a/internal/controller/bootstrap/config_bridge.go
+++ b/internal/controller/bootstrap/config_bridge.go
@@ -27,10 +27,19 @@ import (
 // This allows bootstrap to use controller merge logic without import cycles
 type ConfigMergerFunc func(baseConfig string, cs *apiv3.CommonService, servicesNs string) (string, error)
 
+// AggregatedConfigMergerFunc merges base config with an already aggregated multi-CR view.
+// This lets bootstrap create OperandConfig in its final desired form before ODLM consumes it.
+type AggregatedConfigMergerFunc func(baseConfig string, configs []interface{}, serviceControllerMapping map[string]string, servicesNs string) (string, error)
+
 // SetConfigMerger sets the configuration merger function on the Bootstrap instance.
 // Called during reconciler initialization to inject the merge logic.
 func (b *Bootstrap) SetConfigMerger(merger ConfigMergerFunc) {
 	b.configMerger = merger
+}
+
+// SetAggregatedConfigMerger sets the aggregated configuration merger function on the Bootstrap instance.
+func (b *Bootstrap) SetAggregatedConfigMerger(merger AggregatedConfigMergerFunc) {
+	b.aggregatedConfigMerger = merger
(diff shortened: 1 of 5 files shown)


qpdpQ:
Test resule:
```
PASS|1.1|Global StorageClass Applied to All Services|All services have storageClass
PASS|1.2|Global StorageClass with Service Having Other Configs (PRIMARY BUG FIX)|Global storageClass applied correctly (storage=managed-nfs-storage, wal=managed-nfs-storage)
PASS|2.1|Service-Specific StorageClass Override|Override works, other services use global
PASS|3.1|Configuration Not Lost During Reconciliation|Configs preserved across 3 reconciliations
PASS|4.1|HPA Enabled to Disabled Transition|HPA disabled and field removed successfully
PASS|4.2|FIPS Enabled Configuration|FIPS enabled in OperandConfig
PASS|5.1|ZenFrontDoor Boolean in Service Spec (Lifecycle Test)|All lifecycle steps passed
PASS|5.2|RouteHost Configuration|RouteHost configured in OperandConfig
PASS|6.1|Empty StorageClass String Should Not Override|Empty string did not override
PASS|7.1|Second CR Provides StorageClass When First CR Doesn't|Second CR's storageClass used (first-non-empty-wins)
PASS|8.1|Multi-CR Double Update Bug Detection (CRITICAL)|Single update, no missing storageClass (updates=1)
PASS|9.1|CS-Operator Update Complete - Aggregated Config Before Bootstrap|Aggregated config applied before Bootstrap (gen=310, storage=managed-nfs-storage, wal=managed-nfs-storage)

```

Daniel-Fan (approved):
LGTM

PR #2877, opened 2026-07-16
Increase waiting time to avoid time out error in heavy workloak cluster

qpdpQ:
Increase waiting time to avoid time out error in heavy workload cluster
Fixes # https://github.ibm.com/IBMPrivateCloud/roadmap/issues/69747#issuecomment-222218604


bitscuit (approved):
Just increasing overall timeout period to ~6 min. No functional impact

PR #2879, opened 2026-07-20
Fix certificate secret label issue

qpdpQ:
The certificate secret was not labeled with `operator.ibm.com/watched-by-cert-manager` because it was created directly by cert-manager without this label. As a result, the cs-operator does not add the secret to its cache, making it unavailable through `client.Get()`. To retrieve the secret in this case, use `reader.Get()` instead.
Fixes # https://github.ibm.com/IBMPrivateCloud/roadmap/issues/69759


Daniel-Fan (approved):
LGTM

PR #2881, opened 2026-07-22
Re-implement Set default value for autoScaleConfig

qpdpQ:
re-implement default value for autoScaleConfig, set default value in base merge layer.
Fix the issue when `nil` value could be converted to` false` and overwritten `true` value in autoScaleConfig.
Fixes # https://github.ibm.com/IBMPrivateCloud/roadmap/issues/68616


Daniel-Fan (approved):
LGTM

PR #2882, opened 2026-07-22
Add namespace suffix to pg migrator role and rolebinding

qpdpQ:
Add a namespace suffix to the pg migrator Role and RoleBinding to ensure unique RBAC resource names. Fix the issue when running in single-namespace mode, the migrator RBAC resources generated for the operator namespace and the data namespace will conflict and over-written each other.

Fixes # https://github.ibm.com/IBMPrivateCloud/roadmap/issues/69770


Daniel-Fan (approved):
LGTM

PR #2891, opened 2026-08-17
Adopt Zen 7.0

qpdpQ:
Adopt Zen 7.0 channel in operandregistry
Fixes # https://github.ibm.com/IBMPrivateCloud/roadmap/issues/69882

### DIFF
--- a/internal/controller/constant/odlm.go
+++ b/internal/controller/constant/odlm.go
@@ -2985,7 +2985,9 @@ spec:
     sourceNamespace: "{{ .CatalogSourceNs }}"
   - name: ibm-platformui-operator
     namespace: "{{ .CPFSNs }}"
-    channel: v6.10
+    channel: v7.0
+    fallbackChannels:
+      - v6.10
     packageName: ibm-zen-operator
     scope: public
     installPlanApproval: {{ .ApprovalMode }}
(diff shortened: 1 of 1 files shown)


internal/controller/constant/odlm.go
```diff
@@ -2974,7 +2985,7 @@ spec:
     sourceNamespace: "{{ .CatalogSourceNs }}"
   - name: ibm-platformui-operator
     namespace: "{{ .CPFSNs }}"
-    channel: v6.10
+    channel: v7.0
```
Daniel-Fan: If I understand it correctly, we still pair Bedrock 4.19.x with Zen version 6.10.x as the default option.

Testing Zen 7.0.x with CPFS 4.19.x is only a temporary internal scenario? Please correct me if I was wrong.

Daniel-Fan (changes requested):

internal/controller/constant/odlm.go
qpdpQ: good point, let's keep default channl for now, will update if needed

qpdpQ:
Thanks for @Daniel-Fan pointing out use fallback channels to handle this case,
when user requesting `ibm-platformui-operator`
if zen `v7.0` catalogsource exist, ODLM will install zen in `v7.0` channel
if zen `v7.0` catalogsource not exist, and zen `v6.10` catalogsource exist, ODLM will install zen in `v6.10` channel
if both zen `v7.0` `v6.10` catalogsource not exist, this operandrequest will fail.

Daniel-Fan (approved):
LGTM

PR #2900, opened 2026-08-24
Allow CS operator to run without DaemonSet permissions

YCShen1010:
Enhance the ibm-common-service-operator so that DaemonSet pod refresh is optional and missing DaemonSet permissions do not fail Certificate reconciliation.

This change allows restricted installations to omit DaemonSet RBAC while preserving the existing behavior by default.

Fixes # https://github.ibm.com/IBMPrivateCloud/roadmap/issues/69868

**Changes**:
#### 1. Runtime behavior

- Added a `MANAGE_DAEMONSETS` runtime option.
- Added SelfSubjectAccessReview checks for the required DaemonSet permissions:
  - `list`
  - `watch`
  - `update`
- Skip DaemonSet pod refresh when:
  - DaemonSet management is explicitly disabled.
  - SelfSubjectAccessReview reports missing permissions.
  - SelfSubjectAccessReview cannot be completed.
  - DaemonSet List or Update returns `Forbidden` after a successful access review.
- Missing DaemonSet permissions no longer fail or requeue the complete Certificate reconciliation.
- Deployment and StatefulSet pod refresh continues normally.
- Removed an unused, non-namespaced Deployment List operation from the refresh path.

#### 2. Helm configuration

Added the following value, enabled by default for backward compatibility:

```yaml
podRefresh:
  manageDaemonSets: true

### CODE REVIEW

File: internal/controller/cert-manager/podrefresh_controller.go [hunk 1]
```diff
@@ -292,11 +329,36 @@ func (r *PodRefreshReconciler) restart(secret, cert, namespace string, lastUpdat
...
+	}
+	if !access.allowed {
+		klog.Warningf("DaemonSet pod refresh is not permitted in namespace %q (missing %q permission); skipping DaemonSet management", namespace, access.deniedVerb)
+		return nil
+	}
+
+	daemonsetsToUpdate, err := r.getDaemonSetNeedUpdate(ctx, secret, namespace, lastUpdated)
 	if err != nil {
+		if errors.IsForbidden(err) {
+			klog.Warningf("DaemonSet list is forbidden in namespace %q after a successful access review; skipping DaemonSet pod refresh", namespace)
```
Daniel-Fan: I am wondering that would it make sense to return error here given we already pass daemonsets related permissions assessment and if additional error happens, we need to catch it and return error.

internal/controller/cert-manager/podrefresh_controller.go
```diff
@@ -292,11 +329,36 @@ func (r *PodRefreshReconciler) restart(secret, cert, namespace string, lastUpdat
...
+			return nil
+		}
 		return err
 	}
-	if err := r.updateDaemonSetAnnotations(daemonsetsToUpdate, cert, secret, timeNow); err != nil {
+	if err := r.updateDaemonSetAnnotations(ctx, daemonsetsToUpdate, cert, secret, timeNow); err != nil {
+		if errors.IsForbidden(err) {
+			klog.Warningf("DaemonSet update is forbidden in namespace %q after a successful access review; skipping remaining DaemonSet pod refresh", namespace)
+			return nil
```
Daniel-Fan, line 368: Maybe the same change applies here? We could directly return error.

internal/controller/cert-manager/podrefresh_controller.go
```diff
@@ -292,11 +329,36 @@ func (r *PodRefreshReconciler) restart(secret, cert, namespace string, lastUpdat
...
+	if err != nil {
+		if errors.IsForbidden(err) || errors.IsUnauthorized(err) {
+			klog.Warningf("Unable to review DaemonSet permissions in namespace %q; skipping DaemonSet pod refresh: %v", namespace, err)
+			return nil
+		}
+		return fmt.Errorf("error reviewing daemonset permissions in namespace %q: %w", namespace, err)
+	}
+	if !access.allowed {
+		klog.Warningf("DaemonSet pod refresh is not permitted in namespace %q (missing %q permission); skipping DaemonSet management", namespace, access.deniedVerb)
+		return nil
```
Daniel-Fan, line 346: I am just curious that in what scenario that we will get `access.allowed: false` without any error like `IsForbidden` or `IsUnauthorized`?

internal/controller/cert-manager/podrefresh_controller.go
YCShen1010: The same reasoning applies to Update. Permission may be revoked or an admission policy may reject the update after the SSAR succeeds. Since DaemonSet refresh is optional, a `Forbidden` update should be observable through a warning but should not fail the complete Certificate reconciliation. Other update errors are still returned.

internal/controller/cert-manager/podrefresh_controller.go
YCShen1010: I think the `Forbidden` fallback should remain non-blocking. The SSAR is a point-in-time preflight check, and authorization or admission policy can change before the actual List call. Returning the `Forbidden` error here would allow optional DaemonSet refresh to fail the complete Certificate reconciliation again, which is the behavior this PR is intended to prevent. Non-Forbidden errors are still returned for retry.

internal/controller/cert-manager/podrefresh_controller.go
YCShen1010: A denied target permission is reported through `SelfSubjectAccessReview.Status.Allowed=false`; it is not returned as an API error.
`Forbidden` or `Unauthorized` from `client.Create(review)` means the operator could not submit the SelfSubjectAccessReview itself, while `Allowed=false` means the review completed successfully and the requested DaemonSet verb was not authorized.

YCShen1010:
Test result for podrefresh cases:
```
=== RUN   TestRestartSkipsDaemonSetsWhenPermissionIsDenied
W0826 08:24:42.985024 1032607 podrefresh_controller.go:345] DaemonSet pod refresh is not permitted in namespace "test-ns" (missing "update" permission); skipping DaemonSet management
--- PASS: TestRestartSkipsDaemonSetsWhenPermissionIsDenied (0.19s)
=== RUN   TestRestartReturnsErrorWhenAccessReviewFails
--- PASS: TestRestartReturnsErrorWhenAccessReviewFails (0.00s)
=== RUN   TestRestartSkipsDaemonSetsWhenAccessReviewIsForbidden
W0826 08:24:42.991509 1032607 podrefresh_controller.go:339] Unable to review DaemonSet permissions in namespace "test-ns"; skipping DaemonSet pod refresh: selfsubjectaccessreviews.authorization.k8s.io is forbidden: forbidden
--- PASS: TestRestartSkipsDaemonSetsWhenAccessReviewIsForbidden (0.00s)
=== RUN   TestRestartContinuesWhenDaemonSetListBecomesForbidden
W0826 08:24:42.995891 1032607 podrefresh_controller.go:352] DaemonSet list is forbidden in namespace "test-ns" after a successful access review; skipping DaemonSet pod refresh: error getting daemonsets: daemonsets.apps is forbidden: forbidden
--- PASS: TestRestartContinuesWhenDaemonSetListBecomesForbidden (0.00s)
=== RUN   TestRestartContinuesWhenDaemonSetUpdateBecomesForbidden
W0826 08:24:42.999060 1032607 podrefresh_controller.go:363] DaemonSet update is forbidden in namespace "test-ns" after a successful access review; skipping remaining DaemonSet pod refresh: error updating daemonset: daemonsets.apps "uses-secret" is forbidden: forbidden
--- PASS: TestRestartContinuesWhenDaemonSetUpdateBecomesForbidden (0.00s)
=== RUN   TestRestartUpdatesDaemonSetsWhenPermissionIsGranted
--- PASS: TestRestartUpdatesDaemonSetsWhenPermissionIsGranted (0.00s)
=== RUN   TestSelfSubjectDaemonSetPermissionCheckerChecksRequiredVerbs
--- PASS: TestSelfSubjectDaemonSetPermissionCheckerChecksRequiredVerbs (0.00s)
PASS
ok  	github.com/IBM/ibm-common-service-operator/v4/internal/controller/cert-manager	0.242s
```

internal/controller/cert-manager/podrefresh_controller.go
Daniel-Fan: Got it. It makes sense now. Can you double check if the operator itself has the permission to submit SelfSubjectAccessReview?

YCShen1010: Yes. In a standard Kubernetes/OpenShift cluster, the operator ServiceAccount can submit SelfSubjectAccessReviews. The default `system:basic-user `ClusterRole grants create permission on `selfsubjectaccessreviews.authorization.k8s.io` and is bound to authenticated identities, including ServiceAccounts.
Therefore, we do not need to add this permission explicitly to the operator RBAC. In a restricted cluster where that default permission has been removed, the controller already handles Forbidden or Unauthorized from the SSAR request by logging a warning and skipping DaemonSet management without blocking reconciliation.

I also verified it with:
```
oc auth can-i create selfsubjectaccessreviews.authorization.k8s.io \
  --as="system:serviceaccount:edb-config:ibm-common-service-operator"
Warning: resource 'selfsubjectaccessreviews' is not namespace scoped in group 'authorization.k8s.io'

yes
```

YCShen1010:
Test result on a cluster: 
https://console-openshift-console.apps.installer-cp3pt0.cp.fyre.ibm.com/k8s/ns/cert-owner/operators.coreos.com~v1alpha1~ClusterServiceVersion

The operator ServiceAccount can create `selfsubjectaccessreviews.authorization.k8s.io`.
I also verified both runtime paths:

- With DaemonSet RBAC, the SSAR succeeds and pod-refresh reconciliation completes normally.
- After temporarily removing the DaemonSet rule, the SSAR returns `Allowed=false`; the controller logs a warning and skips DaemonSet management without returning a reconciliation error.

YCShen1010:
@Daniel-Fan I think it is ready to merge now

Daniel-Fan (approved):
LGTM

PR #2910, opened 2026-09-01
Update values.yaml to add ODLM values

qpdpQ:
Context in [Slack](https://ibm-cloudplatform.slack.com/archives/G013X1K33AN/p1769612308578209), since we will merge ODLM chart into CS chart so all the value in `values.yaml` should be in CS chart, `values.yaml` in ODLM chart will be ignored
Fixes # https://github.ibm.com/IBMPrivateCloud/roadmap/issues/70002


bitscuit (approved):
Discussed in slack about the sev 1, and this looks good

PR #2915, opened 2026-09-05
enhance cs_im_mustgather_script

avinash102:

The IAM MustGather script was hardcoded to use the `cp-console` route when obtaining the IAM host for token and API calls. This caused the script to silently fail in environments using the Unified Console, where only the `cpd` route exists. This PR adds `cpd` route fallback support and improves overall robustness of the script.

Changes made:
- Added `cp-console` → `cpd` route fallback: script now checks for `cp-console` first and falls back to `cpd` if it does not exist, supporting both classic and Unified Console environments
- Added guard on empty IAM host — logs a warning and skips API calls instead of making broken `curl` requests to `https://`
- Added guard on IAM admin credentials — skips token fetch if credentials cannot be retrieved
- IDP, users, and SCIM queries now only execute when a valid access token is obtained (previously ran unconditionally even after a token failure)
- `oc cp` calls now use `|| echo WARNING` so a copy failure logs a warning and continues instead of aborting
- No-pods case for `platform-auth-service` changed from `exit 1` to a warning so the rest of collection still runs
- `2>/dev/null` added to `oc get authentications` calls to suppress noise on stderr

Fixes [#69888](https://github.ibm.com/IBMPrivateCloud/roadmap/issues/69888)

1. How the test is done?
   - Run the script against a cluster using the **classic console** (has `cp-console` route) — verify `IAM_HOST` resolves to the `cp-console` host and token/API calls succeed
   - Run the script against a cluster using the **Unified Console** (only has `cpd` route, no `cp-console`) — verify the script falls back to `cpd` and token/API calls succeed
   - Run the script with an invalid namespace or missing secret and verify warnings are printed without the script aborting

### DIFF
--- a/must-gather/cs_im_mustgather.sh
+++ b/must-gather/cs_im_mustgather.sh
@@ -27,24 +27,23 @@ echo "Logs will be stored in: $AUTHMGDIR/$NAMESPACE"
 PODS=$(oc -n "$NAMESPACE" get pods -l component=platform-auth-service --no-headers -o custom-columns=name:.metadata.name)
 
 if [[ -z "$PODS" ]]; then
-  echo "No pods found with label component=platform-auth-service in namespace '$NAMESPACE'"
-  exit 1
+  echo "INFO: No pods found with label component=platform-auth-service in namespace '$NAMESPACE'"
+else
+  for pod in $PODS; do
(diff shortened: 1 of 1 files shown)


Daniel-Fan (approved):
LGTM

PR #2916, opened 2026-09-10
Enforce master CR as sole controller owner

YCShen1010:
The fix enforces only the CommonService CR named common-service (the master CR) may ever be the controller owner of shared resources, regardless of which CR happens to be reconciling at any given moment.

Fixes # https://github.ibm.com/IBMPrivateCloud/roadmap/issues/70051

**What  changed**:
1. `addOwnerReference` (init.go)
Before building the owner reference, the function now checks whether the reconciling instance is the master CR. If it is not, it fetches `common-service` from the API server and uses its `Name` and `UID` instead. This means no secondary CR can ever become an owner going forward, no matter which reconcile loop runs first. If the master CR is not found yet, ownership is silently skipped rather than blocking the reconcile.

2. `EnsureControllerOwnerReference` (ownerreference.go)
When the resource already carries im-common-service as its controller, the master CR needs to reclaim it. This function now allows the incoming owner to replace an existing controller reference when both the old and new controller are the same Kind (CommonService).

### DIFF
--- a/internal/controller/bootstrap/init.go
+++ b/internal/controller/bootstrap/init.go
@@ -631,9 +631,14 @@ func (b *Bootstrap) shouldAddOwnerReference(
 	return false
 }
 
-// addOwnerReference ensures that obj is controlled by instance. It returns true
-// when the object was changed. The caller is responsible for persisting the
-// change and logging only after that persistence succeeds.
+// addOwnerReference ensures that obj is controlled by the master CommonService
+// CR (named constant.MasterCR).  When the reconciling instance is not the
+// master CR itself, the master CR is fetched from the API server so that its
+// UID is always used.  This guarantees that only one, well-known CR can ever
+// become the controller owner regardless of which CR triggered reconciliation.
+//
+// It returns true when the object was changed.  The caller is responsible for
+// persisting the change and logging only after that persistence succeeds.
 func (b *Bootstrap) addOwnerReference(
 	obj *unstructured.Unstructured,
 	instance *apiv3.CommonService,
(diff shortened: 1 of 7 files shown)


internal/controller/bootstrap/init.go
```diff
@@ -642,25 +641,29 @@ func (b *Bootstrap) addOwnerReference(
...
-		return false, nil
+	owner := instance
+	if owner.Name != constant.MasterCR || owner.Namespace != obj.GetNamespace() {
+		owner = &apiv3.CommonService{}
+		key := types.NamespacedName{Namespace: obj.GetNamespace(), Name: constant.MasterCR}
+		if err := b.Reader.Get(ctx, key, owner); err != nil {
+			if errors.IsNotFound(err) {
+				// PropagateDefaultCR creates the local clone later; its create
+				// event requeues the master CR to backfill ownership.
+				klog.V(2).Infof("Waiting for CommonService %s before adding owner reference to %s/%s",
```
Daniel-Fan, line 653: **`addOwnerReference` uses the package-level `context.Background()`, not the reconcile context**

`CreateOrUpdateFromYaml` — the only caller of this function — takes no `context.Context` parameter, so this `b.Reader.Get` call falls through to the package-level `var ctx = context.Background()` defined at line 95. This means:

- **Cancellation is not propagated.** If the reconcile is cancelled or times out, this lookup will not be aborted.
- **Trace/span context is lost**, which hurts observability in environments that inject tracing via context.

The pattern predates this PR (line 95 is pre-existing), so it is a carry-forward rather than a regression introduced here. But this is the first real API call placed on this code path, so it is a good moment to fix it. Suggested approach: thread `ctx context.Context` through `CreateOrUpdateFromYaml` and `addOwnerReference`:

```go
// Before
func (b *Bootstrap) CreateOrUpdateFromYaml(yamlContent []byte, instance *apiv3.CommonService, alwaysUpdate ...bool) error
func (b *Bootstrap) addOwnerReference(obj *unstructured.Unstructured, instance *apiv3.CommonService) (bool, error)

// After
func (b *Bootstrap) CreateOrUpdateFromYaml(ctx context.Context, yamlContent []byte, instance *apiv3.CommonService, alwaysUpdate ...bool) error
func (b *Bootstrap) addOwnerReference(ctx context.Context, obj *unstructured.Unstructured, instance *apiv3.CommonService) (bool, error)
```

internal/controller/common/ownerreference.go
```diff
@@ -53,6 +59,13 @@ func EnsureControllerOwnerReference(object metav1.Object, owner metav1.OwnerRefe
```
Daniel-Fan, line 52: **`replaceExistingSameKind` is invariably `true` at the only production call site**

The parameter was designed as a caller-controlled safety gate, and its doc comment says _"callers should only pass true when owner is the designated master CR"_. However `addOwnerReference` is the only production caller, and it **always** resolves `owner` to the local master CR before calling this function — so the parameter never varies in practice.

A boolean that is always `true` at the only call site does not enforce the invariant it was designed to enforce; a future caller could accidentally pass `true` without understanding the contract. Two cleaner options:

**Option A** — Remove the parameter and make the same-Kind replacement unconditional inside the function (already guarded by the `APIVersion+Kind` match):
```go
func EnsureControllerOwnerReference(object metav1.Object, owner metav1.OwnerReference) (bool, error)
```

**Option B** — Keep the parameter but also assert inside `addOwnerReference` that `owner.Name == constant.MasterCR` before passing `true`, making the implicit contract explicit:
```go
if owner.Name != constant.MasterCR {
    panic(fmt.Sprintf("BUG: replaceExistingSameKind=true called with non-master owner %q", owner.Name))
}
```

Either option makes the intent clearer than relying on doc-comment conventions alone.

Daniel-Fan (approved):
lgtm

